PR #1, opened 2018-04-10
[WIP] setup and product-list

lderrickable:


damienwebdev:
I'd consider shifting the entire project up a directory.

lderrickable:
One thing I'm not quite sure about is how the services inside of "app" are supposed to access the API url inside the environments files. How does it decide which url to use? I assume there should be a configuration somewhere to decide which "example app" to pull the url from.

damienwebdev:
So, I thought about this a bit, and I reviewed a few possible implementations. I'm thinking we will need to provide some "base" environment file as a sample in the project, and our users will have to copy that into their projects environment file.

As an example, I like what they're doing here...
My only concern is potential pathing issues in attempts to access the environment file.

https://github.com/dereklin/nx-demo-jest/blob/lazy-5/apps/app1/src/app/services/index.ts

damienwebdev:
@lderrickable Was thinking about this...

I think the best way to approach this will be to use a config object passed down into DaffodilModule and then used inside the services...

 We'll probably want to do something in the theme's `app.module.ts` like...

```
...
DaffodilModule.forRoot(SOME_CONFIG_OBJECT)
...
```

lderrickable:
I think I'm still a little unclear on how the user will use the themes. I assumed they would install "foundation-daffodil" as a dependency and it would end up in their node_modules directory. If that's the case, we still wouldn't be able to know what their service url is without them reaching into their node_modules/foundation-daffodil. Or will defining SOME_CONFIG_OBJECT on the root of an application just make it available to daffodil?

PR #4, opened 2018-04-30
Add license

damienwebdev:


damienwebdev:
@lderrickable I think we should do this under the MIT License. This repo should be open from the very beginning. Thoughts?

lderrickable:
Sounds good. I had added a `license: "MIT"` to the package.json, but I didn't know you needed an extra file

PR #7, opened 2018-04-30
Cart module with no page view

lderrickable:


apps/foundation-app/src/app/app-routing.module.ts
```diff
@@ -1,8 +1,21 @@
...
+import { CartViewComponent } from './cart/pages/cart-view/cart-view.component';
 
 export const appRoutes: Routes = [
-
+  {
+    path: '',
+    component: ProductListViewComponent
+  },
+  {
+    path: 'product-list',
```
damienwebdev: This is fine for now, but eventually we'll need to add "category" support. So, there won't be just "product-list", there'll be a wildcard like structure /collection/{whatever}

apps/foundation-app/src/app/cart/components/cart/cart.component.html
```diff
@@ -0,0 +1,3 @@
+<div class="cart-container">
+  {{carts[0].id}}
```
damienwebdev: Carts shouldn't be an array.

apps/foundation-app/src/app/cart/components/cart/cart.component.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+      declarations: [ CartComponent ]
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(CartComponent);
+    component = fixture.componentInstance;
+
+    component.carts = new Array(cartFactory.create(), cartFactory.create());
```
damienwebdev: same thing here, a cart is a single object, not an array.

damienwebdev:
@lderrickable I think this needs some minor changes to change the cart "array" implementation to a single cart object,

apps/foundation-app/src/app/cart/components/cart/cart.component.html
lderrickable: At some point I think `carts` needs to be an array, because of how ngrx-entity works. It returns an array of whatever it stores on state, right? I just wasn't sure where it should be changed into an object. I could do it in the `cart.component` module instead:

```
this.carts$ = this.store.pipe(
      select(fromCart.selectAllCarts)
);
```
would change to 
```
this.cart$ = this.store.pipe(
      select(fromCart.selectAllCarts)[0]
);
```

Or maybe I just shouldn't be using ngrx-entity in this case?

lderrickable: changed

apps/foundation-app/src/app/cart/components/cart/cart.component.spec.ts
lderrickable: changed

apps/foundation-app/src/app/cart/components/cart/cart.component.spec.ts
```diff
@@ -0,0 +1,35 @@
...
+    component = fixture.componentInstance;
+
+    component.cart = cartFactory.create();
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+
```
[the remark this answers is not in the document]
lderrickable: Isn't that tested on line 66?

PR #13, opened 2018-05-10
View list of products

lderrickable:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/apps/foundation-app/package.json
+++ b/apps/foundation-app/package.json
@@ -1,0 +1,10 @@
+{
+  "name": "@daffodil/foundation",
+  "version": "0.0.0",
+  "license": "MIT",
+  "scripts": {},
+  "private": true,
+  "dependencies": {
+    "foundation-sites": "^6.4.4-rc1"
+  }
+}
--- a/apps/foundation-app/src/app/product/components/product-list/product-list.component.html
+++ b/apps/foundation-app/src/app/product/components/product-list/product-list.component.html
@@ -1,4 +1,11 @@
-<div class="product-container" *ngFor="let product of products">
-  {{product.id}}
-  {{product.cost}}
+<div class="card-container">
+  <span class="card" *ngFor="let product of products">
+      <img src="https://image.freepik.com/free-icon/shopping-cart-with-product-inside_318-59544.jpg">
+      <div class="card-section">
+          <div class="product-brand single-line-ellipsis">{{product.brand}}</div>
+          <div class="single-line-ellipsis">{{product.name}}</div>
+          <div>${{product.cost}}</div>
+          <div class="view-options">+ More Colors</div>
+      </div>
+    </span>
 </div>
--- a/apps/foundation-app/src/app/product/components/product-list/product-list.component.scss
+++ b/apps/foundation-app/src/app/product/components/product-list/product-list.component.scss
@@ -1,3 +1,54 @@
-.product-container {
-  color: black;
+.card-container {
+  margin: 7% 2%;
+  text-align: center;
+  font-size: 12px;
+  
+  .card {
+    width: 45%;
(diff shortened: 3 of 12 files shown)


apps/foundation-app/package.json
```diff
@@ -0,0 +1,10 @@
+{
```
damienwebdev: Can we add some documentation here about running the foundation-app? I ran into some problems when I ran `ng s --app=foundation-demo` after pulling down for the first time. Since it was missing the foundation sites for the node_modules. Perhaps this is where we use `lerna` to install the child dependencies when we run npm install? Or maybe, we keep the dependencies in the top-level package.json?

apps/foundation-app/src/app/product/components/product-list/product-list.component.html
```diff
@@ -1,4 +1,11 @@
-<div class="product-container" *ngFor="let product of products">
-  {{product.id}}
-  {{product.cost}}
+<div class="card-container">
```
damienwebdev: This is probably two components, the `product-list` and the `product-card`. I'd probably add  another component to further separate. That way, once we add pagination, we can keep the concerns separate.

damienwebdev: Additionally, these class names are a bit imprecise. Luckily in Angular, we have ViewEncapsulation, but let's assume for some reason, we need to turn it off in the future. It would be painful if we happen to need a card/card-container class.

I think like `product-list` and `product-card` in a `BEM` style are good. http://getbem.com/naming/

damienwebdev: On top of that, I'd consider grabbing some sample data from https://github.com/magento/magento2-sample-data/tree/2.3-develop/pub/media to use for the images.

apps/foundation-app/src/app/product/components/product-list/product-list.component.scss
```diff
@@ -1,3 +1,54 @@
-.product-container {
-  color: black;
-}
\ No newline at end of file
+.card-container {
```
damienwebdev: See my earlier comments about `product-list` and `product-card` for naming conventions. I'd also consider switching this over to `display:grid` as that makes this code significantly less complex to read. We should also come up with a common `media` query mixin and some screen size variables that we want to use for daffodil.

apps/foundation-app/src/styles.scss
```diff
@@ -1 +1,8 @@
 /* You can add global styles to this file, and also import other style files */
+@import '../node_modules/foundation-sites/assets/foundation';
+
+.single-line-ellipsis {
```
damienwebdev: This is neat!

libs/core/src/product/testing/factories/product.factory.ts
```diff
@@ -6,11 +6,33 @@ import { Product } from '../../model/product';
 export class ProductFactory {
   
   create() : Product {
-    return {...new MockProduct()};
+    return {...new MockProductShortNames()};
+  }
+
+  createStyleTestingList() : Product[] {
```
damienwebdev: Love it.

apps/foundation-app/package.json
lderrickable: I added some stuff to lerna.json so it will install all dependencies in core and foundation-app with the `lerna bootstrap` command. I still need to figure out how lerna and publishing to npm works, but I'll do that in a separate branch right after this probably. Added documentation too

PR #14, opened 2018-05-14
View cart

lderrickable:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information


lderrickable:
With regard to the `general` comments, a lot of the elements that needed to be bold didn't require any other styling, so they didn't need a class name. That's why I decided to go with a class called `embolden` rather than a mixin. I don't see the difference between having mixins floating around and having classes floating around. Maybe there is one? Do you still want me to change them to mixins?

lderrickable:
"What is the reasoning behind proceed-to-checkout being its own component?"
Because I needed to use it in two places.

"Why is proceed-to-checkout inside both cart-totals and cart-view"
are you asking this from a design perspective or a coding perspective? Sounds like design

lderrickable:
"Cart Total looks like it's a wrapper component for cart-total and cart-totals, I'd consider separating into smaller components"
Not sure what you mean by "cart-totals". The only possibility to break it down further that I can see would be something like:
```
<cart-summary>
     <promotion></promotion>
     <cart-total></cart-total>
     <proceed-to-checkout></proceed-to-checkout>
</cart-summary>
```
Is that what you mean?

xelaint:
## General

- [ ] Use 'rem' for font related styles (font-size, line-height, etc). Avoid percentage since the font-size should be relative to the root element.
- [ ] Use 'px' for padding/margin.
- [ ] Why did you decide to separate the screen breakpoints into its own section rather than keeping it nested inside their class? I think it's a bit confusing since I expect to see all the styles of a class within the same block of code.
- [ ] I think the classes that still use a__b__c can be refactored to just being a__b. If __c is needed, it should probably be its own child block as Damien pointed out.

## Qty Dropdown / Input
- [ ] Update `input` styles to match `selector`styles (padding, margin, etc). Selector has default padding styles from foundation but input doesn't so they don't match up.

## Cart Item
- [ ] Instead of using `cart-item__change__remove` on the remove text, you could do:
```
<div class="cart-item__change">
      <span>Edit</span>
      <span>Remove</span>
</div>
```

```
&__change {
    color: $gray;
    margin-top: 15px;

    & > span {
      text-decoration: underline;
margin-right: 10px;
&:last-child {
margin-right: 0;
}
    }
  }
```

apps/foundation-app/src/app/cart/components/cart-item/cart-item.component.html
```diff
@@ -0,0 +1,25 @@
+<div class="cart-item">
+  <img src="/assets/mh04-green_alt1.jpg">
+  <div class="cart-item__description">
+    <div class="cart-item__description__brand">Brand (HC)</div>
+    <div class="cart-item__description__name embolden">{{item.name}}</div>
```
damienwebdev: My reason for the "embolden" mixin over a class is that it's a lot easier to think about a "class" of an element rather than all of the classes that an element is. For me, at the very least, I think it's just a bit more concise. I don't have to go searching through all the different classes when something goes wrong. I have one class which utilizes several mixins.

apps/foundation-app/src/app/design/molecules/qty-dropdown/qty-dropdown.component.scss
```diff
@@ -0,0 +1,20 @@
...
+    width: 60px;
+
+    &__selector {
+      border-color: $medium-gray;
+      font-size: $global-font-size;
+      height: 35px;
+      padding-top: 5px;
+    }
+
+    &__input {
```
xelaint: update styles to watch with `selector`

apps/foundation-app/src/app/cart/components/cart-item/cart-item.component.html
lderrickable: I made "embolden" into a class and a mixin like we discussed.

xelaint:
@lderrickable I still see the screen breakpoints broken down in its own sections i.e. `cart-item`, `cart`

apps/foundation-app/src/app/design/molecules/qty-dropdown/qty-dropdown.component.scss
lderrickable: done

apps/foundation-app/src/app/design/molecules/qty-dropdown/qty-dropdown.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  styleUrls: ['./qty-dropdown.component.scss']
+})
+export class QtyDropdownComponent implements OnInit {
+
+  @Input() qty: number;
+  @Input() id: number;
+
+  static readonly dropdownRange = 9;
+  dropdownItemCounter: number[];
+  inputHasBeenShown: boolean;
```
lderrickable: I added this boolean, because when the `<input>` is edited, the qty element would be immediately changed back into a `<select>` when the user would type the tens digit of anything over ten.

apps/foundation-app/src/app/design/molecules/qty-dropdown/qty-dropdown.component.ts
```diff
@@ -0,0 +1,74 @@
...
+    if (!this.isQtyOutsideDropdownRange() && !this.inputHasBeenShown) {
+      return false
+    } else {
+      this.inputHasBeenShown = true;
+      return true;
+    }
+  }
+
+  onChangedWrapper(value: any) {
+    value = parseInt(value);
```
lderrickable: I needed to parseInt here, because `onChangedWrapper` seems only to pass in strings, even though the `qty` that it is passing in is of type `number`

damienwebdev:
@xelaint To your earlier point about `proceed to checkout` that actually will have some kind of encapsulated functionality to it (routing/analytics events) so it makes sense to keep that as it's own component.

apps/foundation-app/src/app/design/molecules/qty-dropdown/qty-dropdown.component.ts
damienwebdev: Ah, that's an interesting problem.... maybe we can do a brief timeout before the check happens with `rxjs` that way we give the user some time before it switches, I think 300ms would be sufficient?

apps/foundation-app/src/app/cart/components/cart/cart.component.html
```diff
@@ -1,3 +1,7 @@
-<div class="cart-container">
-  {{cart.id}}
+<div class="cart">
```
damienwebdev: - [ ] What should this look like when the cart is empty? @xelaint let's see if we can't do a simple animation here.

- [ ] What about when we're loading cart items? @xelaint @lderrickable I'd determine whether or not there's a foundation loader, and use that. Otherwise, if we have to roll our own, let's pick one early, because we're definitely going to need it

apps/foundation-app/src/app/cart/components/proceed-to-checkout/proceed-to-checkout.component.html
```diff
@@ -0,0 +1 @@
+<div class="button">proceed to checkout</div>
```
xelaint: @lderrickable Even though we're keeping the component, I think we can still use foundation's button styles and eliminate the duplicate styling?

apps/foundation-app/src/app/cart/components/cart-summary/cart-summary.component.html
```diff
@@ -0,0 +1,9 @@
+<div class="cart-summary">
+  <div class="cart-summary__title">cart summary</div>
+  <promotion></promotion>
+  <cart-subtotal [cart]="cart"></cart-subtotal>
```
damienwebdev: I'd say each of these should be a `cart-total`, I don't think we need a unique component for each of these. I think the API for a `cart-total` is `title` and `value`.

apps/foundation-app/src/app/cart/components/cart/cart.component.html
xelaint: @lderrickable added design to zeplin for `empty cart view`.

apps/foundation-app/src/app/cart/components/proceed-to-checkout/proceed-to-checkout.component.html
lderrickable: Oh, I'm planning on changing it. I just ran out of time

apps/foundation-app/src/app/design/molecules/qty-dropdown/qty-dropdown.component.ts
lderrickable: I was thinking about doing that, but then I thought of myself as a user going: "Well, we want at least ten (types in "1"), but how many...WTF!" So that's why I ultimately went with this. What do you think?

apps/foundation-app/src/app/cart/components/cart/cart.component.html
lderrickable: I think we're best off with font-awesome for the loader: https://fontawesome.com/icons/spinner?style=solid, but @xelaint can make the final call on that. Foundation doesn't have one

xelaint: Yes, the font-awesome loader works.

apps/foundation-app/src/app/cart/components/proceed-to-checkout/proceed-to-checkout.component.html
xelaint: Okay. Thanks!

lderrickable: Oh, I just saw Damien's comment. But anyway, it turned out that there were some styling conflicts between the foundation button class and the class I made that toggled `display: none`, so it would've needed to be wrapped anyway

lderrickable: Ok, I made the styling changes

lderrickable:
@xelaint so it turned out that having `grid-area` style definitions outside of the same block as the `grid-template-area` definition caused the browser to give a warning. So I had to place those styles all in the same media query block. All the other styles are as you requested though.

apps/foundation-app/src/app/cart/components/cart-summary/cart-summary.component.html
lderrickable: Ok, so I made this into two components, `cart-subtotal` and `cart-grand-total`, which both implement a common interface, `cart-total`. The only difference is that `cart-grand-total` has some blue font.

apps/foundation-app/src/app/cart/components/cart/cart.component.html
lderrickable: added loading spinner for cart

apps/foundation-app/src/app/design/molecules/qty-dropdown/qty-dropdown.component.ts
damienwebdev: I think this solution works for now.

xelaint:
@lderrickable can you embolden the value for `grand-total`? lgtm after that!

PR #19, opened 2018-05-30
View product

lderrickable:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[x] Yes
[ ] No
```

The model of Product.cost was changed to Product.price for naming consistency between Product and Cart.
## Other information


apps/foundation-app/src/app/design/molecules/image-gallery/image-gallery.component.scss
```diff
@@ -0,0 +1,60 @@
+@import '~scss/styles';
+
+// parent element must have a margin-bottom of 100vw;
```
lderrickable: I did some thinking about this comment I added to see if I could change it to "parent element must have a margin-bottom of its own width", but, as image-gallery stands now, I don't think it's possible. So basically, image-gallery can only be used when the active image is supposed to be the width of the screen. If image-gallery only needs to be used for galleries of images on mobile, I think this is ok. But I can think of ways to make the image-gallery adhere to the width of its parent, though it would require javascript and some document.getElementByClassNames and at least one passed in variable from the parent.

apps/foundation-app/src/app/design/molecules/image-gallery/image-gallery.component.scss
```diff
@@ -0,0 +1,59 @@
...
+.image-gallery {
```
damienwebdev: So, I think ideally we'll want one main image, and a few smaller images. So, I think functionally, what I'd expect here would be a small set of images that I can click on to switch the main image out. So, I think this nicely fits into two separate components: 
1. `image-list`
2. `image-gallery`

An `image-gallery` is composed of a main `img` and an `image-list`. Selecting on an item in the `image-list` makes that specific `image` the main image, and also applies some kind of distinction that the specific image in the list is selected as the main image (could be a border, or a icon, etc).

apps/foundation-app/src/app/product/components/product/product.component.html
```diff
@@ -1,4 +1,24 @@
...
+<div class="product">
+  <div class="product__description-container">
+    <div class="product__brand">{{product.brand}}</div>
+    <div class="product__name">{{product.name}}</div>
+    <div class="product__price">${{product.price}}</div>
+  </div>
+  <div class="product__image-gallery-wrapper">
+    <image-gallery></image-gallery>
+  </div>
+  <div class="product__images-wrapper">
```
damienwebdev: I think the functionality should remain the same between desktop and mobile, this will reduce code complexity long term and keep ux more consistent. @xelaint what do you think?

.angular-cli.json
```diff
@@ -56,6 +56,32 @@
       },
       "tags": []
     },
+    {
```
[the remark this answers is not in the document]
lderrickable: It isn't necessary, but preferred. We hadn't thought about what the workflow for development in @daffodil/core would actually be like. Currently looks like this:
1. Makes a change in @daffodil/core.
2. Stop the app.
3. run "npm run-scripts build:lib"
4. Restart the foundation-demo app.

With this change, it looks like this:
1. Makes a change in @daffodil/core.
2. Wait a few seconds for the app to reload.

It's about a 20-30 second difference for every change. Maybe more. The only comment I left in your angular6 pull request (https://github.com/graycoreio/daffodil/pull/18#pullrequestreview-124444895) was actually about this.

apps/foundation-app/src/app/app-routing.module.ts
```diff
@@ -15,6 +16,10 @@ export const appRoutes: Routes = [
   {
     path: 'cart',
     component: CartViewComponent
+  },
+  {
+    path: 'product/:id',
```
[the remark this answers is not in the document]
lderrickable: Added a 404 page. I took some design liberties, so feel free to shoot it down if you don't like it. @damienwebdev @xelaint

apps/foundation-app/src/app/product/components/product-card/product-card.component.html
```diff
@@ -7,7 +7,7 @@
         <div class="product-card__name">
             <span class="single-line-ellipsis">{{product.name}}</span>
         </div>
-        <div class="product-card__price">${{product.cost}}</div>
+        <div class="product-card__price">${{product.price}}</div>
```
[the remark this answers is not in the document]
lderrickable: added

apps/foundation-app/src/app/design/molecules/image-gallery/image-gallery.component.scss
lderrickable: done. I made the distinction between selected and deselected images be a difference in opacity.

xelaint:
## Image Gallery
- [ ] For the image gallery/image list, is there a way to contain the image to a specific width and height? (100x100)

PR #21, opened 2018-06-08
Add product to cart

lderrickable:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information


lderrickable:
This branch is dependent on https://github.com/graycoreio/daffodil/pull/19. @xelaint can review it, but @damienwebdev should wait till https://github.com/graycoreio/daffodil/pull/19 is pushed to develop

xelaint:
I'd like a message + cart summary to show when you click on `add to cart` with the option of `view cart` and `checkout` (rather than re-directing to the cart page automatically), but that means we need to add a header/nav bar to the site. We will need to work that out first before we can update this branch, @damienwebdev agree?

damienwebdev:
@lderrickable I'd agree to that. Additionally, I do know that there are several settings in Magento which allow business users to determine how this works (I generally think it's pointless, but there may be a valid use case for it), so while we may just make a decision now, we will have to revisit this later.

@xelaint to achieve that use-case I think we'd need several other components:
- Nav
- Minicart (maybe a reutilized cart-summary would be neat here?)
- Notifications

Could you open up an issue and label as a feature request? The "notifications" should be a fairly simple component that we could probably utilize as a "good first issue". The others will probably have a bit more depth to them.

apps/foundation-demo/src/app/product/components/add-to-cart/add-to-cart.component.ts
```diff
@@ -0,0 +1,29 @@
...
+    this.emitAddToCart();
+    this.redirectToCart();
+  }
+
+  emitAddToCart() {
+    this.addToCart.emit();
+  }
+
+  redirectToCart() {
+    this.router.navigateByUrl('/cart');
```
damienwebdev: Move this to an effect. This is a clear instance of a side-effect that shouldn't be known about by this component.

apps/foundation-demo/src/app/product/components/product/product.component.ts
```diff
@@ -11,12 +11,17 @@ import { Image } from '../../../design/interfaces/image';
 })
 export class ProductComponent implements OnInit {
   @Input() product: Product;
-
+  @Output() addToCart: EventEmitter<any> = new EventEmitter();
+  
+  qty: number;
```
damienwebdev: You're storing a copy of local state here. Can we store this in redux?

libs/core/src/cart/testing/factories/cart.factory.ts
```diff
@@ -8,9 +8,23 @@ import { CartShippingRate } from '../../model/cart-shipping-rate';
...
   create() : Cart {
-    return {...new MockCart()};
+    return this.cart;
+  }
+
+  createCartItem() : CartItem {
+    return new MockCartItem();
+  }
+
+  addCartItemToCart(reqBody) {
```
damienwebdev: This doesn't make sense as a factory method. What's the purpose of this method?

libs/core/src/product/containers/product/product.component.ts
```diff
@@ -35,4 +36,8 @@ export class ProductContainer implements OnInit {
       select(fromProduct.selectSelectedProduct)
     );
   }
+
+  addToCart(payload) {
```
damienwebdev: This isn't the responsibility of "product". I'd argue you'd probably want to keep all "cart" related things in cart. It doesn't make sense to breach the boundaries of these two distinctly separate modules.

libs/core/src/cart/testing/factories/cart.factory.ts
lderrickable: The purpose was to simulate a backend that would be responsible for adding a cartItem to the cart, then returning the new cart. Should this be in the testing service instead?

damienwebdev: @lderrickable I'd say yes. The factory's responsibility is for generating new instances of an object easily. Currently, you're utilizing it to maintain state. If we could make the class readonly, that would enforce the condition I'm describing.

libs/core/src/product/containers/product/product.component.ts
lderrickable: Been thinking about how else to do this, and I'm not sure how exactly to do it if it isn't here. The problem I'm running into is that I can only have one attribute component per `div`. So I can't just add it to the daffodil/core cart-container, and do something like this:

```
<div class="product-container"
  product-container #ProductContainer="ProductContainer" [selectedProductId]="productId"
  cart-container #CartContainer="CartContainer">
    <product [product]="ProductContainer.product$ | async" (addToCart)="CartContainer.addToCart($event)" *ngIf="!(ProductContainer.loading$ | async)"></product>
    <div *ngIf="(ProductContainer.loading$ | async)" class="product-container__loading-icon">
      <i class="fa fa-spinner fa-spin fa-5x"></i>
    </div>
</div>
```

This could work:
```
<div class="product-container" product-container #ProductContainer="ProductContainer" [selectedProductId]="productId">
  <div class="cart-container" cart-container #CartContainer="CartContainer">
    <product [product]="ProductContainer.product$ | async" (addToCart)="CartContainer.addToCart($event)" *ngIf="!(ProductContainer.loading$ | async)"></product>
    <div *ngIf="(ProductContainer.loading$ | async)" class="product-container__loading-icon">
      <i class="fa fa-spinner fa-spin fa-5x"></i>
    </div>
  </div>
</div>
```
but yikes.

I could just have a service in daffodil/core that dispatches the action, but I'd still need to import that service into the product-container, which is basically the same thing with an extra step. This definitely seems like something you should be able to do in Angular, but it seems that everything I think to try, I eventually get to a roadblock that basically says, "that's not how angular works". Any ideas?

apps/foundation-demo/src/app/product/components/add-to-cart/add-to-cart.component.ts
lderrickable: Should this effect be in the foundation-demo? Or daffodil/core? At first I thought daffodil/core, but daffodil/core currently doesn't know anything about any urls or navigation, so we'd be defining what the user's urls need to be, even though they'll be the ones writing out the routing. I guess it could go in foundation-demo, but that seemed weird to me. Can't really think of a good reason that it shouldn't go in foundation-demo though, since we won't really be adding extra dependencies. So it should be added to foundation-demo?

apps/foundation-demo/src/app/product/components/product/product.component.ts
lderrickable: done

apps/foundation-demo/src/app/product/components/add-to-cart/add-to-cart.component.ts
damienwebdev: I'd argue that effect should be in foundation-demo. This should be a hook into the AddToCart success action.

libs/core/src/product/containers/product/product.component.ts
damienwebdev: So, I think this may be a case for some nifty transclusion. The `product` component doesn't need to be aware of "AddToCart". That's not its responsibility. So, I'd migrate the component out, yet allow it to transclude something that has the `add-to-cart` attribute on it.

apps/foundation-demo/src/app/product/components/add-to-cart/add-to-cart.component.ts
lderrickable: done

libs/core/src/cart/testing/factories/cart.factory.ts
lderrickable: done

libs/core/src/product/containers/product/product.component.ts
lderrickable: done. I only did the bare minimum here. I'll need to add some other stuff if I ever need to transclude anything else, but I figure only add what I need

PR #53, opened 2018-08-23
Header

lderrickable:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [X] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/master/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[X] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## Does this PR introduce a breaking change?
```
[ ] Yes
[X] No
```

Issue #30, opened 2018-06-21
feat(header, nav-bar): Header containing logo, nav-bar, and sub-nav.


xelaint:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Current behavior
A header/nav does not exist for the demo. Additionally, when a product is added to cart, there is no proper place to display a minicart or notification.

## Expected behavior
Header for Daffodil sample store. Header includes logo, navbar with basic nav items, and subnav (sign in, account, cart icons)

Mock + Prototype: https://framer.cloud/aFzwn

## What is the motivation / use case for changing the behavior?
General page structuring and UX Experience Improvement

## Environment

<pre><code>
Daffodil version: X.Y.Z

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

damienwebdev:
@xelaint Can you clarify this feature more into a precise spec? To say "nav" is a fairly broad term which can mean many things. Do you have a sample for us to start with as an initial implementation? I was thinking something along the lines of the sidebar on https://www.walmart.com/?

xelaint:
@damienwebdev updated this feature to only include the header and navbar. I noticed that the styling on certain pages are added without consideration of the header, page container, etc. I will come up with a mock asap so we can get this implementation going.

xelaint:
@damienwebdev Mock + Prototype: https://framer.cloud/aFzwn

damienwebdev:
Resolved in #53

PR #65, opened 2018-09-14
Header style updates

xelaint:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[x] Code style update (formatting, local variables)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: Updates to #53 #30 

## What is the new behavior?
Cherry pick commits from previous branch. I stuck with the sidebar list because it has different styles than `daff-list`.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/apps/foundation-demo/src/app/core/sidebar/containers/sidebar/sidebar.component.html
+++ b/apps/foundation-demo/src/app/core/sidebar/containers/sidebar/sidebar.component.html
@@ -1,9 +1,9 @@
 <daff-sidebar class="sidebar" (escapePressed)="onClose()">
   <span class="sidebar__close" (click)="onClose()"><i class="fa fa-times"></i></span>
-  <ul class="sidebar__list">
-    <li><a daff-sidebar-item (click)="onClose()" routerLink="/">Women's</a></li>
-    <li><a daff-sidebar-item (click)="onClose()" routerLink="/">Men's</a></li>
-    <li><a daff-sidebar-item (click)="onClose()" routerLink="/">Accessories</a></li>
-    <li><a daff-sidebar-item (click)="onClose()" routerLink="/">Sale</a></li>
-  </ul>
+  <div class="sidebar__list">
+    <span><a daff-sidebar-item (click)="onClose()" routerLink="/">Women's</a></span>
+    <span><a daff-sidebar-item (click)="onClose()" routerLink="/">Men's</a></span>
+    <span><a daff-sidebar-item (click)="onClose()" routerLink="/">Accessories</a></span>
+    <span><a daff-sidebar-item (click)="onClose()" routerLink="/">Sale</a></span>
+  </div>
 </daff-sidebar>
--- a/apps/foundation-demo/src/app/core/sidebar/containers/sidebar/sidebar.component.html
+++ b/apps/foundation-demo/src/app/core/sidebar/containers/sidebar/sidebar.component.html
@@ -1,7 +1,9 @@
 <daff-sidebar class="sidebar" (escapePressed)="onClose()">
-  <span class="sidebar__close" (click)="onClose()"><i class="fas fa-times"></i></span>
-  <a daff-sidebar-item (click)="onClose()" routerLink="/">women's</a>
-  <a daff-sidebar-item (click)="onClose()" routerLink="/">men's</a>
-  <a daff-sidebar-item (click)="onClose()" routerLink="/">accessories</a>
-  <a daff-sidebar-item (click)="onClose()" routerLink="/">sale</a>
+  <span class="sidebar__close" (click)="onClose()"><i class="fa fa-times"></i></span>
+  <ul class="sidebar__list">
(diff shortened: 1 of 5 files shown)


apps/foundation-demo/src/app/core/sidebar/containers/sidebar/sidebar.component.html
```diff
@@ -1,7 +1,9 @@
 <daff-sidebar class="sidebar" (escapePressed)="onClose()">
-  <span class="sidebar__close" (click)="onClose()"><i class="fas fa-times"></i></span>
-  <a daff-sidebar-item (click)="onClose()" routerLink="/">women's</a>
-  <a daff-sidebar-item (click)="onClose()" routerLink="/">men's</a>
-  <a daff-sidebar-item (click)="onClose()" routerLink="/">accessories</a>
-  <a daff-sidebar-item (click)="onClose()" routerLink="/">sale</a>
+  <span class="sidebar__close" (click)="onClose()"><i class="fa fa-times"></i></span>
+  <ul class="sidebar__list">
```
lderrickable: This should have a daff-list attribute:
```
<ul daff-list class="sidebar__list">
```

apps/foundation-demo/src/app/core/sidebar/containers/sidebar/sidebar.component.scss
```diff
@@ -1,17 +1,20 @@
...
     font-size: $medium-font-size;
     cursor: pointer;
-    line-height: 0;
-    height: 25px;
+  }
+
+  &__list {
+    padding: 0;
+    margin: 0;
+    list-style: none;
```
lderrickable: Once you add the `daff-list` attribute, you shouldn't need some, if not all, of these styles.

apps/foundation-demo/src/app/core/sidebar/containers/sidebar/sidebar.component.html
xelaint: `I stuck with the sidebar list because it has different styles than daff-list.`

damienwebdev:
@xelaint Can you amend your commit messages to following the [commit message guidelines](https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit)?

apps/foundation-demo/src/app/core/sidebar/containers/sidebar/sidebar.component.scss
damienwebdev: So, I closed this thinking it made sense to use the `sidebar__list` since the styles are different. I wonder though, if the sidebar needs `ul`/`li` here in the first place. I can't explicitly state a reason for needing it beyond simple "convention".

xelaint: I guess it doesn't need to be. the `ul/li` can be reserved for future use when/if there are flyout/sub-list items to the main list items.

damienwebdev: Do we need a ul and li then as well? I can't really see a use for it.

xelaint: I guess not. In this case, the footer links don't need to be `ul/li` either?

lderrickable: I don't see why anything ever needs `ul/li`. There might be a reason. I've just never been curious what that reason could be.

lderrickable: Just looked it up. Seems to be just to make HTML a bit more readable. So I'd say that because the footer is composed of attribute components that convey it is a list, there's no need to also use `ul/li`. But I don't really think it matters that much.

lderrickable: Also, this is the second time we're making a list in foundation-demo. If we're already not using the `daff-list` component because the styling doesn't work, doesn't that mean it might be kind of useless?

xelaint: Looked it up and it's becoming more common to not use `ul/li` to increase readability, so I changed it to `div/span`.

And yes, `daff-list` is not needed unless we are thinking about lists for product descriptions, shipping information, etc. That would definitely be styled differently than what we have now since in that case, the bullets would remain.

damienwebdev: So, I'm not sure of this new API is any better. To me it seems just as verbose, merely swapping elements doesn't change much. I'd much prefer something like...

```
<daff-sidebar>
    <a daff-sidebar-item></a>
</daff-sidebar>
```

PR #71, opened 2018-09-20
Separate Backdrop from Sidebar

lderrickable:
## PR Checklist

I didn't actually need to do this for the cart notification, but I'd already done it because I thought I needed to make a modal. So we won't need to do this again in the future when we get to modals.

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information


apps/foundation-demo/src/app/design/molecules/backdrop/backdrop/backdrop.component.scss
```diff
@@ -0,0 +1,21 @@
+$daff-sidebar-backdrop-z-index: 3;
+
+.daff-backdrop {
+  position: absolute;
+  width: 100%;
+  height: 100%;
+  background: rgba(0,0,0,0.3);
+  z-index: $daff-sidebar-backdrop-z-index;
```
damienwebdev: I don't think this should be here. Can we move the z-index and positioning back into `daff-sidebar`?

lderrickable: I moved the z-index out, but I think the positioning needs to stay with the width and height declarations. And those need to stay in the backdrop, so it can control when it is displayed. Moving position to the sidebar breaks the backdrop completely. Moving position, width, and height just places an invisible blanket over the whole page.

PR #95, opened 2018-10-15
Update footer and newsletter styles

xelaint:
## PR Checklist

## PR Type

```
[x] Code style update (formatting, local variables)
```

Style issues with footer and newsletter due to removal of padding on `daff-container`. `daff-newsletter` is not a design package.

Issue Number: N/A

## What is the new behavior?
Update styles to footer and newsletter to reflect change from `daff-container`. Remove `daff-` from newsletter module and component.

```
[ ] Yes
[x] No
```

## Other information


apps/foundation-demo/src/app/core/footer/footer.component.scss
```diff
@@ -2,18 +2,20 @@
...
+    padding: 75px 50px;
   }
 
   @include breakpoint(large) {
-    margin: 0 150px 100px 200px;
     grid-template-columns: repeat(4, 1fr);
   }
+
+  @include breakpoint(xlarge) {
+    padding: 75px 0;
```
lderrickable: Are you sure you want no right or left padding? It looks off to me at anything wider than 1200px.

lderrickable: same with the newsletter styles

xelaint: Yes, I don't want left/right padding. It'll just contain the content more than the specified `daff-container` size. I only added padding to the smaller screen sizes so the content won't be edge to edge.

lderrickable: But because footer and newsletter are in `template.html`, doesn't that mean that they'll never be affected by `daff-container`? Because `daff-container` will wrap the `<ng-router></ng-router>` in:
```
<div class="template">
    <sidebar-viewport-container>
        <header-container></header-container>
        <router-outlet></router-outlet>
        <daff-newsletter></daff-newsletter>
        <footer></footer>
    </sidebar-viewport-container>
</div>
```
? Unless `template` will be wrapped in a `daff-container` instead of it's children.

xelaint: If you check the `newsletter.component.html` and `footer.component.html` file, you'll see that I wrapped the content in both with `daff-container`.

xelaint: If you check the `newsletter.component.html` and `footer.component.html` file, you'll see that I wrapped the content in both with `daff-container`.

lderrickable: oh gotcha. I just take issue with the `daff-container` then. When the width is 1200px, there's no margin or padding left or right. But if that's ok with you, then it's ok with me

xelaint:
@lderrickable I added a new commit to fix what you were talking about! I didn't realize the problem until Damien and I were going over a rebase since I've been checking the screen size by toggling Chrome's sizes rather than dragging the page back and forth to view it at all sizes.

Issue #113, opened 2018-10-30
Feature Request: Theme Palette mixin to augment a directive with a color property


xelaint:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

## Current behavior
Can't style buttons with color variants.

## Expected behavior
Theme palette mixin to augment a directive with a color property. Be able to style buttons or any other components using a color directive.

## Minimal reproduction of the problem with instructions
```
&.daff-primary {
  background: $primary;
  color: $white;

  &:hover {
    background: $daff-color($daff-primary, 600);
  }
}
```

## What is the motivation / use case for changing the behavior?
Add color variants to components.

## Environment

<pre><code>
Daffodil version: X.Y.Z

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

PR #114, opened 2018-11-01
Add `Colorable` and `daffColorableMixin` to support coloring components based on the palette.

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, you have to manually apply button colors.

Issue Number: #113 

## What is the new behavior?
You can now apply the mixin to the component class and get colors~
Fixes: #113 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/libs/design/src/atoms/button/button.component.spec.ts
+++ b/libs/design/src/atoms/button/button.component.spec.ts
@@ -3,20 +3,23 @@ import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { DaffButtonComponent } from './button.component';
 import { By } from '@angular/platform-browser';
+import { DaffPalette } from '../../core/colorable/colorable';
 
 @Component({
   template: `
-    <a daff-button>Link Daff Button</a>
-    <a daff-clear-button>Link Daff Clear Button</a>
-    <a daff-stroked-button>Link Daff Stroked Button</a>
-    <a daff-icon-button>Link Daff Icon Button</a>
-    <button daff-button>Button Daff Button</button>
-    <button daff-clear-button>Button Daff Clear Button</button>
-    <button daff-stroked-button>Button Daff Stroked Button</button>
-    <button daff-icon-button>Button Daff Icon Button</button>
+    <a daff-button [color]="color">Link Daff Button</a>
+    <a daff-clear-button [color]="color">Link Daff Clear Button</a>
+    <a daff-stroked-button [color]="color">Link Daff Stroked Button</a>
+    <a daff-icon-button [color]="color">Link Daff Icon Button</a>
+    <button daff-button [color]="color">Button Daff Button</button>
+    <button daff-clear-button [color]="color">Button Daff Clear Button</button>
+    <button daff-stroked-button [color]="color">Button Daff Stroked Button</button>
+    <button daff-icon-button [color]="color">Button Daff Icon Button</button>
   `
 })
-class TestButtonWrapper {}
+class TestButtonWrapper {
+  color: DaffPalette;
+}
 
 describe('DaffButtonComponent', () => {
   let component: TestButtonWrapper;
@@ -73,4 +76,14 @@ describe('DaffButtonComponent', () => {
       expect(fixture.debugElement.query(By.css('button[daff-icon-button]')).nativeElement.classList.contains('daff-icon-button')).toEqual(true);
     });
   }); 
+
+  describe('using a colored variant of a button',() => {
+    it('should set a color class on the button', () => {
+      component.color = "primary";
+      fixture.detectChanges();
+      
+      let buttonDE = fixture.debugElement.query(By.css('button[daff-button]'));
+      expect(buttonDE.nativeElement.classList.contains('daff-primary')).toEqual(true);
+    });
+  });
 });
--- a/libs/design/src/core/colorable/colorable.spec.ts
+++ b/libs/design/src/core/colorable/colorable.spec.ts
@@ -1,0 +1,138 @@
+import { daffColorMixin, colorInPalette } from "./colorable";
+import { ElementRef } from "@angular/core";
+
+/**
+ * Some of these tests are inspired by the @angular/material
+ * color common-behavior, as it is a really clean implementation.
+ */
+describe('daffColorMixin', () => {
+  let instance;
+  let classWithColor;
+
+  beforeEach(() => {
+    classWithColor = daffColorMixin(TestingClass);
+    instance = new classWithColor();
+  });
+
+  it('should add a color property to an existing class', () => {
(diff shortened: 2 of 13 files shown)


libs/design/src/atoms/button/button.component.spec.ts
```diff
@@ -73,4 +76,14 @@ describe('DaffButtonComponent', () => {
       expect(fixture.debugElement.query(By.css('button[daff-icon-button]')).nativeElement.classList.contains('daff-icon-button')).toEqual(true);
     });
   }); 
+
+  fdescribe('using a colored variant of a button',() => {
```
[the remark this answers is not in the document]
damienwebdev: resolved in 7636bdbf20cdc4116befbc09e4dcdc564e6ace08

libs/design/src/core/constructor.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * A basic constructor type useful for mixins
+ * See https://blog.mariusschulz.com/2017/05/26/typescript-2-2-mixin-classes
+ * for a really explanation of why mixins are useful.
```
lderrickable: typo - *really good explanation

libs/design/src/core/colorable/colorable.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    });
+
+
+    it('should set a namespaced color class if a color is specified', () => {
+        instance.color = "primary";
+
+        expect(instance.element.classList).toContain("daff-primary");
+    });
+
+    it('should default to no color class one is not specified', () => {
```
lderrickable: typo - *if

libs/design/src/core/colorable/colorable.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+describe('daffColorMixin', () => {
+    let instance;
+    let classWithColor;
+
+    beforeEach(() => {
+        classWithColor = daffColorMixin(TestingClass);
+        instance = new classWithColor();
+    });
+
+    it('should add a color property to an existing class', () => {
```
lderrickable: I think this test can be removed

libs/design/src/core/colorable/colorable.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * In order to be colorable, our class must implement these two properties
+ */
+export interface DaffColorable {
+    color: DaffPalette;
+}
+
+/**
+ * This is the valid options that can be passed to a DaffColorable component
```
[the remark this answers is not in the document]
xelaint: Resolved in 6dd93ef

libs/design/src/core/colorable/colorable.ts
```diff
@@ -0,0 +1,69 @@
+import { ElementRef, Type } from "@angular/core";
+import { Constructor } from '../constructor';
+
+/**
+ * In order to be colorable, our class must implement these two properties
```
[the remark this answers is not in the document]
xelaint: Resolved in 6dd93ef

libs/design/src/core/constructor.ts
xelaint: Resolved in 6dd93ef

libs/design/src/core/colorable/colorable.spec.ts
xelaint: Resolved in 5e336aa

libs/design/src/core/colorable/colorable.ts
```diff
@@ -0,0 +1,69 @@
...
+        _color: DaffPalette;
+
+        get color(): DaffPalette{return this._color;}
+        set color(value: DaffPalette) {
+            //Handle the default color
+            const incomingColor = value || defaultColor;
+
+            if(
+                incomingColor !== undefined && 
+                !(<any>Object).values(DaffPaletteEnum).includes(incomingColor)) {
```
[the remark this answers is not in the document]
damienwebdev: @lderrickable I fixed this in e3a265b45d49331fa28c10fc219b1a535f058b31

libs/design/src/core/colorable/colorable.ts
```diff
@@ -0,0 +1,69 @@
...
+                throw new TypeError(incomingColor + " is not a valid color for the DaffPalette");
+            }
+
+            if(incomingColor !== this._color){ //Only run the dom-render if a change occurs
+                //Remove the old color
+                if(this._color){
+                    this._elementRef.nativeElement.classList.remove(`daff-${this._color}`);
+                }
+
+                if(incomingColor){
```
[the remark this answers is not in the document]
damienwebdev: I've added test cases to cover this in e8a0f0a04800a92c48e8de99661cac6ed1eb4b1d

libs/design/src/core/colorable/colorable.spec.ts
damienwebdev: I disagree. I added this to ensure that the mixin actually sets a class on the consuming class in order to ensure that it meets the interface. Mixins seem really janky to me, and this test makes me feel comfortable that the mixin is doing what I expect.

lderrickable: But you're not testing for a class here. That's in the following test. Here, you're just setting a property directly on the instance, then checking the instance to see if that property is still set. Or are you testing that setting a color that exists as an possible color choice on that instance should set that color on the instance? To cover the inverse scenario of throwing an error if the color is not one of the choices on the instance? If that's the case, I think the test needs reworded to something like: `should be able to set a color defined by the mixin instance`.

libs/design/src/core/colorable/colorable.spec.ts
```diff
@@ -55,6 +55,29 @@ describe('daffColorMixin', () => {
         }).toThrow(new TypeError("SOMEFAKEPALETTE is not a valid color for the DaffPalette"));
 
     });
+
```
lderrickable: I don't think we should use the word "if" or "when" in an `it`. We should use `describe` to handle that. By doing this, the code becomes more test driven, and I think it becomes a whole lot easier for a new person to come in and digest both the tests and the code it is testing, because it's easier to read the spec file as a whole rather than test by test. Also, if we add more tests later that share an "if" or "when" with an existing test, we won't need to refactor it into a describe. But mostly because of the TDD reason.

libs/design/src/core/colorable/colorable.spec.ts
damienwebdev: @lderrickable I've resolved this in c4e7c2839bfbc887537b1b197107401c9f9fea7a

PR #223, opened 2019-01-28
Make all randomly generated prices under 1000

lderrickable:
make all randomly generated prices under 1000 to make formatted prices easier to test

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently randomly generated prices can range from 1-1000. A build failed because a formatted price in the DOM was `$1,000` instead of $1000`. 

Issue Number: N/A

## What is the new behavior?
Instead of making the tests formatted too, I just decreased the price range from 1-1000 to 1-999.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/core/testing/src/cart/factories/cart-item.factory.ts
+++ b/libs/core/testing/src/cart/factories/cart-item.factory.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import * as faker from 'faker/locale/en_US';
 
-import { CartItem, range } from '@daffodil/core';
+import { CartItem } from '@daffodil/core';
 
 import { ModelFactory } from "../../factories/factory";
 
@@ -18,13 +18,13 @@ export class MockCartItem implements CartItem {
   description = 'description';
   weight = faker.random.number(1000);
   qty = faker.random.number({min:1, max:100});
-  price = faker.random.number(1000);
+  price = faker.random.number(999);
   discount_percent = faker.random.number(10);
   discount_amount = faker.random.number(100);
   tax_percent = faker.random.number(10);
   tax_amount = faker.random.number(10);
-  row_total = faker.random.number(1000);
-  row_total_with_discount = faker.random.number(1000);
+  row_total = faker.random.number(999);
+  row_total_with_discount = faker.random.number(999);
   row_weight = faker.random.number(100);
   tax_before_discount = faker.random.number(100);
 }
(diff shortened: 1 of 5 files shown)


damienwebdev:
@lderrickable which particular test was failing? This seems more like a stop-gap than a final solution.

xelaint:
@damienwebdev @lderrickable I think there should be a solution for the test to pass for prices over 3 digits as well. There could definitely be a circumstance where a product is worth $1,000+ (i.e. Barney's or Saks)

lderrickable:
> @damienwebdev @lderrickable I think there should be a solution for the test to pass for prices over 3 digits as well. There could definitely be a circumstance where a product is worth $1,000+ (i.e. Barney's or Saks)

I agree that it could happen. I was just saying, do we need to test for a comma in unit tests? But I'll change it

damienwebdev:
@lderrickable can you replicate and give some scenarios with the test?

I get that the test is failing because of formatting, but I'd imagine there's a better way to test than what we're currently doing.

lderrickable:
I added some tests around formatting, and I made the random number generator cap at 1500 instead of 1000, so we'll be less likely to miss this in the future

damienwebdev:
@lderrickable is there a way to `spyOn` the pipe instead? I think we should be testing that we USE the pipe, not precisely what the pipe does.

lderrickable:
I think you're right, and there's probably a way, though I don't think it's worth the time and effort to figure out. Do you want me to try?

lderrickable:
Oh wait, I just thought of a way to maybe test it that wouldn't be super difficult. Stand by

lderrickable:
Ok, the way I was thinking of testing it didn't work out, but then spying on the angular CurrencyPipe ended up being not so bad. I changed the tests to spy on the currency pipe

apps/demo/src/app/cart/components/cart-totals/cart-totals.component.spec.ts
```diff
@@ -13,12 +13,17 @@ class WrapperComponent {
   @Input() cartValue: Cart;
 }
 
+@Pipe({name: 'currency'})
+class MockCurrencyPipe implements PipeTransform {
```
damienwebdev: @lderrickable as I'm sure we'll need to do this in several places, do you think it makes sense to migrate this into `@daffodil/core/testing`?

lderrickable: Yeah, I think that makes sense. I added an angular-mocks folder in core, and I added a `mock-pipes.ts` file.

lderrickable:
done

PR #224, opened 2019-01-29
feat(design): replace max-width with width

xelaint:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

max-width caused an issue when i updated the `hero-title` max-width to something smaller than the container size I set. The width on the containers should be a defined value.

Issue Number: N/A

## What is the new behavior?
width instead of max-width.

```
[ ] Yes
[x] No
```

## Other information


damienwebdev:
I don't think this change makes sense. A finite `width` will break the viewport on mobile devices, that's why `max-width` was used.

xelaint:
@damienwebdev I'm dumb. I reverted the max-width back and also added `width: 100%;` to fix the issue I was having.

PR #278, opened 2019-02-20
Move addtocart function

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

addToCart function is currently in ProductState which creates a dependency on CartState.

## What is the new behavior?
addToCart function is in CartState, so there is no dependency of ProductState on CartState.

```
[x] Yes
[ ] No
```

All breaks that occur in demo from this PR will be addressed in this PR.

### DIFF
--- a/apps/demo/src/app/product/pages/product-view/product-view.component.html
+++ b/apps/demo/src/app/product/pages/product-view/product-view.component.html
@@ -1,6 +1,8 @@
 <div product-container #ProductContainer="ProductContainer" [selectedProductId]="productId" class="product-container">
   <demo-product [product]="ProductContainer.product$ | async" [qty]="ProductContainer.qty$ | async" (updateQty)="ProductContainer.updateQty($event)" *ngIf="!(ProductContainer.loading$ | async)">
-    <demo-add-to-cart (addToCart)="ProductContainer.addToCart($event)" [additive]="ProductContainer.product$ | async"  [qty]="ProductContainer.qty$ | async"></demo-add-to-cart>
+    <div cart-container #CartContainer="CartContainer">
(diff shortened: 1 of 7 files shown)


apps/demo/src/app/product/pages/product-view/product-view.component.html
```diff
@@ -1,6 +1,8 @@
 <div product-container #ProductContainer="ProductContainer" [selectedProductId]="productId" class="product-container">
   <demo-product [product]="ProductContainer.product$ | async" [qty]="ProductContainer.qty$ | async" (updateQty)="ProductContainer.updateQty($event)" *ngIf="!(ProductContainer.loading$ | async)">
-    <demo-add-to-cart (addToCart)="ProductContainer.addToCart($event)" [additive]="ProductContainer.product$ | async"  [qty]="ProductContainer.qty$ | async"></demo-add-to-cart>
+    <div cart-container #CartContainer="CartContainer">
```
damienwebdev: Change to `ng-container`, we don't need the extra div here.

lderrickable: I don't think this is possible with angular. Explained here: https://github.com/angular/material2/issues/3670

Though the use-case isn't the same, and therefore the answer doesn't really answer why it can't be done in angular.

damienwebdev: Ah, understood. Just another reason to hate the template selectors we started with. I'm going to open an issue for us to spend some time re-thinking template selectors then, as they're currently extremely problematic.

Issue #152, opened 2018-12-12
feat: Rework modules into feature areas


damienwebdev:
## I'm submitting a...

<pre><code>
[ ] Regression (a behavior that used to work and stopped working in a new release)
[ ] Bug report  
[ ] Performance issue
[x] Feature request
[ ] Documentation issue or request
[ ] Other... Please describe:
</code></pre>

This issue is meant to track the migration of code into separate "feature" modules.

This is broken into several parts:

- [x] Separate core/product & core/testing/product into `@daffodil/product` and `@daffodil/product/testing`.
- [x] Separate core/cart & core/testing/cart into `@daffodil/cart` and `@daffodil/cart/testing`.
- [x] Separate core/checkout & core/testing/checkout into `@daffodil/checkout` and `@daffodil/checkout/testing`.
- [x] Move state/product into `@daffodil/product`
- [x] Move state/cart into `@daffodil/cart`
- [x] Move state/checkout into `@daffodil/checkout`
- [x] Remove state/src/product 
- [x] Remove state/src/cart
- [x] Remove state/src/checkout
- [ ] Remove product, cart, and checkout models and factories from core
- [x] Move product drivers to `@daffodil/product`
- [x] Move cart drivers to `@daffodil/cart`
- [x] Move checkout drivers to `@daffodil/checkout`

PR #282, opened 2019-02-21
Delete state/src/product

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: https://github.com/graycoreio/daffodil/issues/152

```
[x] Yes
[ ] No
```

All breaks in demo will be fixed in this branch.


apps/demo/src/app/modals/add-to-cart-notification/components/add-to-cart-notification/add-to-cart-notification.component.spec.ts
```diff
@@ -8,8 +8,8 @@ import { DaffProductFactory } from '@daffodil/core/testing';
 import { Product } from '@daffodil/core';
 import { DaffLoadingIconModule } from '@daffodil/design';
 
-// importing from @daffodil/state doesn't work.
-import * as fromProduct from 'libs/state/src/product/reducers/index';
+// importing from @daffodil/product doesn't work.
+import * as fromProduct from 'libs/product/src/reducers/index';
```
damienwebdev: Why doesn't this work, I would anticipate it should? Could you provide specifics into what's going on?

lderrickable: I looked into this again, and it turns out that importing it from @daffodil/product does work, but when I try to spy on the selector for getting a product later I get the error:

`selectProduct is not declared writable or has no setter`

I don't get that error if I import from src, but I still can't get the test to work because the mocked selector always comes back with a null Product for some reason.

damienwebdev: I've moved this into a separate issue #285

Issue #287, opened 2019-02-26
[BUG] Strange Page Redirects to Cart Occur


lderrickable:
# :bug: Bug report

## Current Behavior
After going through the checkout workflow, if you navigate to a product page, it will automatically redirect to the cart page. This happens because a new CartLoad action occurs, the EmptyCartResolver sees that the cart is empty, then it redirects to the cart page.

## Expected Behavior
A user should be able to navigate to a product page without being redirected.

## Minimal reproduction of the problem with instructions
- Add product to the cart
- Go through the checkout workflow and place the order
- Navigate to a product page

## What is the motivation / use case for changing the behavior?
It's an app-breaking bug.

lderrickable:
I think this would be fixed by lazy loading the checkout page. And I don't think we can do lazy loading until we make the modules into sharable instances of state as described here: https://blog.angular-university.io/angular2-ngmodule/

PR #293, opened 2019-02-27
fix(demo): fix bug where a subscription to cart caused a memory leak …

lderrickable:
…and other problems

## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: https://github.com/graycoreio/daffodil/issues/287

```
[ ] Yes
[x] No
```

## Other information
Unfortunate side-effect of this change is that, as we test/mock selectors now, the test suite for place-order.component throws errors and had to be skipped.

### DIFF
--- a/apps/demo/src/app/checkout/components/place-order/place-order.component.spec.ts
+++ b/apps/demo/src/app/checkout/components/place-order/place-order.component.spec.ts
@@ -4,7 +4,7 @@ import { StoreModule, combineReducers, Store } from '@ngrx/store';
 import { of, Subscription } from 'rxjs';
 
 import { PlaceOrder } from '@daffodil/checkout';
-import { Cart } from '@daffodil/cart';
+import { Cart, fromCart } from '@daffodil/cart';
 import { DaffCartFactory } from '@daffodil/core/testing';
 import { DaffDriverTestingModule } from '@daffodil/driver/testing';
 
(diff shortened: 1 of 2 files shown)


apps/demo/src/app/checkout/components/place-order/place-order.component.spec.ts
```diff
@@ -11,7 +11,8 @@ import { DaffDriverTestingModule } from '@daffodil/driver/testing';
 import * as fromDemoCheckout from '../../reducers';
 import { PlaceOrderComponent } from './place-order.component';
 
-describe('PlaceOrderComponent', () => {
+// Because the fromCart selector is now being subscribed to, any test throws an error. Skipping tests until fromCart can be mocked.
```
damienwebdev: Would we be able to import the `cart` state here and be able to test this? I'm aware that's increases the surface area more than necessary, but its better than just `xdescribing` the whole test suite. I've opened #294 to address this issue more thoroughly later.

lderrickable: ok don't laugh. Turns out the tests work if I add the cart store to the StoreModule for the spec even if I import from `@daffodil/cart`. I just can't spy on the selectors or the tests will break.

Issue #295, opened 2019-02-27
[BUG] Quantity Input shows letters after numbers are deleted.


xelaint:
# :bug: Bug report

## Current Behavior
![qty-bug](https://user-images.githubusercontent.com/15809918/53529016-633fd800-3ab9-11e9-9e21-1ee4ce4d2fcc.gif)

## Expected Behavior
No letters appearing if numbers in qty input are deleted.

## What is the motivation / use case for changing the behavior?
Correct UI

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

PR #320, opened 2019-03-16
fix(design): fix bug where qty-dropdown had no validation

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: https://github.com/graycoreio/daffodil/issues/295

```
[ ] Yes
[x] No
```


damienwebdev:
@lderrickable I don't think these changes make too much sense. Firstly, this component is super awkward and requires the use of dom-selectors. I think this component needs to be refactored to implement the [ControlValueAccessor](https://angular.io/api/forms/ControlValueAccessor). 

I also think that it makes more sense to prevent bad inputs (actually blocking the change of the value) instead of not emitting on bad changes.

lderrickable:
ok. I don't think we should do the refactor right now, because we agreed to table most design changes with `v0.0.2`. 

I'll look into the other changes though

lderrickable:
on second thought, I do actually block the change of the value as well as not emit any changes. But in demo, you're still going to run into the problem where if the user enters "-1" they can still add the product to the cart unless demo uses the "valid" flag.

damienwebdev:
Closing in favor of #330

PR #334, opened 2019-03-25
[WIP] docs(daffio): adds initial docs tooling for daffio

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently we don't have a methodology for generating docs.

## What is the new behavior?
This adds some initial tooling for docs generation.

```
[ ] Yes
[x] No
```

## Other information


damienwebdev:
@lderrickable I think we should from this branch and keep all work in this PR and squash when we're done. This'll single PR should probably act as the central trunk for the rest of v0.4. I don't think we should merge any docs related code until this is completed. For example, I think (when we're ready), we should merge #333 into this PR and then merge that as a single commit.

lderrickable:
Uhh...I think we should just do like one example here with the whole process. If we merge #333 here, it'll just needlessly make this PR really big. What if, instead, I just add the gulp and dgeni stuff to all the modules where it's needed in this PR. Then keep actually writing the documentation separate, and merge this branch first.

PR #348, opened 2019-06-17
[FEAT] Prefix cart

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

All exports of cart are using the unprefixed exports.

Fixes: N/A

## What is the new behavior?
This prefixes all the exports with `Daff`.

```
[x] Yes
[ ] No
```

All dependents will have to use the new prefix.

## Other information

### DIFF
--- a/libs/cart/testing/src/factories/cart-item.factory.ts
+++ b/libs/cart/testing/src/factories/cart-item.factory.ts
@@ -1,11 +1,11 @@
 import { Injectable } from '@angular/core';
 import * as faker from 'faker/locale/en_US';
 
-import { CartItem } from '@daffodil/cart';
+import { DaffCartItem } from '@daffodil/cart';
 
 import { DaffModelFactory } from "@daffodil/core/testing";
 
-export class MockCartItem implements CartItem {
+export class MockCartItem implements DaffCartItem {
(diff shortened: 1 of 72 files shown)


libs/cart/testing/src/factories/cart-item.factory.ts
```diff
@@ -1,11 +1,11 @@
 import { Injectable } from '@angular/core';
 import * as faker from 'faker/locale/en_US';
 
-import { CartItem } from '@daffodil/cart';
+import { DaffCartItem } from '@daffodil/cart';
 
 import { DaffModelFactory } from "@daffodil/core/testing";
 
-export class MockCartItem implements CartItem {
+export class MockCartItem implements DaffCartItem {
```
lderrickable: I missed this in the product PR, but I think this should be `MockDaffCartItem`

damienwebdev: So my thought here is that since `MockCartItem` isn't ever exported from the package, it doesn't need to have the prefix. Internal features don't warrant the prefix.

PR #349, opened 2019-06-28
Generate API docs in @daffodil packages using dgeni

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

We don't have api document generation right now.

Fixes: N/A

## What is the new behavior?
Documentation can now be generated from all @daffodil packages using dgeni. 

```
[ ] Yes
[x] No
```


tools/dgeni/karma.conf.js
```diff
@@ -0,0 +1,31 @@
+// Karma configuration file, see link for more information
+// https://karma-runner.github.io/1.0/config/configuration-file.html
+
+module.exports = function (config) {
```
damienwebdev: I don't think we run tests in the browser, so I don't think this file does anything?

tools/dgeni/src/test.ts
```diff
@@ -0,0 +1,22 @@
+// This file is required by karma.conf.js and loads recursively all the .spec and framework files
+
+import 'core-js/es7/reflect';
```
damienwebdev: I don't think we need this file for the tests either.

tools/dgeni/karma.conf.js
lderrickable: you're right. Removed

tools/dgeni/src/test.ts
lderrickable: Removed

PR #358, opened 2019-07-22
Fix some of the typing around the product-entity reducer

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The type for a product entity was still a DaffProduct, while it should be a DaffProductUnion.

Fixes: N/A

## What is the new behavior?
The type for a product entity is a DaffProductUnion and a factory was added for DaffProductUnions

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/product/src/reducers/index.ts
+++ b/libs/product/src/reducers/index.ts
@@ -5,7 +5,7 @@ import * as fromProductEntities from './product-entities.reducer';
 import * as fromProductGrid from './product-grid.reducer';
 import * as fromProduct from './product.reducer';
 import * as fromBestSellers from './best-sellers.reducer';
-import { DaffProduct } from '../models/product';
+import { DaffProductUnion } from '../models/product-union';
 
 export interface State {
   products : fromProductEntities.State;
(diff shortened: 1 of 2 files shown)


libs/product/src/reducers/index.spec.ts
```diff
@@ -11,7 +11,7 @@ import { DaffProduct } from "../models/product";
 describe('selectProductState', () => {
 
   let store: Store<fromProduct.State>;
-  const productFactory: DaffProductFactory = new DaffProductFactory();
+  const productUnionFactory: DaffProductUnionFactory = new DaffProductUnionFactory();
```
damienwebdev: Since the `union` is so broad, can we get away with only using the `DaffProductFactory` here?

lderrickable: So we can get rid of the union factory entirely?

damienwebdev: Yes.

lderrickable: removed ProductUnion factory

PR #360, opened 2019-07-24
docs(product): add docs for using the InMemory driver

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Using the InMemory Product driver is a bit difficult, because there's no documentation.

Fixes: N/A

## What is the new behavior?
Added documentation.

```
[ ] Yes
[x] No
```


libs/product/README.md
```diff
@@ -9,3 +9,58 @@ provides clear interfaces, models, and factories for the frontend of an ecommerc
...
+}
+
+export interface MockMyAppDatabase {
+  products: DaffProduct[];
+  myAppCollection: MyAppCollection[];
+}
+```
+With this, you'll have all get requests for products handled by `@daffodil/product` and all other requests handled by your `MyAppTestingService`.
+
+The only other thing you need to do is to make the product images in `@daffodil/product/assets` available to your application, because these are the image references used in the mocked product data.
```
damienwebdev: How do I do this?

libs/product/README.md
```diff
@@ -9,3 +9,58 @@ provides clear interfaces, models, and factories for the frontend of an ecommerc
 ```
 npm install @daffodil/product
 ```
+
+## Using the InMemory Driver
```
damienwebdev: I don't think we should ever assume a user will only use JUST the product in-mem. While it is possible, its also too confusing to explain the difference between one and the other. We should tend towards "singular" configuration policies vs. multiple.

libs/product/README.md
```diff
@@ -9,3 +9,58 @@ provides clear interfaces, models, and factories for the frontend of an ecommerc
...
+
+The InMemory driver is for rapid development without the need to set up a magento/shopify/etc backend. It will mock out the service calls to get products and return fake data. There are a few steps to wiring up the DaffInMemory driver:
+
+If you only plan on making serving calls for getting products, you can simply import `HttpClientInMemoryWebApiModule.forRoot(DaffInMemoryBackendProductService)` and `DaffProductInMemoryDriverModule.forRoot()` into your AppModule, and it will just work. If you need to make service calls that involve a cart or anything else, you'll need to create a middle layer between your app and the `DaffProductInMemoryDriverModule` so that the `DaffInMemoryBackendProductService` only ever receives product-related requests. This middle layer can just be an `in-memory-web-api` service that uses the `DaffInMemoryBackendProductService` to handle database creation for products and product get requests:
+
+```
```
damienwebdev: Add the language hint.

libs/product/README.md
```diff
@@ -9,3 +9,58 @@ provides clear interfaces, models, and factories for the frontend of an ecommerc
 ```
 npm install @daffodil/product
 ```
+
```
damienwebdev: I would move this into a separate "stories" folder.

libs/product/README.md
lderrickable: What's the language hint?

libs/product/README.md
lderrickable: added some extra stuff and an example with angular-cli

libs/product/README.md
lderrickable: changed to just have the more complex version.

lderrickable: done

libs/product/README.md
damienwebdev: You can add ```ts to colorize a codeblock in the TypeScript language format.

libs/product/README.md
damienwebdev: To clarify, I don't think this should be in the README of `product`. I think we should have a separate `stories` folder for docs which describes a "story" a developer would undergo.

libs/product/README.md
lderrickable: done

libs/product/README.md
lderrickable: done

PR #487, opened 2019-10-21
fix(category): make category in memory service return a category no m…

lderrickable:
…atter what id is given

The category in-memory service needs to return a category for an id given until we make the in memory DB more complete.

## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/category/testing/src/inmemory-backend/category.service.ts
+++ b/libs/category/testing/src/inmemory-backend/category.service.ts
@@ -22,7 +22,6 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
     private productInMemoryBackendService: DaffInMemoryBackendProductService
   ) {
     this.category = this.categoryFactory.create();
-    this.category.id = '1';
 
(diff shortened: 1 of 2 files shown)


libs/category/testing/src/inmemory-backend/category.service.ts
```diff
@@ -21,6 +22,7 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
     private productInMemoryBackendService: DaffInMemoryBackendProductService
   ) {
     this.category = this.categoryFactory.create();
+    this.category.id = '1';
```
damienwebdev: This should be the partial param in the factory `create`

lderrickable: turns out I don't need to set the category Id right now anyway, so I just removed that line

PR #490, opened 2019-10-21
fix(product): make DaffProduct.id actually be the productId instead o…

lderrickable:
…f sku

Right now the ID on DaffProduct is actually the product sku. This PR makes the DaffProduct.id come from the id property, and adds the sku property to the DaffProduct.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[x] Yes
[ ] No
```

## Other information
Probably nothing will break, because none of the apps really knew that the "DaffProduct.id" was actually the sku.


damienwebdev:
@lderrickable this doesn't work in Magento. See https://github.com/magento/graphql-ce/issues/86 Magento treats `sku` as a surrogate key. If you'd like to patch the category transformer to handle this edge case for Magento, I'll accept that as a PR.

PR #525, opened 2019-12-10
feat(contact): added effects to contact library

Nolan-Arnold:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/contact/src/effects/contact.effects.ts
+++ b/libs/contact/src/effects/contact.effects.ts
@@ -1,0 +1,42 @@
+import { Injectable, Inject } from '@angular/core';
+import { Actions, createEffect, ofType } from '@ngrx/effects';
+import { Action } from '@ngrx/store';
+import { Observable, of, EMPTY } from 'rxjs';
+import { switchMap, map, catchError } from 'rxjs/operators';
+
+import { DaffContactDriver } from '../driver/injection-tokens/contact-driver.token';
+import { DaffContactServiceInterface } from '../driver/interfaces/contact-service.interface';
+
+import { DaffContactActionTypes, DaffContactSubmit, DaffContactCancel, DaffContactSuccessSubmit, DaffContactFailedSubmit, DaffContactRetry } from '../actions/contact.actions';
+
+@Injectable()
+export class DaffContactEffects<T, V>{
+  constructor(
+    private actions$: Actions,
+    @Inject(DaffContactDriver) private driver: DaffContactServiceInterface<T, V>
+  ) { }
+
+  trySubmission$: Observable<Action> = createEffect(() => this.actions$.pipe(
+    ofType(DaffContactActionTypes.ContactSubmitAction,
+      DaffContactActionTypes.ContactRetryAction,
+      DaffContactActionTypes.ContactCancelAction),
+    switchMap((action: DaffContactSubmit<T> | DaffContactRetry<T> | DaffContactCancel) => {
+      if (action instanceof DaffContactCancel) {
+        return EMPTY;
+      }
+      else {
+        return this.submitContact(action.payload);
+      }
+    })
+  ));
+  private submitContact(contact: T): Observable<Action> {
+    return this.driver.send(contact).pipe(
+      map((resp: V) => {
+        return new DaffContactSuccessSubmit();
+      }),
+      catchError(error => {
+        return of(new DaffContactFailedSubmit(['Failed to submit']));
+      })
+    )
+  }
+}
--- a/libs/contact/src/effects/contact.effects.spec.ts
+++ b/libs/contact/src/effects/contact.effects.spec.ts
@@ -1,0 +1,104 @@
+import { Observable, of } from 'rxjs';
+import { DaffContactEffects } from './contact.effects';
+import { DaffContactUnion, DaffContactServiceInterface } from '@daffodil/contact';
+import { TestBed } from '@angular/core/testing';
+import { DaffTestingContactService } from 'libs/contact/testing/src/drivers/testing/contact.service';
+import { DaffContactDriver } from '../driver/injection-tokens/contact-driver.token';
+import { provideMockActions } from '@ngrx/effects/testing';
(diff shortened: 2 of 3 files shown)


libs/contact/src/effects/contact.effects.ts
```diff
@@ -0,0 +1,38 @@
...
+      }
+      else if (action instanceof DaffContactSubmit || action instanceof DaffContactRetry) {
+        return this.driver.send(action.payload).pipe(
+          map((resp: V) => {
+            return new DaffContactSuccessSubmit();
+          }),
+          catchError(error => {
+            return of(new DaffContactFailedSubmit(['Failed to submit']));
+          })
+        )
```
lderrickable: Could you move the this.driver.send call into a private function to simplify the effect a bit? Like
```
private submitContact(action: Action): Action {
   return this.driver.send(action.payload).pipe(
       map((resp: V) => {
         return new DaffContactSuccessSubmit();
       }),
       catchError(error => {
         return of(new DaffContactFailedSubmit(['Failed to submit']));
       })
    )
}
```

libs/contact/src/effects/contact.effects.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  trySubmission$: Observable<Action> = createEffect(() => this.actions$.pipe(
+    ofType(DaffContactActionTypes.ContactSubmitAction,
+      DaffContactActionTypes.ContactRetryAction,
+      DaffContactActionTypes.ContactCancelAction),
+    switchMap((action: DaffContactSubmit<T> | DaffContactRetry<T> | DaffContactCancel) => {
+      if ((action.type === DaffContactActionTypes.ContactCancelAction)) {
+        return of(action);
+      }
+      else if (action instanceof DaffContactSubmit || action instanceof DaffContactRetry) {
```
lderrickable: I feel like lines 22 and 25 should use the same logic, right? Kind of weird for them to be different

libs/contact/src/effects/contact.effects.ts
```diff
@@ -0,0 +1,38 @@
...
+          map((resp: V) => {
+            return new DaffContactSuccessSubmit();
+          }),
+          catchError(error => {
+            return of(new DaffContactFailedSubmit(['Failed to submit']));
+          })
+        )
+      }
+    }),
+    ofType(DaffContactActionTypes.ContactSuccessSubmitAction, DaffContactActionTypes.ContactFailedSubmitAction)
```
lderrickable: What is this for? Is it needed?

Nolan-Arnold: I'm now doubting myself on this, but I believe it filters the effect to only return either a `SuccessSubscribe` and a `FailedSubmit` observable. This really only matters so the `DaffContactCancel` is submitted it returns empty

libs/contact/src/effects/contact.effects.ts
Nolan-Arnold: This should be taken care of in c08df8248fc8c53e81472639126bd5feaecdf8fc

libs/contact/src/effects/contact.effects.ts
Nolan-Arnold: Okay, this was fixed to be consistent in c08df8248fc8c53e81472639126bd5feaecdf8fc

libs/contact/src/effects/contact.effects.ts
Nolan-Arnold: Okay this should be addressed in 74bef0998c0734b59d30109584e192d22b6e14c5.
Let me know if those changes look good. After looking at it I do think the effect now looks more simple.

libs/contact/src/effects/contact.effects.ts
```diff
@@ -0,0 +1,42 @@
...
+    switchMap((action: DaffContactSubmit<T> | DaffContactRetry<T> | DaffContactCancel) => {
+      if ((action instanceof DaffContactCancel)) {
+        return EMPTY;
+      }
+      else {
+        return this.submitContact(action, action.payload);
+      }
+    })
+  ));
+  private submitContact(action: Action, payload: any): Observable<Action> {
```
lderrickable: doesn't look like the first `action` parameter is needed

lderrickable: also, payload should be typed as `T`

libs/contact/src/effects/contact.effects.ts
```diff
@@ -0,0 +1,42 @@
...
+    private actions$: Actions,
+    @Inject(DaffContactDriver) private driver: DaffContactServiceInterface<T, V>
+  ) { }
+
+  trySubmission$: Observable<Action> = createEffect(() => this.actions$.pipe(
+    ofType(DaffContactActionTypes.ContactSubmitAction,
+      DaffContactActionTypes.ContactRetryAction,
+      DaffContactActionTypes.ContactCancelAction),
+    switchMap((action: DaffContactSubmit<T> | DaffContactRetry<T> | DaffContactCancel) => {
+      if ((action instanceof DaffContactCancel)) {
```
lderrickable: don't need the extra `()`

libs/contact/src/effects/contact.effects.ts
lderrickable: probably call the parameter contact instead of payload, since `submitContact` doesn't need to know it's getting an action.payload.

Nolan-Arnold: Okay this should be taken care of in e775d4f9f31c6f40edbfe5252e1014a0e244e0c8

lderrickable:
lgtm

lderrickable:
looks like this just needs a squash, then I'll approve

Nolan-Arnold:
Squashed in 7ebcfb11bcb71f17cdc095e9e0879f1b4355fdd8 @lderrickable

PR #549, opened 2020-01-10
feat(checkout): add driver tokens, interfaces, and models for orders …

lderrickable:
…and payments

This pull request is to make a separation between how payment processors and how order processors are provided. 

## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```


libs/checkout/src/drivers/interfaces/payment-service.interface.ts
```diff
@@ -0,0 +1,20 @@
+import { Observable } from 'rxjs';
+import { DaffCharge } from '../../models/payment/request/charge';
+import { DaffTransactionResponse } from '../../models/payment/response/transaction-response';
+
+export interface DaffPaymentServiceInterface {
```
damienwebdev: I think the interface here is wrong, our only consideration here is generating the payment nonce. So, a payment interface I would expect to look something like:

```ts
export interface DaffPaymentServiceInterface {
   generateToken(card: DaffCreditCard): Observable<DaffCreditCardToken>;
}
```

Authorizing and charging cards is the responsibility of a backend system.

libs/checkout/src/models/payment/request/charge.ts
```diff
@@ -0,0 +1,11 @@
+import { DaffCreditCardInfo } from './credit-card-info';
+import { DaffMerchantAuthentication } from './merchant-authentication';
+
+/**
+ * An object for processing a payment.
+ */
+export interface DaffCharge {
```
damienwebdev: This looks too tied to Authorize.net

libs/checkout/src/models/payment/request/merchant-authentication.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Authentication information needed to process a payment.
+ */
+export interface DaffMerchantAuthentication {
```
damienwebdev: This looks too tied to authorize.net.

libs/checkout/src/models/payment/response/transaction-response.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * A base class for transaction responses.
+ */
+export interface DaffTransactionResponse {
```
damienwebdev: This looks too tied to authorize.net.

libs/checkout/src/drivers/interfaces/order-service.interface.ts
```diff
@@ -0,0 +1,10 @@
+import { Observable } from 'rxjs';
+
+import { Order } from '../../models/order/order';
+
+export interface DaffOrderServiceInterface {
```
damienwebdev: If we were to eventually have an "orders" module which handles the process of order management, would this cause a namespace conflict?

libs/checkout/src/drivers/interfaces/payment-service.interface.ts
lderrickable: I redid all of this. See what you think

damienwebdev:
@lderrickable this is looking pretty good. I think this will work!

libs/checkout/src/drivers/interfaces/order-service.interface.ts
lderrickable: no, because they would be imported from different places, `@daffodil/checkout` and `@daffodil/order`

Issue #537, opened 2020-01-05
[FEAT] Add support for Magento 2 GraphQL Guest Cart Driver


damienwebdev:
# :bulb: Feature request

## Feature Name
Magento 2 Rest Cart Driver

## The Desired Behavior
We'd add a Magento 2 driver that implements [the cart interface](https://github.com/graycoreio/daffodil/blob/develop/libs/cart/src/drivers/interfaces/cart-service.interface.ts) and makes API requests to add items to/from a given cart.

## Your Use Case
As a user, I'd like to be able to add items to my cart if my ecommerce platform of choice is Magento 2.

## Prior Work
Here's the Magento 2 tutorial for the [REST API for cart](https://devdocs.magento.com/guides/v2.3/rest/tutorials/orders/order-config-store.html).

Here's the [API Docs](https://devdocs.magento.com/redoc/2.3/guest-rest-api.html#tag/guest-carts)

damienwebdev:
@griest024

griest024:
@damienwebdev Why shouldn't I use the GraphQL API for this driver? Does it not support guest carts?

damienwebdev:
Last I checked with the team at Magento, the GraphQL implementation isn't quite ready yet.

griest024:
Okay. I only ask because the Magento product driver uses GraphQL.

griest024:
Closing because this was addressed.

PR #551, opened 2020-01-11
[WIP] feat(cart): add magento driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No Magento support for the cart package.

Fixes: #537 

## What is the new behavior?
Magento support for the cart package.

```
[x] Yes
[ ] No
```

- Adds necessary driver interfaces for transformers
- Adds new cart service driver

## Other information

### DIFF
--- a/libs/cart/src/drivers/injection-tokens/cart-address-transformer.token.ts
+++ b/libs/cart/src/drivers/injection-tokens/cart-address-transformer.token.ts
@@ -1,0 +1,5 @@
+import { InjectionToken } from '@angular/core';
+import { DaffCartAddressTransformerInterface } from '../interfaces/cart-address-transformer.interface';
+import { DaffCartAddress } from '../../models/cart-address';
+
+export const DaffCartAddressTransformer = new InjectionToken<DaffCartAddressTransformerInterface<DaffCartAddress>>('DaffCartAddressTransformer');
--- a/libs/cart/src/drivers/injection-tokens/cart-item-transformer.token.ts
+++ b/libs/cart/src/drivers/injection-tokens/cart-item-transformer.token.ts
@@ -1,0 +1,5 @@
+import { InjectionToken } from '@angular/core';
+import { DaffCartItemTransformerInterface } from '../interfaces/cart-item-transformer.interface';
+import { DaffCartItem } from '../../models/cart-item';
+
+export const DaffCartItemTransformer = new InjectionToken<DaffCartItemTransformerInterface<DaffCartItem>>('DaffCartItemTransformer');
--- a/libs/cart/src/drivers/injection-tokens/cart-payment-transformer.token.ts
+++ b/libs/cart/src/drivers/injection-tokens/cart-payment-transformer.token.ts
@@ -1,0 +1,5 @@
+import { InjectionToken } from '@angular/core';
+import { DaffCartPaymentTransformerInterface } from '../interfaces/cart-payment-transformer.interface';
+import { DaffCartPayment } from '../../models/cart-payment';
+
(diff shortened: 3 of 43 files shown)


griest024:
Just wanted to see if I'm on the right track here and discuss some potential problems. The transformer tests are just stubs at the moment.

In particular, I wanted to talk about the transformers. It seems to me that we would always want the transformations to be reversible. Its tedious to maintain multiple transformers that implement the same transformation in both directions.

My idea is to provide a simple API of operations to transformers for defining tranformations in such a way that its reversible. Most of the time its a simple copy operation from one key to another, which could be easily reversed.

damienwebdev:
@griest024 this is in the right direction. I have to write a spot of code for Magento to address some CORS errors so that I can do an acceptance test of whether this actually works, but the idea here looks pretty good.

damienwebdev:
In terms of the invertible transformers, the idea seems solid in practice, but I can't quite see a use-case for it yet.

griest024:
> In terms of the invertible transformers, the idea seems solid in theory, but I can't quite see a use-case for it yet.

Well it looks like you don't have many interface methods that actually mutate DB data yet but I assume that you would at some point to provide full featured interaction with the platforms. In this case, you would almost always need to convert daffodil models into platform-specific models.

damienwebdev:
@griest024 Ah, I see what you're getting at. The issue is actually quite broad, and I hadn't even considered it yet. 

We'll need (for essentially every driver call), a transformation in both directions. It's actually almost universal for every driver call. Give me some time to iron out a proposal for exactly how this could theoretically work.

lderrickable:
Just looking at this now. I don't understand why we would ever directly change data that were magento model types. We only ever need to transform into magento requests and out of magento responses. Maybe I'm missing something?

lderrickable:
Also, I was looking into generic tokens, and as far as I can tell, they don't actually enforce strong typing. I'm not super familiar with how angular injects services through injection tokens, but I don't think it actually has a way to check the typing. I tried this and the build still passed:
```
@NgModule({
  imports: [
    CommonModule
  ]
})
export class DaffNavigationMagentoDriverModule {
  static forRoot(): ModuleWithProviders<DaffNavigationMagentoDriverModule> {
    return {
      ngModule: DaffNavigationMagentoDriverModule,
      providers: [
        {
          provide: DaffNavigationDriver,
          useExisting: DaffMagentoNavigationTransformerService      <-------------------
        },
        {
          provide: DaffNavigationTransformer,
          useExisting: DaffMagentoNavigationTransformerService
        }
      ]
    };
  }
}
```

griest024:
@lderrickable Using the example of setting shipping information, the [request payload](https://devdocs.magento.com/redoc/2.3/guest-rest-api.html#operation/checkoutGuestShippingInformationManagementV1SaveAddressInformationPost) needs the address in the form of a Magento `AddressInterface`. Assuming Daffodil (or code that leverages Daffodil) handles the collecting of the customer's shipping address, it will necessarily require the transformation of a `DaffAddress` or `DaffCartAddress` into a Magento `AddressInterface`.

I assume this is what you mean when you say "transform into magento requests". This is one of the use cases I was considering when I suggested the invertible transformers. It seems to me that most, if not all, mutable requests (that is, POST and PUT) will require transforming some Daffodil models into platform-specific models.

lderrickable:
@griest024 I think I see what you mean. In the shipping address example you gave, you mean that when updating the cart with an address we would need to send something like `updateShippingAddress(cartId, address)`, which would return a cart that has an address on it. So if we made just an address transformer, we could use it both ways to transform back and forth between daffodil and magento addresses. Is that right?

griest024:
@lderrickable that's correct.

PR #566, opened 2020-01-16
feat(authorizenet): add Authorize.net (Accept.js) package

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
The authorize.net package is written to be completely separate from `@daffodil/checkout`. It has its own driver tokens and service interfaces. The thinking behind this is because the app dev will have authorizenet specific code in their codebase, so there's not a whole lot of reason to have the `@daffodil/authorizenet` work generically with `@daffodil/checkout`. Making it completely separate also greatly decreases the complexity of the daffodil packages. 

Usage is similar to any other package. Trigger an action, service call and state is handled `@daffodil/authorizenet`, and a success or failure action is triggered. The app dev is responsible for sending the token over to the `@daffodil/cart`/wherever.


lderrickable:
I tested this is the demo app, and it does work, i.e. it calls the authorize.net endpoint and throws an error.

libs/authorizenet/src/drivers/authorize-net.service.ts
```diff
@@ -0,0 +1,43 @@
...
+
+	constructor(
+		@Inject(DaffPaymentTransformer) public paymentTransformer: DaffPaymentTransformerService<DaffAuthorizeNetTokenRequest>,
+		private store: Store<any>
+	) {}
+
+	generateToken(paymentTokenRequest: DaffAuthorizeNetTokenRequest): string {
+		Accept.dispatchData(this.paymentTransformer.transformOut(paymentTokenRequest), this.responseHandler.bind(this));
+
+		return 'this string needs to be returned because authorize.net does not have a normal api';
```
Nolan-Arnold: Can this just be a void function instead of returning a string?

lderrickable: I don't think so, because the `DaffPaymentService.generateToken()` _should_ ideally return a string. And I think that for most payment processor APIs, it should be able to just fine. It's just that the authorize.net API is weird and makes that impossible. So to return void, I'd need to change the `DaffPaymentService` to something like
```
generateToken(paymentTokenRequest: DaffPaymentTokenRequest): string | void;
```
And I'd rather contain the weirdness to authorize.net instead

libs/authorizenet/src/drivers/authorize-net.module.ts
```diff
@@ -0,0 +1,30 @@
...
+import { DaffAuthorizeNetPaymentService } from './authorize-net.service';
+import { DaffAuthorizeNetTransformerService } from './transformers/authorize-net-transformer.service';
+
+@NgModule({
+  imports: [
+    CommonModule
+  ]
+})
+export class DaffAuthorizeNetDriverModule {
+  static forRoot(): ModuleWithProviders {
```
damienwebdev: What happens when there are two providers for the `DaffPaymentDriver` should this be a `multi` provider?

libs/authorizenet/src/authorize-net.module.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Ngrx/store
+     */
+    DaffAuthorizeNetStateModule
+  ]
+})
+export class DaffAuthorizeNetModule { 
+	constructor() {
+		const acceptJsScript = document.createElement('script');
```
damienwebdev: This probably shouldn't be in the constructor of the module. We should have a system to determine when the`accept.js` lib is loaded successfully.

libs/authorizenet/src/authorize-net.module.ts
```diff
@@ -0,0 +1,27 @@
+import { NgModule } from '@angular/core';
+import { CommonModule } from '@angular/common';
+
+import { DaffAuthorizeNetStateModule } from './authorize-net-state.module';
+
+const ACCEPT_LIBRARY = 'https://jstest.authorize.net/v1/Accept.js';
```
damienwebdev: This uses the `sandbox` version of the library, we need a way to configure the production version.

libs/authorizenet/src/drivers/authorize-net.service.ts
```diff
@@ -0,0 +1,43 @@
...
+})
+export class DaffAuthorizeNetPaymentService implements DaffPaymentServiceInterface<DaffAuthorizeNetTokenRequest> {
+
+	constructor(
+		@Inject(DaffPaymentTransformer) public paymentTransformer: DaffPaymentTransformerService<DaffAuthorizeNetTokenRequest>,
+		private store: Store<any>
+	) {}
+
+	generateToken(paymentTokenRequest: DaffAuthorizeNetTokenRequest): string {
+		Accept.dispatchData(this.paymentTransformer.transformOut(paymentTokenRequest), this.responseHandler.bind(this));
```
damienwebdev: This should be an observable, there's a way to port this `node.js` style callback to an observable:

https://stackoverflow.com/questions/43131162/making-an-observable-from-a-callback

libs/authorizenet/src/drivers/authorize-net.service.ts
```diff
@@ -0,0 +1,43 @@
...
+
+		return 'this string needs to be returned because authorize.net does not have a normal api';
+	}
+
+	private responseHandler(response: AuthorizeNetResponse) {
```
damienwebdev: This should be in an effect, not here, I understand why its is here, but hopefully the above refactor will allow it to be moved.

libs/authorizenet/src/drivers/transformers/authorize-net-transformer.service.ts
```diff
@@ -0,0 +1,31 @@
...
+	transformOut(request: DaffAuthorizeNetTokenRequest): AuthorizeNetRequest {
+		return {
+			cardData: {
+				cardNumber: request.creditCard.cardnumber,
+				cardCode: request.creditCard.securitycode,
+				month: request.creditCard.month,
+				year: request.creditCard.year
+			},
+			authData: {
+				clientKey: request.authData.clientKey,
```
damienwebdev: This should probably come from a configuration object and be a part of the driver call, not the transformer.

libs/authorizenet/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "directory": "../../dist/authorizenet"
+  },
+  "scripts": {
+    "build": "ng build authorizenet --prod",
+    "lint": "ng lint authorizenet",
+    "test": "ng test authorizenet --watch=false --browsers=ChromeHeadless",
+    "publish": "(cd ../../dist/authorizenet; npm publish --access=public)"
+  },
+  "homepage": "https://github.com/graycoreio/daffodil",
+  "peerDependencies": {
```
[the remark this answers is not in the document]
lderrickable: yes

libs/authorizenet/package.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "optionalDependencies": {
+    "faker": "0.0.0-PLACEHOLDER",
+    "angular-in-memory-web-api": "0.0.0-PLACEHOLDER",
+    "apollo-angular": "0.0.0-PLACEHOLDER",
+    "apollo-client": "0.0.0-PLACEHOLDER",
+    "graphql-tag": "0.0.0-PLACEHOLDER"
+  },
+  "devDependencies": {
+    "@daffodil/checkout": "0.0.0-PLACEHOLDER"
```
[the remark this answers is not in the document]
lderrickable: authorize.net is dependent on checkout. It needs to import the `DaffPaymentService` and `DaffPaymentTransformer` tokens.

libs/checkout/src/models/payment/payment-token-request.ts
```diff
@@ -0,0 +1,5 @@
+import { DaffCreditCard } from './credit-card';
+
+export interface DaffPaymentTokenRequest {
```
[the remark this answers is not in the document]
lderrickable: I don't understand how you'd implement a disjoint union as an interface. Could you give me a simple example?

libs/authorizenet/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@daffodil/authorizenet",
+  "version": "0.0.0-PLACEHOLDER",
+  "description": "An ecommerce categories package that has multi-platform support and state management features.",
```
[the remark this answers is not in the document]
lderrickable: done

libs/authorizenet/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "peerDependencies": {
+    "@angular/common": "0.0.0-PLACEHOLDER",
+    "@angular/core": "0.0.0-PLACEHOLDER",
+    "@daffodil/checkout": "0.0.0-PLACEHOLDER",
+    "@ngrx/effects": "0.0.0-PLACEHOLDER",
+    "@ngrx/entity": "0.0.0-PLACEHOLDER",
+    "@ngrx/store": "0.0.0-PLACEHOLDER",
+    "rxjs": "0.0.0-PLACEHOLDER"
+  },
+  "optionalDependencies": {
```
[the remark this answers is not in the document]
lderrickable: I removed them

libs/authorizenet/src/drivers/authorize-net.service.ts
```diff
@@ -0,0 +1,43 @@
...
+		DaffPaymentServiceInterface,
+		DaffPaymentTransformer,
+		DaffPaymentTransformerService
+	} from '@daffodil/checkout';
+
+import { DaffAuthorizeNetGenerateTokenSuccess, DaffAuthorizeNetGenerateTokenFailure } from '../actions/authorizenet.actions';
+import { AuthorizeNetResponse } from '../models/response/authorize-net-response';
+import { DaffAuthorizeNetTokenRequest } from '../models/request/authorize-net-token-request';
+
+declare var Accept: any;
```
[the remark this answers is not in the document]
lderrickable: I made an interface for this

libs/authorizenet/src/drivers/authorize-net.service.ts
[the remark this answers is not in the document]
lderrickable: removed

libs/authorizenet/src/drivers/authorize-net.module.ts
lderrickable: yea, I changed it

libs/authorizenet/src/authorize-net.module.ts
lderrickable: moved to an action and effect

libs/authorizenet/src/authorize-net.module.ts
lderrickable: I'm going to add a config object, but all the work around that is in the paypal PR, so can we push this PR without this change, then push the paypal, then I'll come back to this?

libs/authorizenet/src/drivers/authorize-net.service.ts
lderrickable: done

libs/authorizenet/src/drivers/authorize-net.service.ts
lderrickable: done

libs/authorizenet/src/drivers/transformers/authorize-net-transformer.service.ts
lderrickable: I'm going to add a config object, but all the work around that is in the paypal PR, so can we push this PR without this change, then push the paypal, then I'll come back to this?

lderrickable: I decided to separate this module completely from checkout payment. So I added the configuration token

libs/checkout/src/models/payment/payment-token-request.ts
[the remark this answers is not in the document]
lderrickable: I decided to separate this module completely from checkout payment, so I deleted this from checkout

libs/authorizenet/package.json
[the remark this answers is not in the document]
lderrickable: I removed this dependency

PR #572, opened 2020-01-29
feat(cart, demo, checkout): make cart semi-generic

damienwebdev:
BREAKING CHANGE: this makes the interface generic, so client dependencies will need to implement the generic.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The cart doesn't handle generics very well.

Fixes: N/A

## What is the new behavior?
This adds semi-generic cart driver types.

```
[x] Yes
[ ] No
```

Since this makes the interface generic, dependent codebases depending on the interface (as they should) will need to use the generic on the interface.

## Other information

### DIFF
--- a/libs/cart/src/drivers/injection-tokens/cart-driver.token.ts
+++ b/libs/cart/src/drivers/injection-tokens/cart-driver.token.ts
@@ -1,3 +1,4 @@
 import { InjectionToken } from '@angular/core';
+import { DaffCartServiceInterface } from '../interfaces/cart-service.interface';
 
-export const DaffCartDriver = new InjectionToken('DaffCartDriver');
+export const DaffCartDriver = new InjectionToken<DaffCartServiceInterface<any>>('DaffCartDriver');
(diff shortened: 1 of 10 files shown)


libs/cart/src/drivers/injection-tokens/cart-driver.token.ts
```diff
@@ -1,3 +1,4 @@
 import { InjectionToken } from '@angular/core';
+import { DaffCartServiceInterface } from '../interfaces/cart-service.interface';
 
-export const DaffCartDriver = new InjectionToken('DaffCartDriver');
+export const DaffCartDriver = new InjectionToken<DaffCartServiceInterface<any>>('DaffCartDriver');
```
lderrickable: I don't think this actually enforces anything. I've tested it by injecting something else, and the build still doesn't seem to fail. Do you think we should keep it or remove it?

damienwebdev: I vote keep as it will eventually do something in a later version of Angular.

lderrickable: ok sure

lderrickable:
lgtm

PR #573, opened 2020-01-29
feat(cart): make cart clear API return the new cart

damienwebdev:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Depends on #572

Currently, when you clear, you're left without a new cart object to work with.

## What is the new behavior?
This returns a new cart object when you clear the cart.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/cart/src/effects/cart.effects.ts
+++ b/libs/cart/src/effects/cart.effects.ts
@@ -63,8 +63,8 @@ export class DaffCartEffects<T extends DaffCart> {
     ofType(DaffCartActionTypes.CartResetAction),
     switchMap(() =>
       this.driver.clear(this.storage.getCartId()).pipe(
-        map(() => {
-          return new DaffCartResetSuccess();
+        map((resp) => {
+          return new DaffCartResetSuccess(resp);
         }
         ),
         catchError(error => {
--- a/libs/cart/src/effects/cart.effects.ts
+++ b/libs/cart/src/effects/cart.effects.ts
@@ -10,8 +10,11 @@ import {
   DaffCartLoadFailure, 
   DaffAddToCartSuccess,
   DaffAddToCartFailure,
-  DaffAddToCart} from '../actions/cart.actions';
-import { DaffCartServiceInterface, DaffCartDriver } from '../drivers/public_api';
+  DaffAddToCart,
+  DaffCartResetSuccess,
+  DaffCartResetFailure} from '../actions/cart.actions';
+import { DaffCartDriver } from '../drivers/injection-tokens/cart-driver.token';
+import { DaffCartServiceInterface } from '../drivers/interfaces/cart-service.interface';
 import { DaffCart } from '../models/cart';
 import { DaffCartStorageService } from '../storage/cart-storage.service';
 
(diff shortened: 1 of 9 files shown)


libs/cart/src/effects/cart.effects.ts
```diff
@@ -52,4 +54,17 @@ export class DaffCartEffects {
         )
     )
   )
+
+  @Effect()
+  clearCart$ = this.actions$.pipe(
+    ofType(DaffCartActionTypes.CartResetAction),
+    switchMap(() =>
+      this.driver.clear().pipe(
+          map((resp) => new DaffCartResetSuccess(resp)),
```
Nolan-Arnold: I believe this should be `return new DaffCartResetSuccess(resp)),`

libs/cart/src/effects/cart.effects.ts
```diff
@@ -10,7 +10,9 @@ import {
   DaffCartLoadFailure, 
   DaffAddToCartSuccess,
   DaffAddToCartFailure,
-  DaffAddToCart} from '../actions/cart.actions';
+  DaffAddToCart,
+  DaffCartResetSuccess,
+  DaffCartResetFailure} from '../actions/cart.actions';
 import { DaffCartDriver } from '../drivers/injection-tokens/cart-driver.token';
 import { DaffCartServiceInterface } from '../drivers/interfaces/cart-service.interface';
```
Nolan-Arnold: `DaffCartEffects` should now be `DaffCartEffects<T extends DaffCart>`

libs/cart/src/effects/cart.effects.ts
damienwebdev: @Nolan-Arnold this is a short-hand equivalent of the same.

```
() => { return "string"; }
() => "string";
function(){
  return "string";
}
```

are all equivalent in `es6`

lderrickable:
Why do we want a new cart instead of just an old, empty cart? To remove the selected shipping address and payment method? I would think we'd want to keep those.

damienwebdev:
> Why do we want a new cart instead of just an old, empty cart? To remove the selected shipping address and payment method? I would think we'd want to keep those.

This implementation will be dependent on each platform driver. This merely updates the interface so that end users don't have to make additional driver calls after clearing, they'll get back SOME new cart, depending on what the platform does behind the scenes.

lderrickable:
ok sure. lgtm

libs/cart/src/effects/cart.effects.ts
```diff
@@ -53,4 +55,17 @@ export class DaffCartEffects<T extends DaffCart> {
         )
     )
   )
+
+  @Effect()
+  clearCart$ = this.actions$.pipe(
```
lderrickable: This needs tested

damienwebdev:
@lderrickable needs re-review

Nolan-Arnold:
I'm not 100% sure why codeclimate doesn't like this code.

lderrickable:
lgtm

PR #591, opened 2020-02-03
feat(auth): add testing in-memory drivers

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information


libs/auth/testing/src/drivers/in-memory/login/login.service.ts
```diff
@@ -0,0 +1,22 @@
...
+  providedIn: 'root'
+})
+export class DaffInMemoryLoginService implements DaffLoginServiceInterface<DaffLoginRequest, DaffLoginResponse> {
+  url = '/api/auth/';
+
+  constructor(private http: HttpClient) {}
+
+  login({email, password}: DaffLoginRequest): Observable<DaffLoginResponse> {
+    return this.http.post<DaffLoginResponse>(`${this.url}login`, {email, password});
+  }
```
lderrickable: can't this just be: 
```
login(request: DaffLoginRequest): Observable<DaffLoginResponse> {
    return this.http.post<DaffLoginResponse>(`${this.url}login`, request);
}
```
?

griest024: ah yeah it can. I think I was doing something else with `email` and `password` initially that I later removed.

libs/auth/testing/src/inmemory-backend/auth.service.ts
```diff
@@ -0,0 +1,82 @@
...
+    return faker.random.alphaNumeric(16);
+  }
+
+  private setUserInfo(email, password) {
+    this.users[email] = {password};
+  }
+
+  createDb(): any {
+    return {
+      auth: this.users
```
damienwebdev: We don't need to store users. For now, just assume the user exists for logins, and registering returns successfully.

libs/auth/testing/src/drivers/in-memory/register/register.service.ts
```diff
@@ -0,0 +1,42 @@
...
+> {
+  url = '/api/auth/';
+
+  constructor(
+    private http: HttpClient,
+    @Inject(DaffLoginDriver) private loginDriver: DaffLoginServiceInterface<DaffLoginRequest, DaffLoginResponse>
+  ) {}
+
+  register(registration: DaffRegisterRequest<DaffCustomerRegistration>): Observable<DaffRegisterResponse> {
+    return this.http.post<any>(`${this.url}register`, registration).pipe(
```
damienwebdev: This violates the type constraint of the interface.

libs/auth/testing/src/index.ts
```diff
@@ -2,3 +2,8 @@ export { DaffAccountRegistrationFactory } from './factories/account-registration
 export { DaffCustomerRegistrationFactory } from './factories/customer-registration.factory';
 export { DaffAuthTokenFactory } from './factories/auth-token.factory';
 
+export { DaffInMemoryBackendAuthService } from './inmemory-backend/auth.service';
+
+export { DaffAuthInMemoryDriverModule } from './drivers/in-memory/auth-driver.module';
+export { DaffInMemoryLoginService } from './drivers/in-memory/login/login.service';
```
damienwebdev: These latter services should not be exposed at the moment.

libs/auth/testing/src/drivers/in-memory/register/register.service.ts
griest024: The in-memory backend interface?

griest024: Is this what I should do?
```ts
this.http.post<DaffCustomerRegistration>(`${this.url}register`, registration).pipe(
```

damienwebdev: Yes. That's better, but I believe you're returning a `DaffLoginResponse` at the moment.

libs/auth/testing/src/inmemory-backend/auth.service.integration.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+      ],
+    });
+
+    httpClient = TestBed.get(HttpClient);
+  });
+
+  describe('processing a login request', () => {
+    it('should process post requests of the form `/api/auth/login` and return a token', done => {
+      httpClient.post('/api/auth/login', {}).subscribe(result => {
+        expect(result).toBeDefined();
```
damienwebdev: Is being `defined` sufficient evidence that this test passes?

libs/auth/testing/src/inmemory-backend/auth.service.unit.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+      expect(result.body).toBeDefined();
+      expect(result.status).toEqual(STATUS.OK);
+    });
+  });
+
+  describe('processing a register request', () => {
+    it('should process post requests of the form `/api/auth/register` and return the customer info with a CREATED status', () => {
+      reqInfoStub.id = 'register';
+      reqInfoStub.utils.getJsonBody = () => registrationFactory.create();
```
damienwebdev: Let's just return an empty success response on creating an account, or potentially, just return something that looks like a `guid` since that's fairly common amongst providers.

damienwebdev: Can we set the `body` manually, or do we have to rely on this odd internal mechanism?

libs/auth/testing/src/drivers/in-memory/register/register.service.ts
griest024: They are both aliases of `DaffAuthToken`, so its functionally the same. Would you prefer typing it as `DaffLoginResponse` or `DaffRegisterResponse`?

libs/auth/testing/src/inmemory-backend/auth.service.integration.spec.ts
griest024: I can check that its token is a string. Without mocking internals, I can't know the value of the token.

libs/auth/testing/src/inmemory-backend/auth.service.unit.spec.ts
griest024: I can set `reqInfoStub.req.body` but that would require that I set `reqInfoStub.utils.getJsonBody` to mimic to the behavior of the real `RequestInfo`. Unless there is a way to generate mocked `RequestInfo`s that I'm not aware of, I have to deal with `getJsonBody` at some point.

libs/auth/testing/src/inmemory-backend/auth.service.integration.spec.ts
griest024: I checked for a defined token field on the result. Still a stronger check than I had.

libs/auth/testing/src/drivers/in-memory/register/register.service.ts
damienwebdev: I'd expect a `DaffRegisterResponse` for keeping the type the visibly consistent.

griest024:
don't merge yet, needs #614 first.

griest024:
this is g2g now

PR #592, opened 2020-02-03
feat(auth): add testing drivers

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/auth/testing/src/drivers/testing/login/login.service.ts
+++ b/libs/auth/testing/src/drivers/testing/login/login.service.ts
@@ -1,0 +1,21 @@
+import { Injectable } from '@angular/core';
+import { Observable, of } from 'rxjs';
+
+import {
+  DaffLoginServiceInterface,
+  DaffLoginInfo,
+  DaffAuthToken
+} from '@daffodil/auth';
+
+import { DaffAuthTokenFactory } from '../../../factories/auth-token.factory';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffTestingLoginService implements DaffLoginServiceInterface<DaffLoginInfo, DaffAuthToken> {
+  factory = new DaffAuthTokenFactory();
+
+  login(loginInfo: DaffLoginInfo): Observable<DaffAuthToken> {
+    return of(this.factory.create());
+  }
+}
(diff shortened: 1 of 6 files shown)


libs/auth/testing/src/drivers/testing/register/register.service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffTestingRegisterService implements DaffRegisterServiceInterface<
+  DaffAccountRegistration<DaffCustomerRegistration>,
+  DaffCustomerRegistration,
+  DaffAuthToken
+> {
+  factory = new DaffAuthTokenFactory();
```
damienwebdev: Inject the factory.

libs/auth/testing/src/drivers/testing/login/login.service.ts
```diff
@@ -0,0 +1,21 @@
...
+  DaffAuthToken
+} from '@daffodil/auth';
+
+import { DaffAuthTokenFactory } from '../../../factories/auth-token.factory';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffTestingLoginService implements DaffLoginServiceInterface<DaffLoginInfo, DaffAuthToken> {
+  factory = new DaffAuthTokenFactory();
```
damienwebdev: Inject the factory.

libs/auth/testing/src/drivers/testing/register/register.service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffTestingRegisterService implements DaffRegisterServiceInterface<
+  DaffAccountRegistration<DaffCustomerRegistration>,
+  DaffCustomerRegistration,
+  DaffAuthToken
+> {
+  constructor (@Inject(DaffAuthTokenFactory) private factory: DaffAuthTokenFactory) {}
```
damienwebdev: Since this is an `@Injectable` you don't need the `@Inject`

griest024: 🤔 Clearly I misunderstand something about Angular DI. How does Angular know what class to inject for `DaffAuthTokenFactory`? Is it the `providedIn: 'root'` option on the factory `@Injectable` decorator that does it?

damienwebdev: In other instances, (where we use injection tokens) we're relying on an interface that doesn't exist in `js`-land, so we're forced to use an InjectionToken. In this instance, the class name exists in `js`-land, and combined with the `providedIn: 'root'` this providing is handled by the compiler IFF the token is used in the compilation (otherwise it gets tree-shaken out).

PR #593, opened 2020-02-03
feat(auth): add Magento GraphQL driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/auth/src/drivers/magento/login.service.ts
+++ b/libs/auth/src/drivers/magento/login.service.ts
@@ -1,0 +1,48 @@
+import { Injectable, Inject } from '@angular/core';
+import { Observable, pipe } from 'rxjs';
+import { map } from 'rxjs/operators';
+import { Apollo } from 'apollo-angular';
+
+import { DaffLoginServiceInterface } from '../interfaces/login-service.interface';
+import { DaffAuthQueryManager } from '../injection-tokens/auth-query-manager.token';
+import { DaffAuthQueryManagerInterface } from '../interfaces/auth-query-manager.interface';
+import { GenerateTokenResponse } from './models/outputs/generate-token-response';
+import { DaffLoginInfo } from '../../models/login-info';
+import { DaffAuthTransformer } from '../injection-tokens/auth-transformer.token';
+import { DaffAuthTransformerInterface } from '../interfaces/auth-transformer.interface';
+import { DaffAuthToken } from '../../models/auth-token';
+import { DaffAccountRegistration } from '../../models/account-registration';
+import { DaffCustomerRegistration } from '../../models/customer-registration';
+
+/**
+ * Util pipe to get the data from an apollo result
+ */
+const unwrapResult = pipe(
+  map(({data}) =>
+    data
+  )
+);
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoLoginService implements DaffLoginServiceInterface<DaffLoginInfo, DaffAuthToken> {
+  constructor(
+    private apollo: Apollo,
+    @Inject(DaffAuthQueryManager) public queryManager: DaffAuthQueryManagerInterface<
+      DaffAccountRegistration<DaffCustomerRegistration>,
+      DaffCustomerRegistration,
+      DaffLoginInfo
+    >,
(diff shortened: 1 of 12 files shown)


libs/auth/src/drivers/magento/login.service.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  login(loginInfo: DaffLoginInfo): Observable<DaffAuthToken> {
+    return this.apollo.mutate<GenerateTokenResponse>(
+      this.queryManager.generateATokenMutation(loginInfo)
+    ).pipe(
+      unwrapResult,
+      map(data => this.transformer.transform(data))
+    )
+  }
+}
```
lderrickable: I think this should be generic. It is sort of halfway to being generic with the `@Inject(DaffAuthTransformer) public transformer: DaffAuthTransformerInterface<DaffAuthToken>`. But there's not really a point in injecting the transformer service with injector tokens if this whole service isn't generic. If the app dev wants to extend the `DaffLoginInfo` model, then they will need to override the transformer AND this service. If this is generic, they'll only need to override the transformer. So like:
```
export class DaffMagentoLoginService implements DaffLoginServiceInterface<
   TLoginInfo extends DaffLoginInfo, 
   TAuthToken extends DaffAuthToken,
   TCustomerRegistration extends DaffCustomerRegistration,
   TAccountRegistration extends DaffAccountRegistration
> {
  constructor(
    private apollo: Apollo,
    @Inject(DaffAuthQueryManager) public queryManager: DaffAuthQueryManagerInterface<
      TAccountRegistration<TCustomerRegistration>,
      TCustomerRegistration,
      TLoginInfo
    >,
    @Inject(DaffAuthTransformer) public transformer: DaffAuthTransformerInterface<TAuthToken>
  ) {}

  login(loginInfo: DaffLoginInfo): Observable<TAuthToken> {
    return this.apollo.mutate<GenerateTokenResponse>(
      this.queryManager.generateATokenMutation(loginInfo)
    ).pipe(
      unwrapResult,
      map(data => this.transformer.transform(data))
    )
  }
}
```
Not sure if I got everything right in that, but something like that

libs/auth/src/drivers/magento/register.service.ts
```diff
@@ -0,0 +1,46 @@
...
+import { DaffLoginServiceInterface } from '../interfaces/login-service.interface';
+import { DaffLoginInfo } from '../../models/login-info';
+import { DaffAuthToken } from '../../models/auth-token';
+import { DaffCustomerRegistration } from '../../models/customer-registration';
+import { DaffAccountRegistration } from '../../models/account-registration';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoRegisterService implements DaffRegisterServiceInterface<
```
lderrickable: This should be generic too

libs/auth/src/drivers/magento/register.service.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  register(registration: DaffAccountRegistration<DaffCustomerRegistration>): Observable<DaffAuthToken> {
+    return this.apollo.mutate(
+      this.queryManager.createACustomerMutation(registration)
+    ).pipe(
+      mergeMap(() =>
+        this.loginDriver.login({
+          email: registration.customer.email,
+          password: registration.password
+        })
```
lderrickable: which means this should go in a transformer

libs/auth/src/drivers/magento/login.service.ts
griest024: Very good point. Making those changes now.

libs/auth/src/drivers/magento/register.service.ts
griest024: 👌

libs/auth/src/drivers/magento/register.service.ts
griest024: the argument should go into a transformer? So we need a `TAccountRegistration -> TLoginInfo` transformer?

PR #599, opened 2020-02-03
feat(auth): basic package w/ interfaces

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information


libs/auth/src/interfaces/auth-facade.interface.ts
```diff
@@ -0,0 +1,11 @@
+import { Observable } from 'rxjs';
+import { Action } from '@ngrx/store';
+
+import { DaffStoreFacade } from '@daffodil/core';
+
+export interface DaffAuthFacadeInterface extends DaffStoreFacade<Action> {
```
lderrickable: This should be generic, because it'll need have a property called `auth$: Observable<T>` where T extends DaffAuth. That way the facade is completely reusable in the case that the app dev provides their own model that extends `DaffAuth` and transformer and such. But it's probably a good idea to keep `token$`, since that won't change

lderrickable: Honestly, I don't think we even need an interface or token for the facade, since the token won't be referenced anywhere in daffodil, and if someone uses their own facade, there's really no reason for them to follow the pattern we've used. They should really never need to use anything but the facade we provide anyway

griest024: The injection token is useful for the mock facade in the testing section. The interface might still be useful for typing purposes.

lderrickable: In the testing section, you can just do 
```
providers: [
  { provide: DaffAuthFacade, useClass: MockDaffAuthFacade }
]
```
so the injection token isn't needed for that. Unless there's a particular instance you're thinking of where the interface can be useful, I'd say YAGNI it, because I can't think of anytime when it would be used.

damienwebdev: I agree with @griest024 about the interface. If we're going to provide a mock facade, we should be providing an interface to ensure that these remain compatible over time.

libs/auth/src/injection-tokens/auth-facade.token.ts
```diff
@@ -0,0 +1,3 @@
+import { InjectionToken } from '@angular/core';
+
+export const DaffAuthFacade = new InjectionToken('DaffAuthFacade');
```
damienwebdev: Do you need this token?

libs/auth/src/models/auth.ts
```diff
@@ -0,0 +1,3 @@
+export interface DaffAuth {
+  token: string;
```
damienwebdev: `DaffAuthToken` seems more appropriate.

libs/auth/src/models/account-registration.ts
```diff
@@ -0,0 +1,6 @@
+import { DaffCustomerRegistration } from './customer-registration';
+
+export interface DaffAccountRegistration {
```
damienwebdev: This should probably be a generic `<T extends DaffCustomerRegistration>`

libs/auth/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "directory": "../../dist/auth"
+  },
+  "scripts": {
+    "build": "ng build auth --prod",
+    "lint": "ng lint auth",
+    "test": "ng test auth --watch=false --browsers=ChromeHeadless",
+    "publish": "(cd ../../dist/auth; npm publish --access=public)"
+  },
+  "homepage": "https://github.com/graycoreio/daffodil",
+  "peerDependencies": {
```
damienwebdev: Most of these deps aren't real yet, and should not be listed here.

libs/auth/src/injection-tokens/auth-facade.token.ts
griest024: It was useful for implementing the mock facade before the facade itself. I thought there was a reason why I had to create the mock facade first but I can't seem to find it now. I'll move the mock facade later and remove the injection token.

libs/auth/src/interfaces/auth-facade.interface.ts
griest024: I erroneously believed that I had to define the mock facade before the real facade.

lderrickable:
Just needs the build fixed, then lgtm

libs/auth/package.json
griest024: Which ones can be removed? Should they be listed somewhere else?

griest024: Oh do you mean only add them to `package.json` once they are actually needed?

PR #601, opened 2020-02-03
docs(auth): add usage docs

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/auth/README.md
+++ b/libs/auth/README.md
@@ -12,3 +12,35 @@ npm install @daffodil/auth \
     --save
 ```
 
+## Usage
+This package does most of the setup for you.
+
+1. Import the `DaffAuthModule` into your app.
+2. Provide the drivers.
+    1. Use default drivers by importing `DaffAuthDriverModule.forRoot()`.
+    2. Provide your own `DaffLoginDriver` and/or `DaffRegisterDriver`.
+
+### Register
+
+Dispatch the `DaffAuthRegister` action with `DaffRegisterRequest` to register a new user. This will automatically log the user in.
+
+### Login
+
+Dispatch the `DaffAuthLogin` action with `DaffLoginRequest` to log the user in.
+
+## Extension and Customization
+If you feel like what our services return just won't cut it for your business needs, no problem! All of the reducers, effects, selectors, actions, and facades are 100% reusable with a customized model and service.
+
+### Login
+If the `DaffLoginResponse` doesn't have all of the necessary fields you require, you'll need to provide two things.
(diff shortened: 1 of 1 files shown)


lderrickable:
@damienwebdev and I were talking about how providing custom drivers and transformers would work, and it turns out that the driver service almost never needs to be overridden. Usually the only thing an app dev will want that is custom is specific data from a backend platform, which means they need only custom models and custom transformers. So as long as there exists generic transformer interfaces and a generic interface and instance of a driver, the app dev should only need to override the transformer. I just mention this here, because it might make more sense for the readme to say that they usually only need to override the transformers and extend the daffodil models.

griest024:
@lderrickable Should I mention the service customization as an optional step or omit it entirely?

lderrickable:
TLDR: no just omit it.

I think we omit for now, because if they're overriding the driver service, they should have a really good idea what they're doing. If they override the service (presumably to make a completely different service call), then I'm not sure they'll even be able to use the redux state we provide as it exists now. Damien has been talking about possibly getting rid of the actions and effects layer in daffodil and allow the user to provide that for better action hygiene. And maybe there'll be a more refined process for providing a custom driver, but for now it doesn't really make sense to override it.

libs/auth/README.md
```diff
@@ -12,3 +12,35 @@ npm install @daffodil/auth \
...
+### Login
+If the `DaffLoginResponse` doesn't have all of the necessary fields you require, you'll need to provide two things.
+1. A custom service that implements the `DaffAuthTransformerService`. This should be injected with the `DaffAuthTransformer` injection token to override the default transformer.
+2. Define a model that extends the `DaffLoginResponse` and/or the `DaffLoginRequest`.
+
+These steps should allow you to make supported requests to login in whatever way you'd like, while still being compatible with the daffodil redux state, because all redux components are generically defined.
+
+### Register
+If the `DaffRegisterResponse` doesn't have all of the necessary fields you require, you'll need to provide two things.
+1. A custom service that implements the `DaffAuthTransformerService`. This should be injected with the `DaffAuthTransformer` injection token to override the default transformer.
```
lderrickable: This seems wrong. Both login and register don't share the same transformer token and interface, right?

griest024: `register` calls `login` and returns whatever `login` returns, so in this case, they do "share" it. However since `register` doesn't explicitly invoke the transformer, its probably not correct to talk about it in the `register` section here.

It might be useful to move this after the driver interface PR.

damienwebdev:
I don't think that this is sufficient docs to explain the use case.

PR #608, opened 2020-02-05
feat(checkout): replace DaffCart type with Order type

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, the several types in `@daffodil/checkout` strictly return a hastily chosen type.

## What is the new behavior?
 This opts for a more sane choice. The `Order` type makes more sense for many types.

```
[x] Yes
[ ] No
```

Applications depending on the old type will need to shift to the new one.

## Other information

### DIFF
--- a/libs/checkout/src/order/effects/order.effects.spec.ts
+++ b/libs/checkout/src/order/effects/order.effects.spec.ts
@@ -4,19 +4,22 @@ import { Observable, of } from 'rxjs';
 import { hot, cold } from 'jasmine-marbles';
 
 import { DaffCart, DaffCartDriver, DaffCartServiceInterface } from '@daffodil/cart';
-import { DaffCartFactory, DaffTestingCartService } from '@daffodil/cart/testing';
-import { DaffTestingCheckoutService } from '@daffodil/checkout/testing';
+import { DaffTestingCartService, DaffCartFactory } from '@daffodil/cart/testing';
+import { DaffTestingCheckoutService, DaffOrderFactory } from '@daffodil/checkout/testing';
 
 import { OrderEffects } from './order.effects';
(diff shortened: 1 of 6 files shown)


libs/checkout/src/order/effects/order.effects.spec.ts
```diff
@@ -4,19 +4,22 @@ import { Observable, of } from 'rxjs';
...
-import { DaffCartFactory, DaffTestingCartService } from '@daffodil/cart/testing';
-import { DaffTestingCheckoutService } from '@daffodil/checkout/testing';
+import { DaffTestingCartService, DaffCartFactory } from '@daffodil/cart/testing';
+import { DaffTestingCheckoutService, DaffOrderFactory } from '@daffodil/checkout/testing';
 
 import { OrderEffects } from './order.effects';
 import { DaffPlaceOrder, DaffPlaceOrderSuccess, DaffPlaceOrderFailure } from '../actions/order.actions';
 import { DaffCheckoutServiceInterface } from '../../drivers/interfaces/checkout-service.interface';
 import { DaffCheckoutDriver } from '../../drivers/injection-tokens/driver-checkout.token';
+import { Order } from '@daffodil/checkout';
```
lderrickable: Shouldn't this give you a circular dependency?

damienwebdev: Yes, it should. But, since we do a build before we run tests, it doesn't since the token is available from the package.

damienwebdev: Fixed in 3faccf704b20df758896152ff5f4b1c21b7012db

PR #609, opened 2020-02-05
build(contact): updated exports for daffodil/contact

Nolan-Arnold:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information


libs/contact/src/effects/contact.effects.spec.ts
```diff
@@ -1,12 +1,12 @@
 import { Observable, of } from 'rxjs';
 import { DaffContactEffects } from './contact.effects';
-import { DaffContactUnion, DaffContactServiceInterface } from '@daffodil/contact';
 import { TestBed } from '@angular/core/testing';
-import { DaffTestingContactService } from 'libs/contact/testing/src/drivers/testing/contact.service';
-import { DaffContactDriver } from '../driver/injection-tokens/contact-driver.token';
 import { provideMockActions } from '@ngrx/effects/testing';
 import { DaffContactSubmit, DaffContactSuccessSubmit, DaffContactFailedSubmit, DaffContactRetry, DaffContactCancel } from '../actions/contact.actions';
 import { hot, cold } from 'jasmine-marbles';
+import { DaffTestingContactService } from 'libs/contact/testing/src';
```
lderrickable: This import is wrong

libs/contact/src/driver/hubspot/public_api.ts
```diff
@@ -0,0 +1,3 @@
+export { DaffContactHubspotTransformer } from './transformers/contact.transformer';
+export { DaffContactHubspotService } from './contact.service';
```
lderrickable: I don't think we want to export either of these concrete services, because we don't want the app dev ever to import them or try something weird like:
```
{provide: DaffContactHubspotService, useClass: SomeOtherCustomService}
```
We want them to implement the corresponding interfaces instead.

libs/contact/testing/src/index.ts
```diff
@@ -1 +1,2 @@
-export {}
\ No newline at end of file
+export { DaffTestingContactService } from './drivers/testing/contact.service';
+export { DaffInMemoryContactService } from './drivers/in-memory/contact.service';
```
lderrickable: These also shouldn't be exported. There should be a module for each of them that provides them via the corresponding token. Like
```
@NgModule({
  imports: [
    CommonModule
  ]
})
export class DaffContactTestingDriverModule {
  static forRoot(): ModuleWithProviders {
    return {
      ngModule: DaffContactTestingDriverModule,
      providers: [
        {
          provide: DaffContactDriver,
          useExisting: DaffTestingContactService
        }
      ]
    };
  }
}
```
Or something like that.

Nolan-Arnold:
@lderrickable Okay, all of your suggested changes should be made in f23eb4ac5462677f4ce28815f7982a61d1f17bdc

libs/contact/testing/src/index.ts
Nolan-Arnold: So I believe the testing service still needs to be exported, as its purpose is to test different aspects in the base `daffodil/contact` library.

damienwebdev: This isn't correct. you should be importing the `DaffContactTestingDriverModule` in the relevant test.

libs/contact/testing/src/drivers/testing/testing.module.ts
```diff
@@ -0,0 +1,18 @@
...
+      ngModule: DaffContactTestingDriverModule,
+      providers: [
+        {
+          provide: DaffContactDriver,
+          useExisting: DaffTestingContactService
+        }
+      ]
+    };
+  }
+}
```
lderrickable: I think you're going to want one of these for the inmemory driver too

libs/contact/testing/src/index.ts
Nolan-Arnold: Okay I fixed the `effects.spec` that was affected by this change in cd4422f5ca68cd864ae0fb22902b7a88e7ef84cc. Let me know y'alls thoughts because I had a surprisingly tricky time getting it to work without the exported service.

libs/contact/testing/src/drivers/testing/testing.module.ts
Nolan-Arnold: Good call, addressed in cd4422f5ca68cd864ae0fb22902b7a88e7ef84cc

libs/contact/testing/src/index.ts
Nolan-Arnold: All other daffodil packages are testing using an exported testing service so this also may be something we should look into for other packages.

Nolan-Arnold:
Okay this latest commit(9d75626ea1fa97c15511fcaa504294c6cc1e49e7) should address all the changes we discussed this morning. @lderrickable @damienwebdev

libs/contact/src/effects/contact.effects.spec.ts
```diff
@@ -1,28 +1,31 @@
...
   const mockForum = { firstName: 'John', lastName: 'Doe' };
-  let daffContactDriver: DaffContactServiceInterface<DaffContactUnion, any>;
+  let daffContactDriver;
 
   beforeEach(() => {
     TestBed.configureTestingModule({
       providers: [
+        DaffContactEffects,
+        provideMockActions(() => actions$),
         {
```
lderrickable: this should be done through the DriverTestingModule or whatever it's called. Should work correctly with a relative path, right?

Nolan-Arnold: So the `DaffContactTestingDriverModule` still doesn't work correctly with a relative path because the exported module is using the `DaffContactDriver` from dist instead of the one from the source code so the .spec will treat them as two separate tokens. I figured it was okay in this scenario to use the service directly because it is now imported directly from daffodil source code instead of following bad practice of using a daffodil exported service.

damienwebdev: I've resolved this. @lderrickable take a look at the new `tsconfig.json`.

libs/contact/src/effects/contact.effects.spec.ts
```diff
@@ -1,104 +1,113 @@
...
-  let effects: DaffContactEffects<DaffContactUnion, any>;
-  const mockForum = { firstName: 'John', lastName: 'Doe' };
-  let daffContactDriver: DaffContactServiceInterface<DaffContactUnion, any>;
+import {
+	DaffContactSubmit,
+	DaffContactSuccessSubmit,
+	DaffContactFailedSubmit,
+	DaffContactRetry,
+	DaffContactCancel,
+	DaffContactDriver,
```
Nolan-Arnold: So is this the idea here to use the same driver by having both modules use the code from `dist`?

damienwebdev: From `src` but, yes.

lderrickable: I thought we had all agreed not to do this. That we would use relative paths instead.

libs/contact/src/effects/contact.effects.spec.ts
lderrickable: That's not how we'd agreed to address this. Did something change?

libs/contact/src/effects/contact.effects.spec.ts
lderrickable: nvm looks fine

lderrickable:
lgtm

PR #611, opened 2020-02-05
fix(category): fix bug where category inmemory service would delete p…

lderrickable:
…roducts from the product inmemory service

## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The category inmemory api was splicing out the products in the actual product inmemory api.

## What is the new behavior?
The category inmemory api has its own copy of the products when it generates its own list of category product ids.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/category/testing/src/inmemory-backend/category.service.ts
+++ b/libs/category/testing/src/inmemory-backend/category.service.ts
@@ -37,7 +37,6 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
 
   get(reqInfo: any) {
 		const allCategoryProductIds = this.generateProductIdSubset(this.productInMemoryBackendService.products);
-		console.log(allCategoryProductIds)
 
     this.categoryPageConfigurationState = this.categoryPageConfigurationFactory.create({
 			id: reqInfo.id,
--- a/libs/category/testing/src/inmemory-backend/category.service.ts
+++ b/libs/category/testing/src/inmemory-backend/category.service.ts
@@ -77,8 +77,8 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
 	}
 
 	private generateProductIdSubset(products: DaffProduct[]): string[] {
-		const arraySize = Math.floor(Math.random() * Math.floor(products.length/2) + Math.floor(products.length/2));
-		return randomSubset(products, arraySize).map(product => product.id);
+		const subsetSize = Math.floor(Math.random() * Math.floor(products.length/2) + Math.floor(products.length/2));
+		return randomSubset(products, subsetSize).map(product => product.id);
 	}
 
 	private generatePageSize(reqInfo) {
--- a/libs/category/testing/src/inmemory-backend/category.service.ts
+++ b/libs/category/testing/src/inmemory-backend/category.service.ts
@@ -52,6 +52,8 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
 		});
 
     return reqInfo.utils.createResponse$(() => {
+			console.log(this.category.productIds);
+			console.log(this.productInMemoryBackendService.products);
       return {
         body: {
           category: this.category,
@@ -77,13 +79,14 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
 
 	private generateProductIdArray(products: DaffProduct[]) {
 		const arraySize = Math.floor(Math.random() * Math.floor(products.length/2) + Math.floor(products.length/2));
+		const availableProducts = [...products];
 		const productIdArray = [];
 		for(let i=0; i<arraySize; i++) {
-			const productIndex = Math.floor(Math.random() * products.length);
+			const productIndex = Math.floor(Math.random() * availableProducts.length);
 			productIdArray.push(
-				products[productIndex].id
+				availableProducts[productIndex].id
 			);
-			products.splice(productIndex, 1);
+			availableProducts.splice(productIndex, 1);
 		}
 
 		return productIdArray;
--- a/libs/category/testing/src/inmemory-backend/category.service.ts
+++ b/libs/category/testing/src/inmemory-backend/category.service.ts
@@ -37,6 +37,7 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
 
   get(reqInfo: any) {
 		const allCategoryProductIds = this.generateProductIdSubset(this.productInMemoryBackendService.products);
+		console.log(allCategoryProductIds)
 
     this.categoryPageConfigurationState = this.categoryPageConfigurationFactory.create({
 			id: reqInfo.id,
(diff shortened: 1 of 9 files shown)


libs/category/testing/src/inmemory-backend/category.service.ts
```diff
@@ -52,6 +52,8 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
 		});
 
     return reqInfo.utils.createResponse$(() => {
+			console.log(this.category.productIds);
```
damienwebdev: Remove.

libs/category/testing/src/inmemory-backend/category.service.ts
```diff
@@ -77,13 +79,14 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
 
 	private generateProductIdArray(products: DaffProduct[]) {
```
damienwebdev: Can some of this randomization logic be extracted? It took a while for me to understand what precisely was going on in this method.

libs/category/testing/src/inmemory-backend/category.service.ts
lderrickable: done

libs/category/testing/src/inmemory-backend/category.service.ts
lderrickable: put in core

libs/core/src/utils/random-subset.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { randomSubset } from './random-subset';
+
+describe('Core | Utils | random-subset', () => {
+	it('should return a random subset of the given array', () => {
```
damienwebdev: Needs test coverage for the base case.

libs/core/src/utils/random-subset.spec.ts
```diff
@@ -0,0 +1,10 @@
...
+		const initialArray = [1,2,3,4,5,6,7,8,9];
+		const result = randomSubset(initialArray, 7);
+		expect(initialArray.includes(result[0])).toBeTruthy();
+		expect(initialArray.includes(result[result.length-1])).toBeTruthy();
+	});
```
damienwebdev: You should make an assertion that each element of the result exists in the original set (definition of subset).

You should also make an assertion that running this twice will likely give two different results.

You could also write a test covering the probability distribution of small case (3 elements) and assert that over a decent number of calls, the distribution of all options is equally likely.

libs/category/testing/src/inmemory-backend/category.service.ts
```diff
@@ -63,30 +64,21 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
     });
 	}
 	
-	private getTotalPages(allIds: number[], pageSize: number) {
-		return Math.floor(allIds.length / pageSize) + 1;
+	private getTotalPages(allIds: string[], pageSize: number) {
+		return allIds.length % pageSize ? Math.floor(allIds.length / pageSize) + 1: allIds.length/pageSize;
```
damienwebdev: This can just be replaced with `Math.ceiling(allIds.length/pageSize);`

libs/category/testing/src/inmemory-backend/category.service.ts
```diff
@@ -63,30 +64,21 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
...
 	}
 
-	private trimProductIdsToSinglePage(allIds: number[], currentPage: number, pageSize: number) {
+	private trimProductIdsToSinglePage(allIds: string[], currentPage: number, pageSize: number) {
```
damienwebdev: We don't care about the type. This can either be any, or you can make this generic.

libs/category/testing/src/inmemory-backend/category.service.ts
```diff
@@ -63,30 +64,21 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
...
-			const productIndex = Math.floor(Math.random() * products.length);
-			productIdArray.push(
-				products[productIndex].id
-			);
-			products.splice(productIndex, 1);
-		}
-
-		return productIdArray;
+	private generateProductIdSubset(products: DaffProduct[]): string[] {
+		const subsetSize = Math.floor(Math.random() * Math.floor(products.length/2) + Math.floor(products.length/2));
```
damienwebdev: I don't think we should be guaranteeing the size here. I'd like to be able to get small, or 0 products in a response.

libs/core/src/utils/random-slice.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { randomSlice } from './random-slice';
+
+describe('Core | Utils | random-slice', () => {
+	it('should return a slice of the given array', () => {
```
damienwebdev: Neat.

libs/core/src/utils/random-subset.spec.ts
lderrickable: done

libs/core/src/utils/random-subset.spec.ts
lderrickable: I did #1 (before I had tried doing initialArray.forEach I think, and for some reason jasmine doesn't see the expects inside the foreach loop. A regular for loop seems to work though).

#2 I can't really do without doing #3. But that seems kind of extra.

libs/category/testing/src/inmemory-backend/category.service.ts
lderrickable: done

libs/category/testing/src/inmemory-backend/category.service.ts
lderrickable: done

libs/category/testing/src/inmemory-backend/category.service.ts
lderrickable: done

libs/core/src/utils/random-subset.spec.ts
damienwebdev: Ok, I'm ok with skipping 2 and 3 for now.

libs/category/testing/src/inmemory-backend/category.service.ts
```diff
@@ -35,7 +36,8 @@ export class DaffInMemoryBackendCategoryService implements InMemoryDbService {
   }
 
   get(reqInfo: any) {
-		const allCategoryProductIds = this.generateProductIdArray(this.productInMemoryBackendService.products);
+		const allCategoryProductIds = this.generateProductIdSubset(this.productInMemoryBackendService.products);
+		console.log(allCategoryProductIds)
```
damienwebdev: Remove.

lderrickable: oops. removed

PR #589, opened 2020-02-03
feat(cart): add interfaces to support basic cart operations

damienwebdev:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?
Currently, we're missing many of the basic cart operations that exist. 

Fixes: N/A

## What is the new behavior?
This adds interfaces for many of the basic cart operations. Drivers for:
1. Testing
2. In-memory
3. Magento

will follow in subsequent PRs.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

PR #616, opened 2020-02-07
feat(cart): add interfaces to expand cart driver capabilities

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

We need additional service interfaces to support additional cart operations.

I closed #589 because it was overly broad.

Fixes: N/A

## What is the new behavior?
This adds interfaces to support more features left out of #537 

```
[ ] Yes
[x] No
```

## Other information


libs/cart/src/drivers/interfaces/transformers/cart-transformer.interface.ts
```diff
@@ -0,0 +1,16 @@
...
+import { InjectionToken } from '@angular/core';
+
+/**
+ * An interface that must be implemented by cart transformer services.
+ * 
+ * This transformer is responsible for transforming the returned
+ * cart from a given platform into a Generic DaffCart.
+ */
+export interface DaffCartTransformerInterface<RES, T extends DaffCart> {
+  transform(response: RES): T;
```
griest024: What's your reason for using a generic `transform` parameter for just this transformer?

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
```diff
@@ -0,0 +1,9 @@
+import { InjectionToken } from '@angular/core';
+
+export interface DaffCartPaymentMethodsServiceInterface {
+	list(): any[];
```
griest024: Do you think this should be using generics for the return type? I know the payment method will probably be an enum or something in the future but the service interfaces should be as specific as we can make them. I feel like `T extends string` is better than `any`.

libs/cart/src/drivers/interfaces/cart-item-service.interface.ts
```diff
@@ -0,0 +1,14 @@
+import { DaffCartItem } from '../../models/cart-item';
+import { Observable } from 'rxjs';
+import { InjectionToken } from '@angular/core';
+
+export interface DaffCartItemServiceInterface<T extends DaffCartItem> {
+	list(): Observable<T[]>;
+	get(id: T['item_id']): Observable<T>;
+	update(itemId: T['item_id'], changes: any): Observable<T>;
```
griest024: What would `changes` be here? Just the cart item itself? This should be typed if possible.

griest024: I would do what you did with the cart payment service and do `changes: T`.

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
lderrickable: `T extends string` doesn't make sense to me. Isn't that meaningless?

libs/cart/src/drivers/interfaces/cart-payment-service.interface.ts
```diff
@@ -0,0 +1,13 @@
+import { InjectionToken } from '@angular/core';
+import { DaffCartPaymentMethod } from '../../models/cart-payment';
+
+export interface DaffCartPaymentServiceInterface<T extends DaffCartPaymentMethod> {
+	get(): T;
+	add(payment: T): T;
+	update(id: string | number, payment: T): T;
+	delete(id?: string | number): void;
+}
```
lderrickable: Shouldn't `get()` be called with a cartId?

I would expect `add` to return a cart. Otherwise, we'll need to edit the cart in state with the PaymentMethod rather than just getting the entire cart back and replacing the whole cart state (or make a separate call). Seems more correct to let the backend do all the editing of the cart object. Same thing with `update()`

libs/cart/src/drivers/interfaces/cart-service.interface.ts
```diff
@@ -1,9 +1,14 @@
...
 import { Observable } from 'rxjs';
-
 import { DaffCart } from '../../models/cart';
+import { DaffCartServiceInterface } from '../interfaces/cart-service.interface';
 
 export interface DaffCartServiceInterface<T extends DaffCart> {
-  get(): Observable<T>;
-  addToCart(productId: string, qty: number): Observable<T>;
-  clear(): Observable<void>;
+	get(): Observable<T>;
```
lderrickable: Doesn't `get()` need a cart id for an input?

libs/cart/src/drivers/interfaces/cart-service.interface.ts
```diff
@@ -1,9 +1,14 @@
...
+	addToCart(productId: string, qty: number): Observable<T>;
```
lderrickable: Doesn't this need a cart id too?

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
damienwebdev: I didn't do a generic here because I don't really have a good understanding of what the type may be at the moment.

griest024: @lderrickable its future-proofing for when it is replaced by a real type. Not really necessary though and might make it more confusing to read.

libs/cart/src/drivers/interfaces/cart-service.interface.ts
griest024: There should only ever be one cart in the client so we could query the cart id from a service `this.cartService.cartId.pipe(...`. However passing in the cart id would make the driver implementation simpler.

lderrickable: I don't think the driver should access redux state. The effects layer should call the driver with state if need be.

damienwebdev: I agree to that. It makes sense for the driver to be a clear "in-and-out" paradigm without having to reach into various aspects of Daffodil.

libs/cart/src/drivers/interfaces/cart-service.interface.ts
damienwebdev: Yes, added.

libs/cart/src/drivers/interfaces/cart-service.interface.ts
damienwebdev: @lderrickable I added the `cart_id`.

libs/cart/src/drivers/interfaces/cart-payment-service.interface.ts
damienwebdev: Yes, added.

damienwebdev: `add` for a payment method (and the other various `add`s) could return the new cart or the updated object. I have no preference for one over the other.

libs/cart/src/drivers/interfaces/cart-item-service.interface.ts
damienwebdev: @griest024 I've typed this a Partial, let me know what you think.

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
damienwebdev: I've addressed this, LMK what you think.

libs/cart/src/drivers/interfaces/transformers/cart-transformer.interface.ts
damienwebdev: No specific reason. I'm still a bit uncertain about typing transformers at the moment.

damienwebdev: I think typing arguments is probably useless, and we should be only typing the return types.

libs/cart/src/drivers/interfaces/cart-billing-address-service.interface.ts
```diff
@@ -0,0 +1,25 @@
...
+> {
+	/**
+	 * Retrieve the billing address of a cart
+	 */
+	get(cartId: DaffCart['id']): V;
+
+	/**
+	 * Update the billing address of a cart
+	 */
+	update(cartId: DaffCart['id'], address: Partial<V>): V;
```
griest024: Shouldn't these return `Observable`s?

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
```diff
@@ -0,0 +1,17 @@
...
+import { DaffCart } from '../../models/cart';
+
+/**
+ * The interface responsible for retrieving the available payment methods of a cart.
+ */
+export interface DaffCartPaymentMethodsServiceInterface<T extends DaffCartPaymentMethod> {
+	/**
+	 * List the available payment methods of a cart.
+	 */
+	list(cartId: DaffCart['id']): T[];
```
griest024: This should probably return `Observable<T[]>`.

libs/cart/src/drivers/interfaces/cart-payment-service.interface.ts
```diff
@@ -0,0 +1,27 @@
...
+
+	/**
+	 * Update the payment method applied to a cart. 
+	 */
+	update(cartId: DaffCart['id'], payment: Partial<T>): T;
+
+	/**
+	 * Remove the payment method applied to a cart.
+	 */
+	remove(cartId: DaffCart['id']): void;
```
griest024: `Observable` here as well.

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
griest024: lgtm

libs/cart/src/drivers/interfaces/cart-item-service.interface.ts
griest024: lgtm

libs/cart/src/drivers/interfaces/cart-billing-address-service.interface.ts
damienwebdev: Yes, fixed, nice catch.

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
damienwebdev: Yes, fixed, nice catch.

libs/cart/src/drivers/interfaces/cart-payment-service.interface.ts
damienwebdev: Yes, fixed, nice catch.

lderrickable:
I still think that returning the cartitem from `AddToCart()` (and other methods like it), might pose a problem. Because in order to update our local version of the cart, we'd need either to update it ourselves when the `AddToCart` call is successful or we'll need to make a new `getCart` call when the `AddToCart` call is successful. Maybe there's something I'm missing, but I don't see how we'll have a good way to get the updated cart (without making another service call), whereas making `AddToCart` return the updated cart would solve this.

libs/cart/src/models/cart-addition.ts
```diff
@@ -0,0 +1,4 @@
+export interface DaffCartItemInput {
+  productId: string;
+  qty: number;
+}
```
griest024: This file name should match the export.

libs/cart/src/drivers/interfaces/cart-shipping-information-service.interface.ts
```diff
@@ -0,0 +1,30 @@
...
+import { Observable } from 'rxjs';
+
+import { DaffCartShippingRate } from '../../models/cart-shipping-rate';
+import { DaffCart } from '../../models/cart';
+
+/**
+ * The interface responsible for mediating the interaction of the shipping
+ * information of a cart with a given platform.
+ */
+export interface DaffCartShippingInformationServiceInterface<T extends DaffCartShippingRate>{
```
griest024: Should the generic extend from `DaffCartShippingInformation` instead?

griest024:
@lderrickable I agree. The full cart would need to be retrieved every time an update is made.

The only argument I can think of for not automatically returning the cart is if multiple updates are made in succession without needing the updated cart in between those updates. Given that returning the cart from methods like `CartItem#add` could require an extra call to `Cart#get` anyway, we could eliminate an extra call in the best case by leaving it up to the user to call `Cart#get` when needed.

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
```diff
@@ -0,0 +1,19 @@
...
+import { DaffCart } from '../../models/cart';
+
+/**
+ * The interface responsible for retrieving the available payment methods of a cart.
+ */
+export interface DaffCartPaymentMethodsServiceInterface<T extends DaffCartPaymentMethod> {
+	/**
+	 * List the available payment methods of a cart.
+	 */
+	list(cartId: DaffCart['id']): Observable<T>[];
```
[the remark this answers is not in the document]
damienwebdev: Fixed in 09bdf188, was a mistake.

libs/cart/src/models/cart-addition.ts
damienwebdev: Fixed in 86133bc8

libs/cart/src/drivers/interfaces/cart-shipping-information-service.interface.ts
damienwebdev: I don't know what this type really means yet, so I'm electing for the less explicit type to start.

libs/cart/src/drivers/interfaces/cart-payment-methods-service.interface.ts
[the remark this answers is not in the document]
damienwebdev: Agreed, it was a mistake. Fixed.

libs/cart/src/models/cart-addition.ts
damienwebdev: Fixed.

libs/cart/src/drivers/interfaces/cart-shipping-information-service.interface.ts
damienwebdev: I don't what this type truly looks like yet, so I'm opting for the less explicit type.

damienwebdev:
@lderrickable @griest024 I think there's two ways to solve this problem...

1. Via the driver return (in-combination with multiple driver methods)
2. Via additional reactions via effects.

I think I prefer option 1 because it keeps the UI "updating" logic away from the platform interaction. 

I can go either way personally since I don't have great reasoning to choose either.

libs/cart/src/drivers/interfaces/cart-shipping-information-service.interface.ts
griest024: Got it, I think that's fine for now.

lderrickable:
I think that (hopefully) we usually wouldn't need to make multiple driver calls to return the cart. At least in magento, I think a lot of the "add something to cart" methods return a partial of a cart already in its graphQL api. Then we'd just update the parts of the cart returned. I don't know what this looks like in other platforms though. So in short, I'd prefer option 1 as well

griest024:
Option 1 should be fairly easy to implement in Shopify GraphQL as well.

damienwebdev:
@lderrickable @griest024 I've made the requested changes, and fixed a few breaks from the related changes.

This PR is definitely a breaking change, so the next release must be bumped.

lderrickable:
@damienwebdev wouldn't we do a minor version, since we're still in alpha? Does alpha go up to 1.x.x?

damienwebdev:
@lderrickable the "major" version would be `0.4.0` for pre-release semantics.

lderrickable:
lgtm

libs/cart/src/drivers/interfaces/cart-item-service.interface.ts
```diff
@@ -0,0 +1,48 @@
...
+ */
+export interface DaffCartItemServiceInterface<
+	T extends DaffCartItem,
+	U extends DaffCartItemInput,
+	V extends DaffCart
+> {
+	/**
+	 * List all of the available items of a cart
+	 */
+	list(cartId: DaffCart['id']): Observable<T[]>;
```
lderrickable: It might not _really_ matter, but shouldn't this be `V['id']`?

libs/cart/src/effects/cart.effects.ts
```diff
@@ -11,22 +11,23 @@ import {
...
-    @Inject(DaffCartDriver) private driver: DaffCartServiceInterface<T>) {}
+    @Inject(DaffCartDriver) private driver: DaffCartServiceInterface<T>,
+    private storage: DaffCartStorageService) {}
 
   @Effect()
   load$ = this.actions$.pipe(
     ofType(DaffCartActionTypes.CartLoadAction),
     switchMap((action: DaffCartLoad) =>
-      this.driver.get()
+      this.driver.get(this.storage.getCartId())
```
lderrickable: This should be tested

libs/cart/src/drivers/interfaces/cart-item-service.interface.ts
damienwebdev: fixed in 822f0063

libs/cart/src/effects/cart.effects.ts
damienwebdev: tested in 661acdd0

lderrickable:
lgtm

PR #623, opened 2020-02-12
feat(auth): add magento GraphQL driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/auth/src/drivers/magento/auth.service.ts
+++ b/libs/auth/src/drivers/magento/auth.service.ts
@@ -1,0 +1,29 @@
+import { Injectable } from '@angular/core';
+import { Observable, of, throwError } from 'rxjs';
+import { switchMap } from 'rxjs/operators';
+import { Apollo } from 'apollo-angular';
+
+import { DaffAuthServiceInterface } from '../interfaces/auth-service.interface';
+import {
+  checkTokenQuery,
+  MagentoCheckTokenResponse
+} from './queries/public_api';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoAuthService implements DaffAuthServiceInterface {
+  constructor(private apollo: Apollo) {}
+
+  check(): Observable<void> {
+    return this.apollo.query<MagentoCheckTokenResponse>({query: checkTokenQuery}).pipe(
+      switchMap(resp => {
+        if (resp.data.customer.id) {
+          return of(undefined);
+        } else {
+          return throwError(new Error('Unauthenticated'));
+        }
+      })
+    )
+  }
+}
--- a/libs/auth/src/drivers/magento/auth.service.ts
+++ b/libs/auth/src/drivers/magento/auth.service.ts
@@ -1,6 +1,6 @@
 import { Injectable } from '@angular/core';
 import { Observable, of, throwError } from 'rxjs';
-import { switchMap } from 'rxjs/operators';
+import { switchMap, mapTo } from 'rxjs/operators';
 import { Apollo } from 'apollo-angular';
 
 import { DaffAuthServiceInterface } from '../interfaces/auth-service.interface';
@@ -17,13 +17,7 @@ export class DaffMagentoAuthService implements DaffAuthServiceInterface {
 
   check(): Observable<void> {
     return this.apollo.query<MagentoCheckTokenResponse>({query: checkTokenQuery}).pipe(
-      switchMap(resp => {
-        if (resp.data.customer.id) {
-          return of(undefined);
-        } else {
-          return throwError(new Error('Unauthenticated'));
-        }
-      })
+      mapTo(undefined)
     )
   }
 }
--- a/libs/auth/src/drivers/magento/login.service.spec.ts
+++ b/libs/auth/src/drivers/magento/login.service.spec.ts
@@ -102,7 +102,7 @@ describe('Driver | Magento | Auth | LoginService', () => {
       });
     });
 
-    it('should return the correct observable', done => {
+    it('should return an auth token', done => {
       service.login(mockLoginInfo).subscribe((auth) => {
         expect(auth).toEqual(mockAuth);
         done();
@@ -147,7 +147,8 @@ describe('Driver | Magento | Auth | LoginService', () => {
       });
     });
 
-    describe('when the result is false', () => {
+    // TODO: add error handling
+    xdescribe('when the result is false', () => {
       beforeEach(() => {
         result = false;
         response = {
--- a/libs/auth/src/drivers/magento/login.service.spec.ts
+++ b/libs/auth/src/drivers/magento/login.service.spec.ts
@@ -1,0 +1,177 @@
+import { TestBed } from '@angular/core/testing';
+import {
+  ApolloTestingModule,
+  ApolloTestingController,
+} from 'apollo-angular/testing';
+import { cold } from 'jasmine-marbles';
+import { catchError } from 'rxjs/operators';
+
+import {
+  DaffAccountRegistrationFactory,
+  DaffAuthTokenFactory
+} from '@daffodil/auth/testing';
+
+import { DaffMagentoLoginService } from './login.service';
+import { DaffAccountRegistration } from '../../models/account-registration';
+import { DaffAuthToken } from '../../models/auth-token';
+import { DaffLoginInfo } from '../../models/login-info';
+import {
+  generateTokenMutation,
+  revokeCustomerTokenMutation,
+  MagentoRevokeCustomerTokenResponse
+} from './queries/public_api';
+import { DaffMagentoAuthTransformerService } from './transforms/auth-transformer.service';
+
+describe('Driver | Magento | Auth | LoginService', () => {
+  let controller: ApolloTestingController;
+  let service: DaffMagentoLoginService;
+
+  const authTransformerServiceSpy = jasmine.createSpyObj('DaffMagentoAuthTransformerService', ['transform'])
+
+  const registrationFactory: DaffAccountRegistrationFactory = new DaffAccountRegistrationFactory();
+  const authTokenFactory: DaffAuthTokenFactory = new DaffAuthTokenFactory();
+
+  let mockAuth: DaffAuthToken;
+  let mockLoginInfo: DaffLoginInfo;
+  let token: string;
+  let email: string;
+  let password: string;
+  let firstName: string;
+  let lastName: string;
+  let mockRegistration: DaffAccountRegistration;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [
+        ApolloTestingModule
+      ],
+      providers: [
+        DaffMagentoLoginService,
+        {
+          provide: DaffMagentoAuthTransformerService,
+          useValue: authTransformerServiceSpy
+        }
+      ]
+    });
+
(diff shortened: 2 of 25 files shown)


libs/auth/src/drivers/magento/register.service.ts
```diff
@@ -0,0 +1,55 @@
...
+      TCustomerRegistration,
+      TLoginInfo
+    >
+  ) {}
+
+  register(registration: TAccountRegistration): Observable<TAuthToken> {
+    return this.apollo.mutate(
+      this.queryManager.createACustomerMutation(registration)
+    ).pipe(
+      mergeMap<any, Observable<TAuthToken>>((): Observable<TAuthToken> =>
```
lderrickable: I think this should be a switchMap instead. MergeMap is for waiting for multiple observables.

libs/auth/src/drivers/magento/register.service.ts
```diff
@@ -0,0 +1,55 @@
...
+        this.loginDriver.login(this.loginInfoTransformer.transform(registration))
```
lderrickable: I think this should be done in the effects layer instead. Like when the `DaffRegisterSuccess` action is triggered, there should be an effect for firing a `DaffLogin` action.

libs/auth/src/drivers/magento/login.service.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+    //   });
+    // });
+
+    it('should return the correct observable', done => {
+      loginService.login(mockLoginInfo).subscribe((auth) => {
+        expect(auth).toEqual(mockAuth);
+        done();
+      });
+
+      const op = controller.expectOne(authGraphQlQueryManagerService.generateATokenMutation({email, password}).mutation);
```
lderrickable: probably use the mockLoginInfo here instead, since you already have it defined `generateATokenMutation(mockLoginInfo)`

libs/auth/src/drivers/magento/login.service.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+      response = {
+        generateCustomerToken: {
+          token
+        }
+      };
+
+      transformerServiceSpy.transform.withArgs(response).and.returnValue(mockAuth);
+    });
+
+    // TODO: determine if mocked tests are appropriate
```
lderrickable: You could do this if you want. Maybe do `expect(transformerServiceSpy.transform).toHaveBeenCalledWith(response.generateCustomerToken);` instead if you do though.

libs/auth/src/drivers/magento/register.service.ts
griest024: So maybe have the `register` method just return the customer ID or something like that?

libs/auth/src/drivers/magento/login.service.spec.ts
griest024: ah good catch

libs/auth/src/drivers/magento/transforms/auth-transformer.service.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+        token: 'sdaf87sadbnfnsd'
+      }
+    };
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+
+  describe('transform | transforming the token response into a DaffAuthToken', () => {
```
lderrickable: I think you should be an object comparison if possible:
```
let mockToken = 'sdaf87sadbnfnsd';
mockResponse = {
   generateCustomerToken: {
     token: mockToken
   }
};
expectedDaffAuthToken {
  token: mockToken
};

describe('transform', () => {
  it('should transform a magento token response into a DaffAuthToken', () => {
    const result: DaffAuthToken = service.transform(mockResponse);

    expect(result).toEqual(expectedDaffAuthToken);
  });
});
```

libs/auth/src/drivers/magento/transforms/login-info-transformer.service.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    mockRegistration = accountRegistrationFactory.create();
+  });
+
+  it('should be created', () => {
+    expect(service).toBeTruthy();
+  });
+
+  describe('transform | transforming the account registration into login info', () => {
+    it('should return a DaffAuthToken with the correct token field', () => {
```
lderrickable: I think this 'it' caption is wrong

griest024: yeah it is, I have fixed it.

libs/auth/src/drivers/magento/login.service.spec.ts
griest024: good advice

libs/auth/src/drivers/magento/transforms/auth-transformer.service.spec.ts
griest024: changed to this.

libs/auth/src/drivers/magento/register.service.ts
lderrickable: I guess it would need to return whatever is required for the login effect -> login call. So whatever that is.

griest024: @damienwebdev are you good with changing the register interface to return a `DaffLoginInfo`?

damienwebdev: Yes. I like @lderrickable 's breakdown here better. We could even make the effect an optional effect so that if people don't want that behavior, they won't get it.

libs/auth/src/drivers/magento/register.service.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    email = mockRegistration.customer.email;
+    password = mockRegistration.password;
+    mockLoginInfo = {email, password};
+  });
+
+  it('should be created', () => {
+    expect(registerService).toBeTruthy();
+  });
+
+  describe('register | getting login info', () => {
```
lderrickable: typo

libs/auth/src/drivers/magento/register.service.ts
griest024: this has been implemented. LMK what you think

griest024:
Code climate is failing my build for refactoring that doesn't need to happen. Can we exclude tests from refactor checking?

libs/auth/src/drivers/magento/auth.service.ts
```diff
@@ -0,0 +1,29 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoAuthService implements DaffAuthServiceInterface {
+  constructor(private apollo: Apollo) {}
+
+  check(): Observable<void> {
+    return this.apollo.query<MagentoCheckTokenResponse>({query: checkTokenQuery}).pipe(
+      switchMap(resp => {
+        if (resp.data.customer.id) {
```
damienwebdev, line 21: I would expect Magento to have a error code on the response, instead of checking a field. Is this not the case?

libs/auth/src/drivers/magento/login.service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  login({email, password}: DaffLoginInfo): Observable<DaffAuthToken> {
+    return this.apollo.mutate<MagentoGenerateTokenResponse>({
+      mutation: generateTokenMutation,
+      variables: {
+        email,
+        password
+      }
+    }).pipe(
+      map(res => this.authTransformer.transform(res.data.generateCustomerToken))
```
damienwebdev, line 34: How do we accommodate errors like if the username/password doesn't match?

libs/auth/src/drivers/magento/transforms/login-info-transformer.service.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Transforms magento auths into an object usable by daffodil.
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoLoginInfoTransformerService {
+  transform(registration: DaffAccountRegistration): DaffLoginInfo {
+    return registration ? {
```
damienwebdev, line 14: Why does this have a conditional?

libs/auth/src/drivers/magento/transforms/auth-transformer.service.ts
```diff
@@ -0,0 +1,23 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoAuthTransformerService {
+
+  /**
+   * Transforms the magento AuthNode from the magento auth query into a DaffAuthToken.
+   * @param auth the auth from a magento auth query.
+   */
+  transform(auth: MagentoAuth): DaffAuthToken {
```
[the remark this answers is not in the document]
griest024: I think null checking is better done inside the transformer since typescript doesn't prevent `null` from being passed into the transformers.

libs/auth/src/drivers/interfaces/register-service.interface.ts
```diff
@@ -1,19 +1,19 @@
 import { Observable } from 'rxjs';
 import { DaffCustomerRegistration } from '../../models/customer-registration';
 import { DaffAccountRegistration } from '../../models/account-registration';
-import { DaffAuthToken } from '../../models/auth-token';
+import { DaffLoginInfo } from '../../models/login-info';
 
 export interface DaffRegisterServiceInterface<
```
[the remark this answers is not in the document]
griest024: removed

libs/auth/src/drivers/interfaces/auth-query-manager.interface.ts
```diff
@@ -11,4 +11,6 @@ export interface DaffAuthQueryManagerInterface<
   generateATokenMutation(request: LoginRequest): MutationOptions<any>;
```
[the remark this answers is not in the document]
griest024: removed

libs/auth/src/drivers/magento/auth.service.ts
griest024: You are correct, I added error handling to this call/

libs/auth/src/drivers/magento/login.service.ts
griest024: I added error handling to the login driver

libs/auth/src/drivers/interfaces/auth-query-manager.interface.ts
```diff
@@ -0,0 +1,12 @@
+import { MutationOptions } from 'apollo-client';
```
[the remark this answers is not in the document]
griest024: I dunno why the rebase decided to add these back in.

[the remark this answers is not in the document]
griest024: changed

libs/auth/src/drivers/interfaces/auth-transformer.interface.ts
```diff
@@ -0,0 +1,11 @@
+import { DaffAuthToken } from '../../models/auth-token';
```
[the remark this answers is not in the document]
griest024: changed

PR #625, opened 2020-02-13
build(ci): remove codecov, its not good

damienwebdev:


drazisil:
@thomasrockhu, I was pinged on gitter. Can you take a look and assist, since I'm on vacation?

damienwebdev:
@drazisil @thomasrockhu the issue at hand here is that we have ~14 unique packages. I tried a number of things for about ~8 hours. Here are my comments on the matter:

1. The `codecov-node` module is inconsistent with the `codecov` bash uploader.
2. `codecov-node` doesn't actually utilize the provided `yaml` (or doesn't appear to)
3. If I've assigned flags in my `codecov.yml` file, I don't see why I have to additionally assign flags when I upload the report (I would expect that this should be inferred from my yaml).

PR #627, opened 2020-02-13
feat(auth): change LoginService#logout return type from void to boolean

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Returning `void` in observable chains will often cause them not to complete. This is hard to test and will cause issues in effects, among other places.

Fixes: N/A

## What is the new behavior?
Returning a defined value in the observable chain alleviates these issues.

```
[x] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/auth/src/drivers/interfaces/login-service.interface.ts
+++ b/libs/auth/src/drivers/interfaces/login-service.interface.ts
@@ -15,6 +15,8 @@ export interface DaffLoginServiceInterface<
 
   /**
    * Logs the user out by revoking their access token.
+   *
+   * @returns true if the token was revoked.
    */
   logout(): Observable<boolean>;
 }
--- a/libs/auth/src/drivers/interfaces/login-service.interface.ts
+++ b/libs/auth/src/drivers/interfaces/login-service.interface.ts
@@ -16,5 +16,5 @@ export interface DaffLoginServiceInterface<
   /**
    * Logs the user out by revoking their access token.
    */
-  logout(): Observable<void>;
+  logout(): Observable<boolean>;
 }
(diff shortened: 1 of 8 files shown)


libs/auth/src/drivers/interfaces/login-service.interface.ts
```diff
@@ -15,6 +15,8 @@ export interface DaffLoginServiceInterface<
 
   /**
    * Logs the user out by revoking their access token.
+   *
+   * @returns true if the token was revoked.
    */
-  logout(): Observable<void>;
```
damienwebdev: I would assume that this would throw an error into the stream if the token wasn't revoked. I'm not a big fan of this change.

griest024: That is how I currently have it implemented in my WIP magento driver. It could also return `false`.

damienwebdev: I don't see any issue with returning void.

griest024: Return `void` or throw an error if the revocation fails?

damienwebdev: What I'm saying is that there is no return type to this method. 

We log the user out. If it succeeds, great.

If we don't, we throw an error. A return value doesn't help in either of these situations.

griest024: Is `and.returnValue(of())` the correct way to mock the successful `logout` behavior?

damienwebdev: You could test/mock this in many different ways. I think people would tend towards depending on the testing service, but if you wanted to mock it, yes, you could mock it that way.

PR #661, opened 2020-03-02
feat(cart): add magento driver model factories

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
Adds the model factories required by the Magento cart shipping address driver.

```
[ ] Yes
[ ] No
```

## Other information
Relies on #665.


libs/cart/testing/src/factories/magento/product.factory.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  private money(): MagentoMoney {
+    return (new MagentoMoneyFactory()).create()
+  }
+}
+
+@Injectable({
+  providedIn: 'root'
+})
+export class MagentoProductFactory extends DaffModelFactory<MagentoProduct> {
```
lderrickable: I think this and the product model in @daffodil/cart is going to be a problem. Right now, the one in @daffodil/product is called `ProductNode`, but it should really be called `MagentoProduct`. It's just weird because the cart package is dependent on product package. Maybe this one could be called `MagentoCartItemProduct` or something.

griest024: Well really they are they same backend model. I think ideally we would have a product model that resembled the actual [Magento `ProductInterface`](https://devdocs.magento.com/guides/v2.3/graphql/product/product-interface.html) that both the cart and product drivers could use.

lderrickable: Sounds good to me. Want to remove the product factory here, then do a quick PR for adding it to the product package?

lderrickable: Or do you need to update the interface too?

griest024: I don't need to change the interface, `ProductNode` has everything that I need for `DaffCartItem`. I'll submit that PR in a sec.

griest024: `MagentoProductFactory` has been removed from this PR.

PR #664, opened 2020-03-02
feat(core, cart): move MagentoMoney from cart into core

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information


libs/core/src/magento/models/outputs/index.ts
```diff
@@ -0,0 +1 @@
+export { MagentoMoney } from './money';
```
lderrickable: let's use "requests" instead of "outputs". I was talking to Damien about this and he made some good points about the input-output model paradigm being a little confusing. Are they the outputs of a response or a transformer function or are they the outputs of daffodil toward a platform? It's a little more straightforward to just have them be requests and not requests instead of outputs and inputs.

griest024: Wouldn't the inputs be the requests and the outputs be the not requests?

griest024: changed

lderrickable:
lgtm

libs/core/src/index.ts
```diff
@@ -7,4 +7,5 @@ export { shuffle } from './utils/shuffle';
 export { randomSubset } from './utils/random-subset';
 export { randomSlice } from './utils/random-slice';
 
-export * from './storage/public_api';
\ No newline at end of file
+export * from './storage/public_api';
+export * from './magento';
```
damienwebdev: I think `drivers` is a way better place to put this than `core`.

griest024: So `driver/src/magento`? We would need to add a testing package to `driver`. FWIW I just realized that this isn't required for the `MagentoProduct` factory.

lderrickable: yea, you'd need to add the testing subpackage

damienwebdev:
@griest024 Is this PR is valid? Looks pretty stale. I'm ok with the intent of this PR so long as we move this to `@daffodil/drivers/magento`.

griest024:
@damienwebdev yeah its pretty stale. Updating now.

griest024:
@damienwebdev done

libs/core/src/index.ts
griest024: done

PR #666, opened 2020-03-03
Magento Bundled Product Query and models

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

Magento part is easy. I'm not entirely sure about the Daffodil part. I basically combined the magento BundledProductItemOption and the Product it has inside it into a BundledProductItemOptionProduct for daffodil. It seemed like the extra layer for the product was unnecessary.

### DIFF
--- a/libs/product/src/models/bundled-product-item.ts
+++ b/libs/product/src/models/bundled-product-item.ts
@@ -1,0 +1,25 @@
+import { DaffProduct } from "./product";
+
+export enum DaffPriceTypeEnum {
+	fixed = 'FIXED',
+	percent = 'PERCENT',
+	dynamic = 'DYNAMIC'
+}
+
+export interface DaffBundledProductItem {
+	bundle_id: string;
+	bundle_option_id: number;
+	required: boolean;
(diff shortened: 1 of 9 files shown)


libs/product/src/models/bundled-product-item.ts
```diff
@@ -0,0 +1,25 @@
...
+export interface DaffBundledProductItem {
+	bundle_id: string;
+	bundle_option_id: number;
+	required: boolean;
+	title: string;
+	type: string;
+	options: DaffBundledProductItemOption[];
+}
+
+export interface DaffBundledProductItemOption extends DaffProduct {
```
lderrickable: This is the tricky part, I think. Magento has this object have an additional `product` field, whereas I just made this object a simple product itself.

damienwebdev: I think this is ok as a starting point.

PR #654, opened 2020-02-27
feat(cart): add Magento payment methods driver

griest024:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Fixes: N/A

## What is the new behavior?
Adds a Magento driver that implements `DaffCartPaymentMethodsServiceInterface`.

## Does this PR introduce a breaking change?
```
[ ] Yes
[ ] No
```

## Other information

PR #668, opened 2020-03-03
feat(cart): add Magento GraphQL driver transformers

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
Adds a number of transformers that are needed for the shipping address driver.

```
[ ] Yes
[ ] No
```

## Other information
This is the last of the setup PRs needed for the remainder of the drivers.

Relies on #669.

### DIFF
--- a/libs/cart/src/drivers/magento/transforms/outputs/cart-item.service.ts
+++ b/libs/cart/src/drivers/magento/transforms/outputs/cart-item.service.ts
@@ -1,0 +1,36 @@
+import { Injectable } from '@angular/core';
+
+import { MagentoCartItem } from '../../models/outputs/cart-item';
+import { DaffCartItem } from '../../../../models/cart-item';
+
+/**
+ * Transforms magento carts into an object usable by daffodil.
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoCartItemTransformer {
+
+  /**
+   * Transforms the magento MagentoCartItem from the magento cart query into a DaffCartItem.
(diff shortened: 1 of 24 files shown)


libs/cart/src/drivers/magento/transforms/outputs/cart-item.service.ts
```diff
@@ -0,0 +1,36 @@
...
+})
+export class DaffMagentoCartItemTransformer {
+
+  /**
+   * Transforms the magento MagentoCartItem from the magento cart query into a DaffCartItem.
+   * @param response the response from a magento cart query.
+   */
+  transform(cartItem: MagentoCartItem): DaffCartItem {
+    return {
+      ...cartItem,
```
lderrickable: I see this in a lot of places, and I don't think we want this. Won't this just put a bunch of garbage on the DaffCartItem on state? Because we shouldn't assume that the magento and daffodil models will be the same.

griest024: This is to keep the transformers covariant as per [this discussion](https://github.com/graycoreio/daffodil/pull/654#discussion_r385334319) with Damien.

lderrickable: That make negative sense to me. I'll ask him about it

PR #671, opened 2020-03-05
feat(cart): add Magento GraphQL billing address driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Fixes some minor issues in unrelated specs
- Adds the billing address input transformer and driver

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/src/drivers/magento/cart-billing-address.service.ts
+++ b/libs/cart/src/drivers/magento/cart-billing-address.service.ts
@@ -65,7 +65,6 @@ export class DaffMagentoCartBillingAddressService implements DaffCartBillingAddr
     ).pipe(
       map(([updateResult, setEmailResult]) => this.cartTransformer.transform({
         ...updateResult.data.setBillingAddressOnCart.cart,
-        // only set the email from the set email result since we don't know if the billing address update is included in the response
         email: setEmailResult.data.setGuestEmailOnCart.cart.email
       }))
     )
--- a/libs/cart/src/drivers/magento/cart-billing-address.service.ts
+++ b/libs/cart/src/drivers/magento/cart-billing-address.service.ts
@@ -1,0 +1,73 @@
+import { Injectable } from '@angular/core';
+import { Apollo } from 'apollo-angular';
+import { Observable, zip } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+import { DaffCartBillingAddressServiceInterface } from '../interfaces/cart-billing-address-service.interface';
+import { DaffCart } from '../../models/cart';
+import { DaffCartAddress } from '../../models/cart-address';
+import {
+  getBillingAddress,
+  updateBillingAddress,
+  setGuestEmail
+} from './queries';
+import {
+  MagentoGetBillingAddressResponse,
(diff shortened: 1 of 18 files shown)


libs/cart/src/drivers/magento/cart-billing-address.service.ts
```diff
@@ -0,0 +1,73 @@
...
+        mutation: setGuestEmail,
+        variables: {
+          cartId,
+          email: address.email
+        }
+      })
+    ).pipe(
+      map(([updateResult, setEmailResult]) => this.cartTransformer.transform({
+        ...updateResult.data.setBillingAddressOnCart.cart,
+        // only set the email from the set email result since we don't know if the billing address update is included in the response
```
lderrickable: Do we not know this, because we're not sure which mutation will finish first? Like if the setGuestEmail mutation finishes, then the updateBillingAddress returns after, the email might be on the updateBillingAddress response?

libs/cart/src/drivers/magento/queries/set-guest-email.ts
```diff
@@ -0,0 +1,17 @@
...
+import { cartFragment } from './fragments';
+
+export const setGuestEmail = gql`
+  mutation SetGuestEmail($cartId: String!, $email: String!) {
+    setGuestEmailOnCart(
+      cart_id: $cartId,
+      email: $email
+    ) {
+      cart {
+        ...cart
```
lderrickable: As far as I can tell, we won't be using the `setGuestEmail` query alone, because daffodil will set the email when the address is set. Maybe this should just query for the email and not the whole cart? Because in the billing address driver, we're querying for the entire cart twice.

libs/cart/src/drivers/magento/cart-billing-address.service.ts
griest024: Correct except we don't really care if the billing address response has the updated email or not since we are explicitly setting it from the set email response. I added the comment to explain why I wasn't spreading `setEmailResult` in, i.e. we don't know if the rest of `setEmailResult` is the most up to date.

This is based on the assumption that setting the email on the cart only changes the email and nothing else. Updating the billing address is more likely to affect other parts of the cart like `payment` for example.

libs/cart/src/drivers/magento/queries/set-guest-email.ts
griest024: Good point. Will change.

griest024: Changed in [`0c43f4f`](https://github.com/graycoreio/daffodil/pull/671/commits/0c43f4f175486f52edefc55c0806641d76db840a).

libs/cart/src/drivers/magento/cart-billing-address.service.ts
griest024: With the recent change to the set email mutation, this comment is probably not needed.

PR #673, opened 2020-03-05
feat(cart): add Magento GraphQL shipping information driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/src/drivers/magento/cart-shipping-information.service.ts
+++ b/libs/cart/src/drivers/magento/cart-shipping-information.service.ts
@@ -1,0 +1,68 @@
+import { Injectable } from '@angular/core';
+import { Apollo } from 'apollo-angular';
+
+import { Observable } from 'rxjs';
+import { map } from 'rxjs/operators';
+
+import { DaffCartShippingInformationServiceInterface } from '../interfaces/cart-shipping-information-service.interface';
+import { DaffCart } from '../../models/cart';
+import { DaffMagentoCartShippingRateTransformer } from './transforms/outputs/cart-shipping-rate.service';
+import {
+  getSelectedShippingMethod,
+  setSelectedShippingMethod
+} from './queries';
+import { DaffMagentoShippingMethodInputTransformer } from './transforms/inputs/shipping-method.service';
+import { DaffMagentoCartTransformer } from './transforms/outputs/cart.service';
+import { DaffCartShippingRate } from '../../models/cart-shipping-rate';
+import { MagentoGetSelectedShippingMethodResponse, MagentoSetSelectedShippingMethodResponse } from './models/responses';
+
+/**
+ * A service for making Magento GraphQL queries for carts.
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoCartShippingInformationService implements DaffCartShippingInformationServiceInterface<DaffCartShippingRate, DaffCart> {
+  constructor(
+    private apollo: Apollo,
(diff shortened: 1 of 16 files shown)


libs/cart/src/drivers/magento/transforms/inputs/shipping-method.service.ts
```diff
@@ -0,0 +1,14 @@
...
+
+import { MagentoShippingMethodInput } from '../../models/inputs/shipping-method';
+import { DaffCartShippingRate } from '../../../../models/cart-shipping-rate';
+
+@Injectable()
+export class DaffMagentoShippingMethodInputTransformer {
+  transform(method: Partial<DaffCartShippingRate>): MagentoShippingMethodInput {
+    return {
+      carrier_code: method.carrier,
+      method_code: method.method_title
```
griest024: I'm not sure if this is the correct transform for `method_code`.

libs/cart/src/drivers/magento/cart-shipping-information.service.ts
```diff
@@ -0,0 +1,68 @@
...
+      mutation: setSelectedShippingMethod,
+      variables: {
+        cartId,
+        method: {
+          carrier_code: '',
+          method_code: ''
+        }
+      }
+    }).pipe(
+      map(result => this.cartTransformer.transform(result.data.setShippingMethodsOnCart.cart))
```
lderrickable: why do the daffodil CartShippingInformation and SelectPaymentOption driver interfaces differ here? In this delete method, the return is a DaffCart, but for the set payment option driver, the remove method returned void.

griest024: I believe the rationale was that removing the payment didn't need to return the cart because it wouldn't affect other parts of the cart; it was sufficient just to set `payment` to `null`. However removing shipping info could affect totals, discounts, etc so the entire cart must be returned to be safe.

libs/cart/src/drivers/magento/transforms/inputs/shipping-method.service.ts
lderrickable: doesn't this just come from magento, then we give it back to them? Can't we just reverse the magento -> daffodil transform?

libs/cart/src/drivers/magento/cart-shipping-information.service.ts
lderrickable: gotcha. makes sense

libs/cart/src/drivers/magento/transforms/inputs/shipping-method.service.ts
griest024: Yeah and looking back at the [shipping rate transformer](https://github.com/graycoreio/daffodil/blob/develop/libs/cart/src/drivers/magento/transforms/outputs/cart-shipping-rate.service.ts) `method_code` is not preserved anywhere in `DaffCartShippingRate`. Do you think its appropriate to store `method_code` in the `id` field?

griest024: Changed in [`b93ce21`](https://github.com/graycoreio/daffodil/pull/673/commits/b93ce2127ebd54b8dce07e820cd44da3a25343b6).

PR #674, opened 2020-03-06
feat(product): enable get product magento call to get bundled products

lderrickable:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
The get products call for magento can now retrieve bundled products.

```
[ ] Yes
[ ] No
```

## Other information
There is one particular thing I had a lot of trouble with. The fragment for the bundled product query doesn't seem to retrieve any fields in the first layer of the bundled product so:
```
fragment bundledProduct on BundleProduct {

----------------THESE FIELDS ARE NOT RETRIEVED AND RETURN AS NULL-----------------------
        dynamic_price
	dynamic_sku
	dynamic_weight
	price_view
	ship_bundle_items
-------------------------------------------------------------------------------------------------
	items {
		option_id
		position
		required
		sku
		title
		type
		options {
			can_change_quantity
			id
			is_default
			label
			position
			price_type
			price
			quantity
			product {
				id
				name
				sku
			}
		}
	}
}
```
But the rest of the fields come back just fine. I don't know why. Seems to have something to do with fragments, because if I don't use fragments and just paste the BundledProducts query into the GetProductQuery, it works fine.

### DIFF
--- a/libs/product/src/models/bundled-product.ts
+++ b/libs/product/src/models/bundled-product.ts
@@ -1,18 +1,12 @@
 import { DaffProduct } from './product';
 import { DaffBundledProductItem } from './bundled-product-item';
 
-export enum MagentoPriceViewEnum {
+export enum DaffBundledProductPriceViewEnum {
 	price_range = 'PRICE_RANGE',
 	as_low_as = 'AS_LOW_AS'
 }
 
-export enum MagentoBundleItemsEnum {
-	together = 'TOGETHER',
-	separately = 'SEPARATELY'
-}
-
 export interface DaffBundledProduct extends DaffProduct {
-	price_view: MagentoPriceViewEnum;
-	ship_bundle_items: MagentoBundleItemsEnum;
+	price_view: DaffBundledProductPriceViewEnum;
 	items: DaffBundledProductItem[];
 }
--- a/libs/product/src/models/product.ts
+++ b/libs/product/src/models/product.ts
@@ -1,11 +1,16 @@
 import { DaffProductImage } from './product-image';
 
+export enum DaffProductTypeEnum {
+	Simple = 'simple',
+	Composite = 'composite'
+}
+
 /**
  * An interface for a product usable by the @daffodil/product library.
  */
 export interface DaffProduct {
 	id: string;
-	__typename?: string;
+	__typename?: DaffProductTypeEnum;
   price?: string;
   name?: string;
   brand?: string;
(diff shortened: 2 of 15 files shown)


libs/product/src/models/product.ts
```diff
@@ -1,11 +1,17 @@
 import { DaffProductImage } from './product-image';
 
+export enum DaffProductTypeEnum {
+	SimpleProduct = 'Simple Product',
+	BundledProduct = 'Bundled Product',
```
damienwebdev: Let's instead call this a `composite` product. The types here should be shorter:

```ts
Simple = 'simple',
Variant = 'variant',
Composite = 'composite'
```

libs/product/src/models/product.ts
```diff
@@ -1,11 +1,17 @@
...
+	ConfigurableProduct = 'Configurable Product'
+}
+
 /**
  * An interface for a product usable by the @daffodil/product library.
  */
 export interface DaffProduct {
 	id: string;
-	__typename?: string;
+	__typename?: DaffProductTypeEnum;
```
damienwebdev: `type` instead of `__typename`

libs/product/src/models/bundled-product.ts
```diff
@@ -1,18 +1,18 @@
...
 import { DaffBundledProductItem } from './bundled-product-item';
 
-export enum MagentoPriceViewEnum {
+export enum DaffBundledProductPriceViewEnum {
 	price_range = 'PRICE_RANGE',
 	as_low_as = 'AS_LOW_AS'
 }
 
-export enum MagentoBundleItemsEnum {
+export enum DaffBundledItemsEnum {
```
damienwebdev: We don't need this at the moment.

libs/product/src/models/product.ts
damienwebdev: We actually don't support `variant` products yet, so I'd drop that type as well.

lderrickable:
> I think this PR should be split. It's rather large and difficult to understand.
> 
> We need the following PRs for this I think...
> 
> 1. "variable" DaffProduct types (simple and composite).
> 2. The Magento 2 Bundled Product Type
> 3. The driver changes necessary to translate between the types.

I actually did have a few PR before this one. It's just that they weren't correct, and I had to make a lot of changes to get the call to work. But sure, I'll try splitting it

libs/product/src/models/product.ts
lderrickable: I wouldn't say we depend on GraphQl by calling it __typename. We still transform from the magento object to a daffodil one. We don't spread. I just think it'll feel more familiar to people who know graphQL. Why change it for no reason?

PR #675, opened 2020-03-06
bug(category): fix some bugs with the get category query

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
There were a lot of little things I did wrong when setting up the category get request. I tested all these changes in demo and everything is working perfectly now.

```
[ ] Yes
[x] No
```

What was fixed:
1. Some of the query variables for the magento GetProducts call can't be null. They need not to exist.
2. The expected object for the magento category response just wasn't correct.
3. The sort input of the category call for both the daffodil and magento models needed to be defined as enums instead of strings.
4. Misnaming of a magento query argument type.
5. The category id needed to be added to the `filters` argument of the get products call.

### DIFF
--- a/libs/category/src/drivers/magento/category.service.ts
+++ b/libs/category/src/drivers/magento/category.service.ts
@@ -49,7 +49,7 @@ export class DaffMagentoCategoryService implements DaffCategoryServiceInterface 
 	}
 	
 	private getProductsQueryVariables(request: DaffCategoryRequest) {
-		let queryVariables = {
+		const queryVariables = {
 			filter: this.magentoAppliedFiltersTransformer.transform(request.id, request.applied_filters)
 		};
 		if(request.page_size) queryVariables['pageSize'] = request.page_size;
--- a/libs/category/src/drivers/magento/category.service.ts
+++ b/libs/category/src/drivers/magento/category.service.ts
@@ -14,7 +14,6 @@ import { MagentoGetProductsQuery } from './queries/get-products';
 import { DaffMagentoAppliedFiltersTransformService } from './transformers/applied-filter-transformer.service';
 import { DaffMagentoAppliedSortOptionTransformService } from './transformers/applied-sort-option-transformer.service';
 import { DaffMagentoCategoryResponseTransformService } from './transformers/category-response-transform.service';
-import { CompleteCategoryResponse } from './models/outputs/complete-category-response';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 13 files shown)


libs/category/src/drivers/magento/transformers/applied-sort-option-transformer.service.ts
```diff
@@ -1,15 +1,25 @@
...
+  transform(appliedOption: string, appliedDirection: DaffSortDirectionEnum): MagentoSortFieldAction {
+		if(!appliedOption) return null;
 		return {
-			[appliedOption]: appliedDirection
+			[appliedOption]: this.transformDirection(appliedDirection)
 		}
-  }
+	}
+	
+	private transformDirection(direction: DaffSortDirectionEnum): MagentoSortDirectionEnum {
```
damienwebdev: What if direction is null?

libs/category/src/drivers/magento/category.service.ts
```diff
@@ -49,23 +50,32 @@ export class DaffMagentoCategoryService implements DaffCategoryServiceInterface
...
-			filter: this.magentoAppliedFiltersTransformer.transform(request.applied_filters),
+		const queryVariables = {
+			filter: this.magentoAppliedFiltersTransformer.transform(request.id, request.applied_filters),
 			search: null,
 			pageSize: request.page_size,
 			currentPage: request.current_page,
 			sort: this.magentoAppliedSortTransformer.transform(request.applied_sort_option, request.applied_sort_direction)
 		}
+
+		if(!queryVariables.pageSize) delete queryVariables.pageSize;
```
damienwebdev: Instead of deleting after-the-fact, can we build this along the way for the nullable props?

libs/category/src/drivers/magento/transformers/applied-sort-option-transformer.service.ts
lderrickable: fixed. But moved this code to the category service

libs/category/src/drivers/magento/category.service.ts
lderrickable: sure. Done

libs/category/src/drivers/magento/category.service.ts
```diff
@@ -49,23 +49,29 @@ export class DaffMagentoCategoryService implements DaffCategoryServiceInterface
 	}
 	
 	private getProductsQueryVariables(request: DaffCategoryRequest) {
```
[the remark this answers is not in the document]
lderrickable: done

PR #676, opened 2020-03-06
feat(cart): add Magento GraphQL cart item driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/src/drivers/magento/cart-item.service.ts
+++ b/libs/cart/src/drivers/magento/cart-item.service.ts
@@ -1,0 +1,112 @@
+import { Injectable, Inject } from '@angular/core';
+import { Apollo } from 'apollo-angular';
+
+import { Observable } from 'rxjs';
+import { map, switchMap } from 'rxjs/operators';
+
+import {
+  DaffProductServiceInterface,
+  DaffProductUnion,
+  DaffProductDriver
+} from '@daffodil/product'
+
+import { DaffCartItemServiceInterface } from '../interfaces/cart-item-service.interface';
+import { DaffCart } from '../../models/cart';
+import { DaffCartItem } from '../../models/cart-item';
+import { DaffCartItemInput } from '../../models/cart-item-input';
+import { DaffMagentoCartTransformer } from './transforms/outputs/cart.service';
+import {
+  listCartItems,
+  addCartItem,
+  removeCartItem,
+  updateCartItem
+} from './queries';
+import { DaffMagentoCartItemTransformer } from './transforms/outputs/cart-item.service';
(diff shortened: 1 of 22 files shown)


libs/cart/src/drivers/magento/models/inputs/cart-item-with-sku.ts
```diff
@@ -0,0 +1,6 @@
+import { DaffCartItemInput } from '../../../../models/cart-item-input';
+import { DaffCartItem } from '../../../../models/cart-item';
+
+export interface MagentoDaffCartItemInputWithSku extends DaffCartItemInput {
```
lderrickable: This should be able to be removed with the other changes mentioned

libs/cart/src/drivers/magento/cart-item.service.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  get(cartId: string, itemId: number): Observable<DaffCartItem> {
+    return this.list(cartId).pipe(
+      map(items => items.find(item => Number(item.item_id) === itemId))
+    )
+  }
+
+  add(cartId: string, product: DaffCartItemInput): Observable<Partial<DaffCart>> {
+    return this.product.get(product.productId).pipe(
```
lderrickable: I think I understand why you did this. Because you needed the sku of the product? But with magento, we treat the product.id as the same as the sku (because you can't actually filter for products on the magentoProduct.id. You need to use the magentoProduct.sku. Apparently you can't add a product to the cart by magentoProduct.id either). So you can just take in the product id from the user, and just transform it back into the sku. Then you can get rid of that MagentoDaffCartInputWithSku object too.

griest024: Ah well that will certainly be simpler. Changing now.

griest024: Changed in [`931e356`](https://github.com/graycoreio/daffodil/pull/676/commits/931e356340d5188dea7fc0865152377868fef06b).

libs/cart/src/drivers/magento/models/inputs/cart-item-with-sku.ts
griest024: Changed in [`931e356`](https://github.com/graycoreio/daffodil/pull/676/commits/931e356340d5188dea7fc0865152377868fef06b).

PR #679, opened 2020-03-06
feat(cart): add cart actions

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The cart actions don't correspond to the current cart service interface.
Fixes: N/A

## What is the new behavior?
- Updates the actions to the new interface
- Adds generics on the actions where needed for the payload

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/src/actions/cart.actions.ts
+++ b/libs/cart/src/actions/cart.actions.ts
@@ -3,27 +3,28 @@ import { Action } from '@ngrx/store';
 import { DaffCart } from '../models/cart';
 
 export enum DaffCartActionTypes {
-    CartLoadAction = '[DaffCart] Load Action',
-    CartLoadSuccessAction = '[DaffCart] Load Success Action',
-    CartLoadFailureAction = '[DaffCart] Load Failure Action',
-    AddToCartAction = '[DaffCart] Add To Cart Action',
-    AddToCartSuccessAction = '[DaffCart] Add to Cart Success Action',
-    AddToCartFailureAction = '[DaffCart] Add to Cart Failure Action',    
-    CartResetAction = '[DaffCart] Reset Action',
-    CartResetSuccessAction = '[DaffCart] Reset Success Action',
-    CartResetFailureAction = '[DaffCart] Reset Failure Action'
+  CartLoadAction = '[DaffCart] Cart Load Action',
+  CartLoadSuccessAction = '[DaffCart] Cart Load Success Action',
+  CartLoadFailureAction = '[DaffCart] Cart Load Failure Action',
+  CartCreateAction = '[DaffCart] Cart Create Action',
+  CartCreateSuccessAction = '[DaffCart] Cart Create Success Action',
+  CartCreateFailureAction = '[DaffCart] Cart Create Failure Action',
+  AddToCartAction = '[DaffCart] Cart Add To Cart Action',
+  AddToCartSuccessAction = '[DaffCart] Cart Add to Cart Success Action',
+  AddToCartFailureAction = '[DaffCart] Cart Add to Cart Failure Action',
+  CartResetAction = '[DaffCart] Cart Reset Action',
+  CartResetSuccessAction = '[DaffCart] Cart Reset Success Action',
+  CartResetFailureAction = '[DaffCart] Cart Reset Failure Action'
 }
 
 export class DaffCartLoad implements Action {
   readonly type = DaffCartActionTypes.CartLoadAction;
-
-  constructor() {}
 }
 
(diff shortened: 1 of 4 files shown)


libs/cart/src/actions/cart.actions.ts
```diff
@@ -3,27 +3,28 @@ import { Action } from '@ngrx/store';
...
+  CartLoadAction = '[DaffCart] Cart Load Action',
+  CartLoadSuccessAction = '[DaffCart] Cart Load Success Action',
+  CartLoadFailureAction = '[DaffCart] Cart Load Failure Action',
+  CartCreateAction = '[DaffCart] Cart Create Action',
+  CartCreateSuccessAction = '[DaffCart] Cart Create Success Action',
+  CartCreateFailureAction = '[DaffCart] Cart Create Failure Action',
+  AddToCartAction = '[DaffCart] Cart Add To Cart Action',
+  AddToCartSuccessAction = '[DaffCart] Cart Add to Cart Success Action',
+  AddToCartFailureAction = '[DaffCart] Cart Add to Cart Failure Action',
+  CartResetAction = '[DaffCart] Cart Reset Action',
```
lderrickable: Will this action trigger a clearCart API call? Either way, I think you're going to need a clear cart action

libs/cart/src/actions/cart.actions.ts
```diff
@@ -3,27 +3,28 @@ import { Action } from '@ngrx/store';
...
+  CartResetSuccessAction = '[DaffCart] Cart Reset Success Action',
+  CartResetFailureAction = '[DaffCart] Cart Reset Failure Action'
 }
 
 export class DaffCartLoad implements Action {
```
lderrickable: So the idea here is to get the id from state/localstorage? Then make a createCart call if there is no id?

libs/cart/src/actions/cart.actions.ts
```diff
@@ -32,46 +33,66 @@ export class DaffCartLoadFailure implements Action {
...
-    constructor(public payload: {productId: string, qty: number}) {}
+  constructor(public payload: {productId: string, qty: number}) {}
 }
 
 export class DaffAddToCartSuccess implements Action {
-    readonly type = DaffCartActionTypes.AddToCartSuccessAction;
+  readonly type = DaffCartActionTypes.AddToCartSuccessAction;
 
-    constructor(public payload: DaffCart) {}
+  constructor(public payload: DaffCart) {}
```
lderrickable: Should be generic

libs/cart/src/actions/cart.actions.ts
griest024: Yeah I repurposed the reset action for the clear call but didn't rename the actions.

libs/cart/src/actions/cart.actions.ts
griest024: Yeah that was my idea.

libs/cart/src/actions/cart.actions.ts
```diff
@@ -32,46 +33,66 @@ export class DaffCartLoadFailure implements Action {
...
+
+  constructor(public payload: string) {}
+}
+
 export class DaffAddToCart implements Action {
-    readonly type = DaffCartActionTypes.AddToCartAction;
+  readonly type = DaffCartActionTypes.AddToCartAction;
 
-    constructor(public payload: {productId: string, qty: number}) {}
+  constructor(public payload: {productId: string, qty: number}) {}
```
lderrickable: seems inconsistent that the `DaffCartCreateSuccess` action takes a payload of `{id: T['id']}` while this takes a payload of `{productId: string, qty: number}`. Seems like it should be something like `{productId: T['id'], qty: T['quantity']}`

griest024: You're right, it is inconsistent. I didn't update anything related to the deprecated `addToCart` call.

lderrickable: oh, right. I forgot it was deprecated. Nevermind then.

libs/cart/src/actions/cart.actions.ts
lderrickable: looks good then

libs/cart/src/actions/cart.actions.ts
lderrickable: Do you think it would make your life much more difficult to rename them to "clear"?

griest024: Nope, will do.

PR #682, opened 2020-03-07
feat(product, cart, category): clean up magento models/factories/queries

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

A lot of the magento product models (and consequently the queries and factories) aren't strongly typed all the time, or just aren't in sync. There's also a product query fragment in cart.

## What is the new behavior?
Moved the product fragment into the product package. Fixed up and neatened up the product models for magento.

```
[x] Yes
[ ] No
```

I addressed the breaking change in this PR.


libs/cart/src/drivers/magento/queries/fragments/cart-item.ts
```diff
@@ -26,6 +27,6 @@ export const cartItemFragment = gql`
       }
     }
   }
-  ${productFragment}
+  ${magentoProductFragment}
```
damienwebdev: I don't agree with this change. We shouldn't be retrieving the full product on the cart. It's wasteful of bandwidth.

libs/product/src/drivers/magento/queries/fragments/product.ts
```diff
@@ -0,0 +1,35 @@
+import gql from 'graphql-tag';
+
+export const magentoProductFragment = gql`
+  fragment product on MagnetoProduct {
```
damienwebdev: Magneto?

libs/product/src/drivers/magento/transforms/product-transformer.service.ts
```diff
@@ -27,7 +27,7 @@ export class DaffMagentoProductTransformerService {
           return {
             url: response.storeConfig.secure_base_media_url + 'catalog/product' + image.file,
             label: image.label,
-            id: image.id
+            id: image.id.toString()
```
damienwebdev: Why does this need to be typecast?

lderrickable: Because the id from magento is a number, and the daffodil one is a string. The magento object that was typed as `any` is now typed as 
```
{
   label: string,
   file: string,
   position: number,
   disabled: boolean,
   id: number
}
```

libs/cart/src/drivers/magento/queries/fragments/cart-item.ts
lderrickable: It isn't actually any different. The other fragment query in cart was just wrong. We could do another PR to change what we get back if you want. Here is the old one:
```
export const productFragment = gql
  fragment product on MagnetoProduct {
    id
    image {
      label,
      url
    }
    name
    description      <------------this will cause a graphQL failed call. needs to be "description { html }"
    sku
    url_key
    price {
      regularPrice                             <-----------------------This will cause a failure
    }
    media_gallery_entries                             <---------------this will cause a failure
    short_description                                    <---------------this will cause a failure
  }
;
```
So yea there's really no difference. We can trim the request down later?

libs/product/src/drivers/magento/queries/fragments/product.ts
lderrickable: fixed

libs/cart/src/drivers/magento/queries/fragments/cart-item.ts
damienwebdev: That's fine.

libs/product/src/drivers/magento/models/product-node.ts
```diff
@@ -9,9 +6,29 @@ export interface ProductNode {
...
+		position: number,
+		disabled: boolean,
+		id: number
+	}[];
+  short_description?: {
+		html: string;
+	};
+  description?: {
+		html: string;
+	};
```
griest024: I think eventually we will want these extracted to their own "core" models but that can be done later.

libs/product/src/drivers/magento/queries/fragments/product.ts
```diff
@@ -0,0 +1,35 @@
+import gql from 'graphql-tag';
+
+export const magentoProductFragment = gql`
+  fragment product on ProductInterface {
```
griest024: I think this maybe should be a concrete type that the fragment operates on. Not totally sure though and we can come back to perfecting the fragments later. I doubt it will actually cause any problems.

lderrickable: This needs to be "ProductInterface" or the magento call will fail

libs/product/src/drivers/magento/models/product-node.ts
lderrickable: I thought about doing that, but it seems like it's difficult to keep track of what all the return values should be unless they're all in the same place. It's a lot easier to go back and forth between magento documentation and this, than between magento documentation and a bunch of objects. That was my thinking anyway

griest024: At least for models such as image or money, I think many different models will end up using it and that would make it worth the additional complexity of having multiple files. We definitely do have to consider exactly what fields we need from the response so all of these might not get extracted. Either way, its something that can be better addressed later.

libs/product/src/drivers/magento/queries/fragments/product.ts
griest024: hm maybe I'm wrong then. In that case the cart fragments might need changing.

griest024:
This PR lgtm.

PR #685, opened 2020-03-09
feat(category): add action for toggling a category filter

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
An action was added for toggling a filter. Simplifies a front end application by extracting management of the applied filters to daffodil

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/category/src/effects/category.effects.ts
+++ b/libs/category/src/effects/category.effects.ts
@@ -166,11 +166,11 @@ export class DaffCategoryEffects {
 		toggledFilter: DaffCategoryFilterAction, 
 		appliedFilters: DaffCategoryFilterAction[]
 	): DaffCategoryFilterAction[] {
-		return appliedFilters.indexOf(toggledFilter) 
-		// filter is applied, remove it
-		? appliedFilters.filter(filter => filter !== toggledFilter)
+		return appliedFilters.indexOf(toggledFilter) < 0
 		// filter is not applied, add it
-		: appliedFilters.concat(toggledFilter);
+		? appliedFilters.concat(toggledFilter)
+		// filter is applied, remove it
+		: appliedFilters.filter(filter => filter !== toggledFilter);
 	}
 
   private processCategoryGetRequest(payload: DaffCategoryRequest) {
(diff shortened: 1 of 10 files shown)


libs/category/src/effects/category.effects.ts
```diff
@@ -135,7 +160,20 @@ export class DaffCategoryEffects {
...
+		appliedFilters: DaffCategoryFilterAction[]
+	): DaffCategoryFilterAction[] {
+		if(!appliedFilters) return [toggledFilter];
+
+		const nonMatchingFilters = appliedFilters.filter(filter => filter !== toggledFilter);
+
+		if(nonMatchingFilters.length === appliedFilters.length) nonMatchingFilters.push(toggledFilter);
+
+		return nonMatchingFilters.length ? nonMatchingFilters : null;
+	}
```
griest024: This method is hard to understand. At the very least it needs comments explaining what is going on and what the purpose of the method is but I would rewrite it to make the code more clear. As I understand it, this method will either add or remove the toggled filter from the applied filters based on whether or not it is already applied. Assuming my understanding is correct, a clearer implementation would be:
```ts
private toggleCategoryFilter(
  toggledFilter: DaffCategoryFilterAction, 
  appliedFilters: DaffCategoryFilterAction[] = []
): DaffCategoryFilterAction[] {
  return appliedFilters.includes(toggledFilter)
    // filter is applied, remove it
    ? appliedFilters.filter(filter => filter !== toggledFilter)
    // filter is not applied, add it
    : appliedFilters.concat([toggledFilter])
}
```

Unless there is a reason you need to return `null`, I think this is better.

griest024:
Needs build fixed. I'm not sure changing all the state to arrays was needed; its possible that a default argument would have been sufficient. However, I do think its better to have state that is an array to be initialized to an empty array so you don't have to defensively program against it.

lderrickable:
> Needs build fixed. I'm not sure changing all the state to arrays was needed; its possible that a default argument would have been sufficient. However, I do think its better to have state that is an array to be initialized to an empty array so you don't have to defensively program against it.

Yea, sorry. I thought it was going to be necessary, because the Magento category call blows up if you pass it anything but a valid filter. So I thought it would end up being necessary to change a lot of stuff, because that method returns an empty array instead of null. But yea, I think it ended up being unnecessary.

griest024:
Its all good though, I think having empty arrays is better than `null`.
> the Magento category call blows up if you pass it anything but a valid filter
seems about right 🙄

PR #691, opened 2020-03-11
fix(cart): importing magento driver module breaks app builds

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Importing `DaffCartMagentoDriverModule` and using it in an app will break the build with error:
```
Error during template compile of 'MagentoModule'
  Function calls are not supported in decorators but 'DaffCartMagentoDriverModule' was called.
```

## What is the new behavior?
The app using the module will correctly build.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/apps/demo/src/app/app.module.ts
+++ b/apps/demo/src/app/app.module.ts
@@ -27,6 +27,7 @@ import { ThankYouModule } from './thank-you/thank-you.module';
 import { DemoRoutingComponentModule } from './routing/routing-component.module';
 import { InMemoryModule } from './in-memory.module';
 import { DemoCartRootModule } from './cart/cart-root.module';
+import { MagentoModule } from './magento.module';
 
 @NgModule({
   declarations: [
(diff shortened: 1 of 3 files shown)


damienwebdev:
@griest024 In theory, this should be breaking the demo app. Can you determine why it isn't? I'd like to be able to verify this behavior and prevent future regressions.

griest024:
@damienwebdev I don't think that demo is even importing the Magento module.

damienwebdev:
@griest024 can we import this into demo (even if we don't use the driver) and validate that this bug is in-fact fixed?

apps/demo/src/app/app.module.ts
```diff
@@ -38,16 +39,17 @@ import { DemoCartRootModule } from './cart/cart-root.module';
 
     HttpClientModule,
 
-    InMemoryModule,
-    
+    // InMemoryModule,
```
damienwebdev: I don't think we should make any changes in this file.

griest024: I thought it was necessary to break the demo but I was wrong. Changed in [`284a5d1`](https://github.com/graycoreio/daffodil/pull/691/commits/284a5d18ce9adaf28c04a15a92ff7d9cd9556ee9).

PR #696, opened 2020-03-12
feat(cart): handle empty and null responses in Magento driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

- Magento responses that contain null or empty fields will cause the transformers to throw when passed
- Queries use incorrect parameters

## What is the new behavior?
- The drivers and cart transformer will check for `null` and `[]` responses and return `null` without invoking the transformer
- Queries now use the correct parameters

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/src/drivers/magento/cart-billing-address.service.spec.ts
+++ b/libs/cart/src/drivers/magento/cart-billing-address.service.spec.ts
@@ -145,6 +145,25 @@ describe('Driver | Magento | Cart | CartBillingAddressService', () => {
       });
     });
 
+    describe('when the response is null', () => {
+      beforeEach(() => {
+        mockGetBillingAddressResponse.cart.billing_address = null;
+      });
+
+      it('should return null and not throw', done => {
+        service.get(cartId).subscribe(result => {
+          expect(result).toBeNull();
+          done();
+        });
+
+        const op = controller.expectOne(getBillingAddress);
+
+        op.flush({
+          data: mockGetBillingAddressResponse
+        });
+      });
+    });
+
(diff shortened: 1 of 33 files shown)


lderrickable:
I feel like the most complicated method is typically the driver call, so I think you should move the null checking part into the transformers when possible.

griest024:
@lderrickable I was thinking that its not really the responsibility of the transformer to determine if the response needs to be transformed or not. Plus not all checks can be done inside the transformer and I wanted to keep it consistent where the check were done.

Moving the check inside the transformer would prevent some duplication of logic though; I'm open to that option. LMK what you think.

lderrickable:
@griest024 I could also argue that transformers should be able to handle transforming any variation of the type of object they expect. I think null/undefined should always be considered by any service/function. I think this is why I don't like how magneto accepts arguments for graphQL calls. Passing null values makes a magento graphQL query fail. This is a little different because we control the inputs by the driver service into a transformer, but I still think as a rule a method should be responsible for handling a null/undefined input.

griest024:
@lderrickable You've convinced me. That makes sense especially because TS allows `null` to be passed into the transformers.

griest024:
@lderrickable Changed in [`24db473`](https://github.com/graycoreio/daffodil/pull/696/commits/24db473c7cba041db4725c9c3044fb4390025e8d). I added a check in all of the response transformers. Some drivers still needed a check.

PR #698, opened 2020-03-12
feat(category): add product_ids and total_products to the category state

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Categories need their own product Ids in order to have multiple categories loaded on one page.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/category/src/selectors/category.selector.ts
+++ b/libs/category/src/selectors/category.selector.ts
@@ -165,7 +165,7 @@ export const selectProductsByCategory = createSelector(
 export const selectTotalProductsByCategory = createSelector(
 	selectCategoryEntities,
 	fromProduct.selectAllProducts,
-	(entities, products, props) => entities[props.id] && entities[props.id].product_ids
-		? products.filter(product => entities[props.id].product_ids.indexOf(product.id) >= 0).length
+	(entities, products, props) => selectProductsByCategory.projector(entities, products, { id: props.id })
+		? selectProductsByCategory.projector(entities, products, { id: props.id }).length
 		: null
 );
--- a/libs/category/src/selectors/category.selector.ts
+++ b/libs/category/src/selectors/category.selector.ts
@@ -151,18 +151,21 @@ export const selectCategoryPageProducts = createSelector(
 
 export const selectCategory = createSelector(
 	selectCategoryEntities,
-	(entities, props) => {
-		return entities[props.id];
-	}
+	(entities, props) =>  entities[props.id]
 );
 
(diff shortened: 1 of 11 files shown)


libs/category/src/selectors/category.selector.ts
```diff
@@ -151,18 +151,21 @@ export const selectCategoryPageProducts = createSelector(
...
+		? products.filter(product => entities[props.id].product_ids.indexOf(product.id) >= 0)
+		: null
+);
+
+export const selectTotalProductsByCategory = createSelector(
+	selectCategoryEntities,
+	fromProduct.selectAllProducts,
+	(entities, products, props) => entities[props.id] && entities[props.id].product_ids
+		? products.filter(product => entities[props.id].product_ids.indexOf(product.id) >= 0).length
+		: null
```
griest024: So this selector is just getting a count of the products? Is there any way that it can use `selectProductsByCategory` to simplify the logic?

lderrickable: oh lol, yea

lderrickable: It's a little simpler. Probably more useful for more complex selectors, but definitely the right call.

griest024: hm yeah not as simple as I envisioned it but that seems to be the best way. One last thing, can you only make one call to the other selector, i.e.
```ts
export const selectTotalProductsByCategory = createSelector(
  selectCategoryEntities,
  fromProduct.selectAllProducts,
  (entities, products, props) => {
    const products = selectProductsByCategory.projector(entities, products, { id: props.id });
    return products ? products.length : null
  }
)
```

lderrickable: I think it's better this way, because it's just adding an extra subscription to a selector, whereas the other way adds a copy of all products in the app to memory.

lderrickable: actually..hmm. nevermind, it'll just save an observable of the products in state. I'll make the change

lderrickable: nevermind again. I think it must save the actual array of products (for the category, not all of the products (I misspoke earlier)), because you can check the length of it. So it must be the actual value and not an observable.

griest024: hm good point. okay I'm good with this then.

PR #707, opened 2020-03-17
bug(cart): fix bug where clearing an empty cart failed

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Attempting to clear an empty cart results in an error.

Fixes: N/A

## What is the new behavior?
Fixed.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/cart/src/drivers/magento/cart.service.ts
+++ b/libs/cart/src/drivers/magento/cart.service.ts
@@ -61,7 +61,7 @@ export class DaffMagentoCartService implements DaffCartServiceInterface<DaffCart
       ),
       // find the most up to date delete response by looking for one with no items
       // make a get request if there were no items in the cart
-			switchMap(updatedCarts => !!updatedCarts ? 
+			switchMap(updatedCarts => updatedCarts ? 
 				of(updatedCarts.filter(cart => cart.items.length === 0).shift()) : 
 				this.get(cartId))
     )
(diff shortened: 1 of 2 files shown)


libs/cart/src/drivers/magento/cart.service.ts
```diff
@@ -53,14 +53,17 @@ export class DaffMagentoCartService implements DaffCartServiceInterface<DaffCart
...
+        items.length ? zip(...items.map(item =>
           this.cartItemDriver.delete(cartId, item.item_id)
-        ))
+        )) : of(null)
       ),
       // find the most up to date delete response by looking for one with no items
-      // it might be necessary to make a get request instead if delete does not return needed fields
-      map(updatedCarts => updatedCarts.find(cart => cart.items.length === 0))
+      // make a get request if there were no items in the cart
+			switchMap(updatedCarts => !!updatedCarts ?
```
griest024: I think `!!` is unnecessary if you don't actually need an explicit boolean. For the purposes of a conditional it should be sufficient to do `updatedCarts ?`.

lderrickable:
> What do you think about doing this logic in the effects layer instead? It would save an additional `get` request. Also, if this is behavior that we want across daffodil as a whole, it would be nice not to have to ensure consistent behavior across multiple drivers.
> 
> LMK your thoughts on this.

No, it should be here, I think, because other platforms might have a clear cart mutation rather than just a remove item which would make it unnecessary. We should treat this as a magento-only problem, I think

libs/cart/src/drivers/magento/cart.service.ts
lderrickable: changed

griest024:
@lderrickable good point

PR #709, opened 2020-03-17
fix(cart): fix cart paths to use public_api

damienwebdev:
When doing AOT compilation, things like barrel exports and paths like 'path/to/dir' where you're actually referring 'path/to/dir/index' can make some wild errors that really only show up at runtime AFTER an AOT build (e.g. in production). We really want to avoid these kinds of bugs, as they're catastrophic failures.

As a result, we should only use paths like public_api NOT index for exports.

See: https://github.com/angular/angular/issues/20931


damienwebdev:
@griest024 good point. Added that as well.

PR #716, opened 2020-03-20
feat(geography): outline basic driver model and types

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information


libs/geography/src/drivers/interfaces/geography-service.interface.ts
```diff
@@ -0,0 +1,13 @@
...
+	/**
+	 * Retrieves the list of countries available to the given store.
+	 */
+	list(): DaffCountry[];
+
+	/**
+	 * Retrieve precise information about a specific country.
+	 */
+	get(id: DaffCountry['id']): DaffCountry;
+}
```
griest024: We need an injection token as well.

libs/geography/src/models/subdivision.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * The subdivisions of a country as per ISO-3166-2
+ * For the United States this is commonly termed as "States"
+ */
+export interface DaffSubdivision {
+  id: string;
+  name: string;
+  name_en: string;
```
griest024: Do you think that `name_en` field should be nullable? Magento for example does not provide a separate english name field for [Region](https://devdocs.magento.com/guides/v2.3/graphql/queries/directory-countries.html#region). LMK what you think.

libs/geography/src/drivers/interfaces/geography-service.interface.ts
damienwebdev: fixed

libs/geography/src/models/subdivision.ts
damienwebdev: fixed

PR #718, opened 2020-03-20
feat(cart): add cart item in-memory driver and backend

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/testing/src/drivers/in-memory/cart-item/cart-item.service.spec.ts
+++ b/libs/cart/testing/src/drivers/in-memory/cart-item/cart-item.service.spec.ts
@@ -51,9 +51,8 @@ describe('Driver | In Memory | Cart | CartItemService', () => {
   });
 
   describe('list | getting all the cart items', () => {
-    it('should send a get request and return the cart item', done => {
+    it('should send a get request', done => {
       cartItemService.list(cartId).subscribe(res => {
-        expect(res).toEqual(mockCart.items);
         done();
       });
 
@@ -65,9 +64,8 @@ describe('Driver | In Memory | Cart | CartItemService', () => {
   });
 
   describe('get | getting a cart item', () => {
-    it('should send a get request and return the cart item', done => {
+    it('should send a get request with the item id', done => {
       cartItemService.get(cartId, itemId).subscribe(res => {
-        expect(res).toEqual(jasmine.objectContaining(mockCartItem));
         done();
       });
 
@@ -80,7 +78,6 @@ describe('Driver | In Memory | Cart | CartItemService', () => {
 
   describe('add | adding an item to the cart', () => {
     let cartItemInput: DaffCartItemInput;
-    let expected: DaffCartItem;
     let productId;
     let qty;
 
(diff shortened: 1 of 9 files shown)


libs/cart/testing/src/drivers/in-memory/cart-item/cart-item.service.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
+
+import { DaffCart, DaffCartItem } from '@daffodil/cart';
+
+import { DaffCartFactory } from '../../../factories/cart.factory';
+import { DaffInMemoryCartItemService } from './cart-item.service';
+import { DaffCartItemFactory } from '@daffodil/cart/testing';
+import { DaffCartItemInput } from 'dist/cart/models/cart-item-input';
+
+describe('Driver | In Memory | Cart | CartItemService', () => {
```
lderrickable: I think these tests are a little misleading, and I'm guessing we probably do it in the other in memory drivers. But really, the only thing this should test is that calling a method starts a request. The part about actually returning what is supposed to be returned isn't actually done here. So I think the tests should just be like:
```
describe('list', () => {
    it('should send a get request for a list of cart items', done => {
      cartItemService.list(cartId).subscribe(res => {
        done();
      });

      const req = httpMock.expectOne(`${cartItemService.url}/${cartId}/`);

      expect(req.request.method).toBe('GET');
      req.flush();                           <------I'm not sure whether you'll need to flush with actual data or not.
    });
  });
```

libs/cart/testing/src/in-memory-backend/cart-items/cart-items.service.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    beforeEach(() => {
+      reqInfoStub.url = cartUrl;
+      reqInfoStub.req.body = mockCartItemInput;
+      productId = mockCartItemInput.productId;
+
+      result = service.post(reqInfoStub);
+    });
+
+    it('should return a cart item with the product id', () => {
```
lderrickable: should be "return a cart with the added item"

libs/cart/testing/src/drivers/in-memory/cart-item/cart-item.service.spec.ts
griest024: Changed

libs/cart/testing/src/in-memory-backend/cart-items/cart-items.service.spec.ts
griest024: Changed

lderrickable:
I think it might make more sense to only have the root-cart service extend the InMemoryService, because only it needs to createDb and do the InMemory stuff. The cart and cart-item inmemory service really only need to be a set of pure functions that mutate the cart object in the root-cart service. What are your thoughts on this?

But I think we should push this PR as it stands, finish the other mock http calls for the other cart drivers, then we'll have a better picture for how this is going to work by testing it in demo. And we can make the other changes then.

griest024:
@lderrickable hm thats a good point, I guess the child backends don't use any of the `InMemoryDbService` features. Additionally, they can't be used on their own so its actually probably a bad idea to have them implement `InMemoryDbService`.

We can push this in its current state. I'll open PRs for the drivers.

Issue #719, opened 2020-03-20
[BUG] Daff Cart Item discounted_row_total is not defined.


xelaint:
# :bug: Bug report

## Current Behavior
`discounted_row_total` is not defined for cart items that are discounted. The only cart item total calculation that exist is `row_total`.

## Expected Behavior

`discounted_row_total` needs to be defined in order to calculate a cart item that is on discount.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

PR #720, opened 2020-03-20
feat(cart): add total_discount to DaffCartItem

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #719 

## What is the new behavior?
- Adds `total_discount` to `DaffCartItem`
- Transforms `total_discount` in the Magento cart item transformer

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/src/drivers/magento/transforms/outputs/cart-item.service.ts
+++ b/libs/cart/src/drivers/magento/transforms/outputs/cart-item.service.ts
@@ -29,9 +29,10 @@ export class DaffMagentoCartItemTransformer {
       product_id: String(cartItem.product.id),
 			image: {
 				id: cartItem.product.image.label,
-				url: cartItem.product.image.url, 
+				url: cartItem.product.image.url,
 				label: cartItem.product.image.label
-			},
+      },
+      discounted_row_total: cartItem.prices.row_total.value - cartItem.prices.total_item_discount.value,
 
       // TODO: implement
       parent_item_id: 0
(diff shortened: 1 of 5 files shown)


libs/cart/src/drivers/magento/transforms/outputs/cart-item.service.ts
```diff
@@ -29,9 +29,10 @@ export class DaffMagentoCartItemTransformer {
       product_id: String(cartItem.product.id),
 			image: {
 				id: cartItem.product.image.label,
-				url: cartItem.product.image.url, 
+				url: cartItem.product.image.url,
 				label: cartItem.product.image.label
-			},
+      },
+      discounted_row_total: cartItem.prices.row_total.value - cartItem.prices.total_item_discount.value,
```
lderrickable: I'm pretty sure the total_item_discount is the total of discounts applied to an item without consideration of the quantity. The row_total does consider quantity.  We should probably just put a total_discounts on the DaffCartItem and return the discounted row total in a selector. That way, as we add more platforms, we don't need to keep doing this computation over and over (presuming other platforms also don't have a ready-made discounted_row_total field).

griest024: Doing that computation in a selector seems better. Changed.

libs/cart/testing/src/factories/cart-item.factory.ts
```diff
@@ -14,14 +14,15 @@ export class MockCartItem implements DaffCartItem {
   qty = faker.random.number({min:1, max:100});
   price = faker.random.number(1500);
   row_total = this.qty * this.price
+  total_discount = faker.random.number(1000);
```
lderrickable: I think we should guarantee that this will be less than the price field

griest024: Changed

PR #725, opened 2020-03-22
fix(category): allow categories to be sorted based on platform response

damienwebdev:
When a category loads its products, they're in a certain order based on the platforms defined order. Previously we were filtering through an array, picking products as we found them in the array. This means that the order was determined client side instead of server-side.

I've changed this to map to the element in the entities dictionary to properly select the products in the appropriate order. I've also tacked on a filter to prevent undefined elements from being in the final array. This can be deleted once we have dependency injectable reducers.


lderrickable:
Please separate the formatting changes and the actual change into separate commits. I can't tell what the actual change is.

PR #729, opened 2020-03-23
feat(geography): add factories

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
Adds geography models factories

```
[ ] Yes
[ ] No
```

## Other information


libs/geography/testing/src/factories/subdivision.factory.ts
```diff
@@ -0,0 +1,21 @@
+import { Injectable } from '@angular/core';
+import * as faker from 'faker/locale/en_US';
+
+import { DaffSubdivision } from '@daffodil/geography';
+
+import { DaffModelFactory } from '@daffodil/core/testing';
+
+export class MockSubdivision implements DaffSubdivision {
+  id = faker.random.number(1000);
```
damienwebdev: `id` is a string on the type. Either cast this or expand the type. It may be preferable to expand the type, as we can't say what it will explicitly be for every platform.

libs/geography/testing/src/factories/country.factory.ts
```diff
@@ -0,0 +1,24 @@
+import { Injectable } from '@angular/core';
+import * as faker from 'faker/locale/en_US';
+
+import { DaffCountry } from '@daffodil/geography';
+
+import { DaffModelFactory } from '@daffodil/core/testing';
+
+export class MockCountry implements DaffCountry {
+  id = faker.random.number(1000);
```
damienwebdev: id is a string on the type. Either cast this or expand the type. It may be preferable to expand the type, as we can't say what it will explicitly be for every platform.

libs/geography/testing/src/factories/subdivision.factory.ts
griest024: Changed

damienwebdev: Works for me.

libs/geography/testing/src/factories/country.factory.ts
griest024: Changed

PR #737, opened 2020-03-25
fix(cart): throw error when in memory cart does not exist

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Every time the app is refreshed while using the in memory cart driver, the cart Id in local storage no longer matches the one in the in memory api and it fails all cart calls silently.

## What is the new behavior?
A daffodil 404 error is thrown when the cart is not found, ensuring that the app can deal with it in the same way it deals with a platform 404 error.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/cart/testing/src/drivers/in-memory/cart/cart.service.spec.ts
+++ b/libs/cart/testing/src/drivers/in-memory/cart/cart.service.spec.ts
@@ -1,10 +1,12 @@
 import { TestBed } from '@angular/core/testing';
 import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
 
-import { DaffCart } from '@daffodil/cart';
+import { DaffCart, DaffCartNotFoundError } from '@daffodil/cart';
 
 import { DaffCartFactory } from '../../../factories/cart.factory';
 import { DaffInMemoryCartService } from './cart.service';
+import { catchError } from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


libs/cart/testing/src/in-memory-backend/cart-root.service.ts
```diff
@@ -27,10 +27,12 @@ export class DaffInMemoryBackendCartRootService implements InMemoryDbService {
 
   constructor(
     private cartService: DaffInMemoryBackendCartService,
-    private cartItemsService: DaffInMemoryBackendCartItemsService,
+		private cartItemsService: DaffInMemoryBackendCartItemsService,
+		private storageService: DaffCartStorageService
   ) {}
 
   createDb(reqInfo: RequestInfo) {
+		this.storageService.removeCartId();
```
griest024: This will throw an error in SSR. Not sure if that is a problem here, just wanted to make sure that you have considered it and will handle the error if appropriate.

lderrickable: yea, you're right, there's definitely issues with this. Going to change

lderrickable: completely changed how it works

PR #738, opened 2020-03-25
bug(cart): fix bug where cart mutation was not detected because the o…

lderrickable:
…bject reference point did not change

## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When using the in memory api, redux doesn't detect a change to the cart items, because the reference point to the cart does not change in memory. In other words, when we do a `cart.items.push(newCartItem)`, the `cart.items` property does not change in memory, and the memoized selectors don't register that a change in the items array has occurred. A reproduction of this bug can be done by adding and removing items to and from the cart and watching the DaffCartFacade.items$ property. The selector behind that property will not detect a change, so it will not notify its Observables. 

Fixes: N/A

## What is the new behavior?
Every time a mutation to the cart occurs in memory, a brand new object is generated and set to the mutated part of the cart.

```
[ ] Yes
[x] No
```

## Other information
This fix is impossible to test without a kind of integration test of the whole cart package, so no tests were added.

### DIFF
--- a/libs/cart/testing/src/in-memory-backend/cart/cart.service.ts
+++ b/libs/cart/testing/src/in-memory-backend/cart/cart.service.ts
@@ -56,6 +56,7 @@ export class DaffInMemoryBackendCartService implements InMemoryDbService {
     const cart = this.getCart(reqInfo);
 
     cart.items = [];
+		cart.items = Object.assign([], cart.items);
 
     return cart
   }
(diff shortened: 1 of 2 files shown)


libs/cart/testing/src/in-memory-backend/cart/cart.service.ts
```diff
@@ -56,6 +56,7 @@ export class DaffInMemoryBackendCartService implements InMemoryDbService {
     const cart = this.getCart(reqInfo);
 
     cart.items = [];
+		cart.items = Object.assign([], cart.items);
```
griest024: I don't think this one is necessary. The previous line already sets `cart.items` to a new object.

lderrickable: removed

PR #742, opened 2020-03-26
feat(product): add price to magento product responses

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
There are no tests for this file yet, so I didn't add a test for this change. I'll probably add tests for the whole file when I do Composite product stuff

### DIFF
--- a/libs/product/src/drivers/magento/transforms/product-transformer.service.ts
+++ b/libs/product/src/drivers/magento/transforms/product-transformer.service.ts
@@ -75,10 +75,10 @@ export class DaffMagentoProductTransformerService {
 	 * A function for null checking an object.
 	 */
 	private getPrice(product: ProductNode): string {
-		return !product.price || 
-			!product.price.regularPrice || 
-			!product.price.regularPrice.amount || 
-			!!product.price.regularPrice.amount.value
-		? product.price.regularPrice.amount.value.toString() : '';
+		return !product.price_range || 
+			!product.price_range.maximum_price || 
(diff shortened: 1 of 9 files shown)


libs/product/src/drivers/magento/transforms/product-transformer.service.ts
```diff
@@ -68,4 +70,15 @@ export class DaffMagentoProductTransformerService {
...
+	/**
+	 * A function for null checking an object.
+	 */
+	private getPrice(product: ProductNode): string {
+		return !product.price || 
+			!product.price.regularPrice || 
+			!product.price.regularPrice.amount || 
+			!!product.price.regularPrice.amount.value
+		? product.price.regularPrice.amount.value.toString() : '';
+	}
```
griest024: I think you can use optional chaining to simplify this to
```ts
private getPrice(product: ProductNode): string {
		return product?.price?.regularPrice?.amount?.value ? product.price.regularPrice.amount.value.toString() : '';
	}
```

griest024: ah nvm I don't think TS optional chaining actually works like that

griest024: We are not on a version of TS that has this feature. RIP.

PR #745, opened 2020-03-26
fix(cart): fix bug where clearing a cart with more than one item failed

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The cartDriver.clear() method tries to remove all the cart items at the same time but through different cart item requests. The logic in cartDriver.clear() assumes that there will be a "final" cart item remove request, and that the payload of that request will have no cart items. Because these cart item requests are asynchronous, there is rarely a case where any cart item remove responses have no items when the original cart has more than one item.

Fixes: N/A

## What is the new behavior?
A bit of a hacky patch that will be readdressed in the future. It is assumed that the cart item requests are successful (though individual cart item remove errors will still bubble through), and just returns an otherwise updated cart with no items.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/cart/src/drivers/magento/cart.service.ts
+++ b/libs/cart/src/drivers/magento/cart.service.ts
@@ -1,7 +1,7 @@
 import { Injectable, Inject } from '@angular/core';
 import { Apollo } from 'apollo-angular';
 
-import { Observable, zip, of, throwError } from 'rxjs';
+import { Observable, zip, of, throwError, forkJoin } from 'rxjs';
 import { map, switchMap, catchError } from 'rxjs/operators';
 
 import { DaffCartServiceInterface } from '../interfaces/cart-service.interface';
@@ -57,15 +57,26 @@ export class DaffMagentoCartService implements DaffCartServiceInterface<DaffCart
       switchMap(items =>
 				// make the delete requests in parallel and collect them into a single observable
 				// return null if there are no items in the cart
-        items.length ? zip(...items.map(item =>
+        items.length ? forkJoin(...items.map(item =>
           this.cartItemDriver.delete(cartId, item.item_id)
         )) : of(null)
       ),
-      // find the most up to date delete response by looking for one with no items
-      // make a get request if there were no items in the cart
+			// make a get request if there were no items in the cart
+			// todo: find a better way to do this. Because all of the remove item requests are made
+			// asynchronously, there is no guarantee that any cart response will be the "last" one, i.e.
+			// the one with no more items.
 			switchMap(updatedCarts => updatedCarts ? 
-				of(updatedCarts.filter(cart => cart.items.length === 0).shift()) : 
+				of(this.removeLeftoverItems(updatedCarts)) : 
 				this.get(cartId))
     )
-  }
+	}
+	
+	/**
+	 * This assumes that the remove requests all succeed, and manually sets the cart items to an empty array.
+	 */
+	private removeLeftoverItems(carts: DaffCart[]): DaffCart {
+		carts[0].items = [];
+
+		return carts[0];
+	}
 }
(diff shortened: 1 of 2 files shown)


griest024:
@lderrickable does it work to make a `get` request after all the delete item requests complete in all cases, not just when there are no items?

libs/cart/src/drivers/magento/cart-item.service.ts
```diff
@@ -94,7 +94,10 @@ export class DaffMagentoCartItemService implements DaffCartItemServiceInterface<
         itemId
       }
     }).pipe(
-      map(result => this.cartTransformer.transform(result.data.removeItemFromCart.cart))
+      map(result => {
+				console.log(result.data.removeItemFromCart.cart);
```
damienwebdev: log

libs/cart/src/drivers/magento/cart.service.ts
```diff
@@ -57,15 +57,26 @@ export class DaffMagentoCartService implements DaffCartServiceInterface<DaffCart
...
           this.cartItemDriver.delete(cartId, item.item_id)
         )) : of(null)
       ),
-      // find the most up to date delete response by looking for one with no items
-      // make a get request if there were no items in the cart
+			// make a get request if there were no items in the cart
+			// todo: find a better way to do this. Because all of the remove item requests are made
+			// asynchronously, there is no guarantee that any cart response will be the "last" one, i.e.
+			// the one with no more items.
 			switchMap(updatedCarts => updatedCarts ?
```
damienwebdev: We can remove the condition here.

libs/cart/src/drivers/magento/cart.service.ts
```diff
@@ -57,15 +57,26 @@ export class DaffMagentoCartService implements DaffCartServiceInterface<DaffCart
...
+				of(this.removeLeftoverItems(updatedCarts)) : 
 				this.get(cartId))
     )
-  }
+	}
+	
+	/**
+	 * This assumes that the remove requests all succeed, and manually sets the cart items to an empty array.
+	 */
+	private removeLeftoverItems(carts: DaffCart[]): DaffCart {
```
damienwebdev: We don't need this.

libs/cart/src/drivers/magento/cart.service.ts
lderrickable: We still need it in case you try to clear an empty cart. This was a bug from before

libs/cart/src/drivers/magento/cart-item.service.ts
lderrickable: oops, removed

lderrickable:
> @lderrickable does it work to make a `get` request after all the delete item requests complete in all cases, not just when there are no items? So instead of looking for the delete item request that has no items, just wait for them all to complete and then make a `get` request. That should ensure that the final response includes the most up to date data.

This should work, but there's a weird bug happening where the second time you try to clear the cart it never works. Apollo just never makes the getCart graphQL call and uses the cached cart instead.

griest024:
> This should work, but there's a weird bug happening where the second time you try to clear the cart it never works. Apollo just never makes the getCart graphQL call and uses the cached cart instead.

I see...this caching happens client-side? I think we would not want this behavior in general.

lderrickable:
> > This should work, but there's a weird bug happening where the second time you try to clear the cart it never works. Apollo just never makes the getCart graphQL call and uses the cached cart instead.
> 
> I see...this caching happens client-side? I think we would not want this behavior in general.

Turned out to be a fairly complicated issue, possibly involving those heuristic fragment errors we've been getting. Bug still exists, but we're going to fix it a different way in the future.

PR #751, opened 2020-03-31
feat(cart): make cart package generic

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[x] Yes
[ ] No
```

## Other information
I changed the place order button to use the DaffCartFacade instead of the DaffCartContainer, because it looked like the the testing would have been difficult to fix otherwise.

### DIFF
--- a/apps/demo/src/app/cart/routing-resolvers/effects/cart-resolver.effects.ts
+++ b/apps/demo/src/app/cart/routing-resolvers/effects/cart-resolver.effects.ts
@@ -9,8 +9,7 @@ import {
 	DaffCart,
 	DaffCartDriver,
 	DaffCartServiceInterface,
-	DaffCartStorageService,
-	getDaffCartSelectors,
+	DaffCartStorageService
 } from '@daffodil/cart';
 
 import {
@@ -18,6 +17,7 @@ import {
 	CartResolverActionTypes,
 	ResolveCartFailure,
 } from '../actions/cart-resolver.actions';
+import { daffCartSelectors } from '../../selectors/cart-selector';
 
 @Injectable()
 export class CartResolverEffects {
@@ -46,7 +46,7 @@ export class CartResolverEffects {
 	);
 
 	selectStoreCart(): Observable<DaffCart> {
-		return this.store.pipe(select(getDaffCartSelectors<DaffCart>().selectCartValue));
+		return this.store.pipe(select(daffCartSelectors.selectCartValue));
 	}
 
 	private getCartHandler(): Observable<Action> {
--- a/apps/demo/src/app/cart/routing-resolvers/effects/cart-resolver.effects.ts
+++ b/apps/demo/src/app/cart/routing-resolvers/effects/cart-resolver.effects.ts
@@ -10,6 +10,7 @@ import {
 	DaffCartDriver,
 	DaffCartServiceInterface,
 	DaffCartStorageService,
+	getDaffCartSelectors,
 } from '@daffodil/cart';
 
 import {
@@ -45,7 +46,7 @@ export class CartResolverEffects {
 	);
 
 	selectStoreCart(): Observable<DaffCart> {
-		return this.store.pipe(select(fromCart.selectCartValue()));
+		return this.store.pipe(select(getDaffCartSelectors<DaffCart>().selectCartValue));
 	}
 
 	private getCartHandler(): Observable<Action> {
--- a/apps/demo/src/app/cart/routing-resolvers/effects/cart-resolver.effects.ts
+++ b/apps/demo/src/app/cart/routing-resolvers/effects/cart-resolver.effects.ts
@@ -22,7 +22,7 @@ import {
 export class CartResolverEffects {
 	constructor(
 		private actions$: Actions,
-		private store: Store<fromCart.State>,
+		private store: Store<fromCart.State<DaffCart>>,
 		private cartStorage: DaffCartStorageService,
 		@Inject(DaffCartDriver) private driver: DaffCartServiceInterface<DaffCart>,
 	) {}
@@ -45,7 +45,7 @@ export class CartResolverEffects {
 	);
 
 	selectStoreCart(): Observable<DaffCart> {
-		return this.store.pipe(select(fromCart.selectCartValue));
+		return this.store.pipe(select(fromCart.selectCartValue()));
 	}
 
 	private getCartHandler(): Observable<Action> {
--- a/apps/demo/src/app/cart/selectors/cart-selector.ts
+++ b/apps/demo/src/app/cart/selectors/cart-selector.ts
@@ -1,9 +1,9 @@
 import { createSelector, MemoizedSelector, ActionReducerMap } from '@ngrx/store';
 
-import { fromCart,fromCartReducer } from '@daffodil/cart';
+import { fromCart,fromCartReducer, DaffCart } from '@daffodil/cart';
 
 export interface CartState {
-  cart: fromCartReducer.State;
+  cart: fromCartReducer.State<DaffCart>;
 }
 
 export interface State {
@@ -15,7 +15,7 @@ export const reducers : ActionReducerMap<CartState> = {
 }
 
 export const selectCartItemCount : MemoizedSelector<object, number> = createSelector(
-  fromCart.selectCartValue,
+  fromCart.selectCartValue(),
   cart => {
     let itemCount = 0;
     cart.items.forEach(cartItem => {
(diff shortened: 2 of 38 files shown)


griest024:
@lderrickable What do you think about using generic type defaults? It would make the usage less verbose on the application side when not using custom types.

libs/cart/src/facades/cart/cart.facade.ts
```diff
@@ -65,33 +65,33 @@ export class DaffCartFacade implements DaffStoreFacade<Action> {
   availablePaymentMethods$: Observable<DaffCart['available_payment_methods']>;
   isCartEmpty$: Observable<boolean>;
 
-  constructor(private store: Store<DaffCartReducersState>) {
-    this.loading$ = this.store.pipe(select(selectCartLoading));
-    this.cart$ = this.store.pipe(select(selectCartValue));
+  constructor(private store: Store<DaffCartReducersState<T>>) {
+    this.loading$ = this.store.pipe(select(selectCartLoading()));
```
griest024: Should you be passing `T` into the selectors?

libs/cart/src/cart-state.module.ts
```diff
@@ -15,7 +15,7 @@ import { DaffCartPaymentMethodsEffects } from './effects/cart-payment-methods.ef
 
 @NgModule({
   imports: [
-    StoreModule.forFeature('cart', daffCartReducers),
+    StoreModule.forFeature('cart', daffCartReducers()),
```
griest024: You made `daffCartReducers` generic but how would a type parameter be passed into it?

lderrickable: The way I understand it, it all comes together through the driver. If you don't provide a correct DaffCartDriver that extends the DaffCartServiceInterface, then you should get a build error (not sure it actually does right now though, because injection tokens can't be typed yet). Then that makes it incompatible with the effects layer, which makes it incompatible with the reducers layer. I'm not super solid on how exactly everything with generics work, but it seems like when an instance of something doesn't care about what the concrete object is, it's ok to stay generic. Here, the daffCartReducers() function doesn't require a type other than `T extends DaffCart`, so the generic version is enough.

libs/cart/src/facades/cart/cart.facade.ts
lderrickable: I don't think so. Any selector daffodil defines will only ever access the base `DaffCart` model. It shouldn't ever need to know specifically what the model is. Just that it is `T extends DaffCart`.

lderrickable:
> @lderrickable What do you think about using generic type defaults? It would make the usage less verbose on the application side when not using custom types.

I'm not exactly sure how we would do this. The only thing I can think of is to have a sort of specific model for the reducer state, facade, etc.:
```
export interface DaffSpecificCartReducersState extends DaffCartReducersState<DaffCart> {}
```

But I don't really like that idea. I had to do something sort of similar in the navigation package, because the DaffNavigationTree is a recursive model. But I had to do it there. Can you think of another way?

libs/cart/src/facades/cart/cart.facade.ts
```diff
@@ -37,7 +37,7 @@ import { DaffCartErrorType } from '../../reducers/cart-error-type.enum';
 @Injectable({
   providedIn: 'root'
 })
-export class DaffCartFacade implements DaffStoreFacade<Action> {
+export class DaffCartFacade<T extends DaffCart> implements DaffStoreFacade<Action> {
   loading$: Observable<boolean>;
   cart$: Observable<DaffCart>;
```
griest024: Should this be `cart$: Observable<T>` ? That would probably mean that you do need to pass the generic into the selector, at least for this one.

lderrickable: Yea. I made the selectors that don't need to be generic not generic too.

libs/cart/src/facades/cart/cart.facade.ts
lderrickable: I changed this to include `T` for selectors that need it

lderrickable:
> @lderrickable What do you think about using generic type defaults? It would make the usage less verbose on the application side when not using custom types.

I'm actually going to do this in another PR. It's a lot of little changes

apps/demo/src/app/checkout/components/place-order/place-order.component.ts
```diff
@@ -19,11 +18,14 @@ export class PlaceOrderComponent implements OnInit {
...
+		private cartFacade: DaffCartFacade<DaffCart>
+	) { }
 
   ngOnInit() {
     this.enablePlaceOrderButton$ = this.store.pipe(select(fromDemoCheckout.selectEnablePlaceOrderButton));
-    this.cart$ = this.store.pipe(select(fromCart.selectCartValue));
+    this.cart$ = this.cartFacade.cart$;
   }
 
   placeOrder() {
```
griest024, line 31: Can you dispatch with the facade here so you can remove the store from this component?

lderrickable: I can't remove the dependency on store for this component, because it references demo store and not just daffodil store. But I did change the dispatch to go through the facade

libs/cart/src/cart-state.module.ts
damienwebdev, line 18: Does this need a call? Can the underlying function just be a reference?

libs/cart/src/containers/cart/cart.component.ts
```diff
@@ -6,29 +6,27 @@ import { Store, select } from '@ngrx/store';
...
-    private store: Store<fromCart.DaffCartReducersState>
+    private store: Store<fromCart.DaffCartReducersState<T>>
   ) { }
 
   ngOnInit() {
     this.store.dispatch(new DaffCartLoad());
 
-    this.loading$ = this.store.pipe(select(selectCartLoading));
-    this.cart$ = this.store.pipe(select(selectCartValue));
+    this.loading$ = this.store.pipe(select(getDaffCartSelectors<T>().selectCartLoading));
```
damienwebdev, line 28: Can the initial call be moved to a const either in the ngOnInit or constructor?

libs/cart/src/facades/cart/cart.facade.ts
```diff
@@ -65,33 +40,32 @@ export class DaffCartFacade implements DaffStoreFacade<Action> {
   availablePaymentMethods$: Observable<DaffCart['available_payment_methods']>;
   isCartEmpty$: Observable<boolean>;
 
-  constructor(private store: Store<DaffCartReducersState>) {
-    this.loading$ = this.store.pipe(select(selectCartLoading));
-    this.cart$ = this.store.pipe(select(selectCartValue));
+  constructor(private store: Store<DaffCartReducersState<T>>) {
+    this.loading$ = this.store.pipe(select(getDaffCartSelectors<T>().selectCartLoading));
```
damienwebdev, line 44: Can the initial call be moved to a const either in the ngOnInit or constructor?

libs/cart/src/cart-state.module.ts
lderrickable: changed

libs/cart/src/containers/cart/cart.component.ts
lderrickable: done

libs/cart/src/facades/cart/cart.facade.ts
lderrickable: done

apps/demo/src/app/cart/routing-resolvers/effects/cart-resolver.effects.ts
```diff
@@ -45,7 +46,7 @@ export class CartResolverEffects {
 	);
 
 	selectStoreCart(): Observable<DaffCart> {
-		return this.store.pipe(select(fromCart.selectCartValue));
+		return this.store.pipe(select(getDaffCartSelectors<DaffCart>().selectCartValue));
```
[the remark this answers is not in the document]
lderrickable: done

apps/demo/src/app/cart/selectors/cart-selector.ts
```diff
@@ -15,7 +15,7 @@ export const reducers : ActionReducerMap<CartState> = {
 }
 
 export const selectCartItemCount : MemoizedSelector<object, number> = createSelector(
-  fromCart.selectCartValue,
+  getDaffCartSelectors<DaffCart>().selectCartValue,
```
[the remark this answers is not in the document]
lderrickable: done

PR #752, opened 2020-03-31
feat(tools): add generate fragment types builder

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Daffodil does not provide a packaged way to generate GraphQL fragment data.

## What is the new behavior?
Daffodil provides a new package `@daffodil/builders` which contains a builder for generating GraphQL fragment type data.

```
[ ] Yes
[ ] No
```

## Other information
Tests are causing issues at the moment so they have been omitted.


lderrickable:
What do you mean by this? `Tests are causing issues at the moment so they have been omitted.`

griest024:
@lderrickable Importing the `@graphql-codegen/cli` package in the testing environment is what the root of the issues are, I think. It balloons the compilation to 2500+ modules and throws a million lines of errors to STDOUT, mostly about browser not being a valid alias configuration or something like that.

lderrickable:
Could you make it so that someone can follow the docs and it works? I followed the docs and it doesn't seem to work, but maybe I'm doing something wrong. I don't see how it could work though because the build script references `@daffodil/builders` but I don't think that's a thing

griest024:
I don't think I have perms to publish `@daffodil` packages but this would be a published package. For testing I used `npm link` to test. I can open a PR with an implementation in demo if it makes it easier.

lderrickable:
We wouldn't publish it until after it got pushed to develop anyway. If it is publishable, we should be able to still run it from the dist folder, right? There is currently no way to build it into the dist folder I don't think.

griest024:
Its dependent on file structure so I left it building to its own personal dist folder for initial testing. I'm not sure how the publish/release process works with daffodil packages but I can make it build to the main dist folder if that is a necessary part of it. I think it needs to read `package.json` in order to run the builder though, so just running based off what is in the `dist` folder is not possible.

package.json
```diff
@@ -52,13 +52,16 @@
     "zone.js": "~0.9.1"
   },
   "devDependencies": {
+    "@angular-devkit/architect": "^0.901.0",
```
damienwebdev, line 55: Can you pull this down to pre-ivy?

griest024: changed

PR #760, opened 2020-04-03
feat(auth): add reducers

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information


libs/auth/src/reducers/auth-initial-state.ts
```diff
@@ -0,0 +1,8 @@
+import { DaffAuthReducerState } from './auth-reducer-state.interface';
+import { DaffAuthToken } from '../models/auth-token';
+
+export const daffAuthInitialState: DaffAuthReducerState<any> = Object.freeze({
```
damienwebdev, line 4: Why freeze the Object this is inconsistent with other packages?

libs/auth/src/reducers/auth-reducer-state.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { DaffAuthToken } from '../models/auth-token';
+
+export interface DaffAuthReducerState<T extends DaffAuthToken> {
```
damienwebdev, line 3: I think this model is wrong. "login" and "register" probably have split "Loadable" states (errors + loading).

griest024: that makes sense

libs/auth/src/reducers/auth-initial-state.ts
griest024: Unless we specifically want to allow the initial state to be modified it seems safest to freeze it.

griest024: Removed `Object.freeze` from the initial states

PR #802, opened 2020-04-14
feat(paypal): provide default values for generics

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/paypal/src/drivers/magento/paypal.service.ts
+++ b/libs/paypal/src/drivers/magento/paypal.service.ts
@@ -16,15 +16,15 @@ import { DaffPaypalConfig } from '../injection-tokens/paypal-config.token';
 @Injectable({
   providedIn: 'root'
 })
-export class DaffMagentoPaypalService<T extends DaffPaypalTokenRequest, V extends DaffPaypalTokenResponse> implements DaffPaypalServiceInterface<T, V> {
+export class DaffMagentoPaypalService implements DaffPaypalServiceInterface {
   
   constructor(
     private apollo: Apollo,
-		@Inject(DaffPaypalTransformer) private transformer: DaffPaypalTransformerInterface<T, V>,
+		@Inject(DaffPaypalTransformer) private transformer: DaffPaypalTransformerInterface<DaffPaypalTokenRequest, DaffPaypalTokenResponse>,
 		@Inject(DaffPaypalConfig) private config: DaffMagentoPaypalConfig
 	) {}
 
-  generateToken(tokenRequest: T): Observable<V> {
+  generateToken(tokenRequest: DaffPaypalTokenRequest): Observable<DaffPaypalTokenResponse> {
     return this.apollo.mutate<MagentoPaypalTokenResponse>({
       mutation: GenerateTokenMutation,
       variables: {
--- a/libs/paypal/src/selectors/paypal.selector.ts
+++ b/libs/paypal/src/selectors/paypal.selector.ts
@@ -4,7 +4,7 @@ import { DaffPaypalReducerState } from '../reducers/paypal/paypal-reducer.interf
 import { DaffPaypalReducersState } from '../reducers/paypal-reducers.interface';
 import { DaffPaypalTokenResponse } from '../models/paypal-token-response';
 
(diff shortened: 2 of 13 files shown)


libs/paypal/src/drivers/magento/paypal.service.ts
```diff
@@ -16,15 +16,15 @@ import { DaffPaypalConfig } from '../injection-tokens/paypal-config.token';
 @Injectable({
   providedIn: 'root'
 })
-export class DaffMagentoPaypalService<T extends DaffPaypalTokenRequest, V extends DaffPaypalTokenResponse> implements DaffPaypalServiceInterface<T, V> {
+export class DaffMagentoPaypalService implements DaffPaypalServiceInterface {
   
   constructor(
     private apollo: Apollo,
-		@Inject(DaffPaypalTransformer) private transformer: DaffPaypalTransformerInterface<T, V>,
+		@Inject(DaffPaypalTransformer) private transformer: DaffPaypalTransformerInterface<DaffPaypalTokenRequest, DaffPaypalTokenResponse>,
```
griest024, line 23: The transformer interface and injection token can be removed. That can be done in a separate PR though.

libs/paypal/src/selectors/paypal.selector.ts
```diff
@@ -53,7 +53,7 @@ const createPaypalSelectors = <T extends DaffPaypalTokenResponse>(): DaffPaypalM
 
 export const daffPaypalSelectors = (() => {
 	let cache;
-	return <T extends DaffPaypalTokenResponse>(): DaffPaypalMemoizedSelectors<T> => cache = cache 
+	return <T extends DaffPaypalTokenResponse = DaffPaypalTokenResponse>(): DaffPaypalMemoizedSelectors<T> => cache = cache
```
griest024, line 56: I don't think this is necessary.

lderrickable:
> I think you forgot to do the effects class as well.

I didn't do the effects, because I figured they'd never be referenced anywhere ever

libs/paypal/src/selectors/paypal.selector.ts
lderrickable: why not?

libs/paypal/src/drivers/magento/paypal.service.ts
lderrickable: yea, will do in another PR

libs/paypal/src/selectors/paypal.selector.ts
griest024: without the default value set, `daffPaypalSelectors()` still has the type `DaffPaypalMemoizedSelectors<DaffPaypalTokenResponse>`. Check to make sure thats what you see as well.

libs/paypal/src/selectors/paypal.selector.ts
```diff
@@ -53,7 +53,7 @@ const createPaypalSelectors = <T extends DaffPaypalTokenResponse>(): DaffPaypalM
 
 export const daffPaypalSelectors = (() => {
```
griest024, line 54: Should be `getDaffPaypalSelectors`

libs/paypal/src/selectors/paypal.selector.ts
lderrickable: hmmm, yea that's what I see too. Seems pretty weird though. Like maybe typescript automatically picks a default for function calls that are generic. I think it should probably stay here though, so it's clear that this will pick a default for you. I'd rather rely on something more explicit, rather than some obscure rule buried in typescript somewhere. What do you think?

libs/paypal/src/selectors/paypal.selector.ts
lderrickable: done

libs/paypal/src/selectors/paypal.selector.ts
griest024: yeah I agree, its probably better to be explicit. You should make this change in product and category as well.

lderrickable: It should be done in product and category already

PR #814, opened 2020-04-15
feat(cart): add cart-order service interface

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information
I made `payment` a nullable parameter since Shopfiy requires it but Magento does not.

### DIFF
--- a/libs/cart/src/drivers/interfaces/cart-order-service.interface.ts
+++ b/libs/cart/src/drivers/interfaces/cart-order-service.interface.ts
@@ -5,14 +5,14 @@ import { DaffCart } from '../../models/cart';
 import { DaffCartPaymentMethod } from '../../models/cart-payment';
 
 /**
- * The interface responsible for managing a customer's cart.
+ * The interface responsible for placing an order for the customer's cart.
  */
 export interface DaffCartOrderServiceInterface<
   T extends DaffCart = DaffCart,
   V extends DaffCartPaymentMethod = DaffCartPaymentMethod
 > {
 	/**
-	 * Place an order.
+	 * Place an order and return the order ID.
 	 */
   placeOrder(id: T['id'], payment?: V): Observable<{id: string | number}>;
 }
--- a/libs/cart/src/drivers/interfaces/cart-order-service.interface.ts
+++ b/libs/cart/src/drivers/interfaces/cart-order-service.interface.ts
@@ -7,11 +7,14 @@ import { DaffCartPaymentMethod } from '../../models/cart-payment';
 /**
  * The interface responsible for managing a customer's cart.
  */
-export interface DaffCartOrderServiceInterface<T extends DaffCart = DaffCart> {
+export interface DaffCartOrderServiceInterface<
+  T extends DaffCart = DaffCart,
+  V extends DaffCartPaymentMethod = DaffCartPaymentMethod
+> {
 	/**
 	 * Place an order.
 	 */
-  placeOrder(id: T['id'], payment?: DaffCartPaymentMethod): Observable<T>;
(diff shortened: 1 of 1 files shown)


libs/cart/src/drivers/interfaces/cart-order-service.interface.ts
```diff
@@ -0,0 +1,19 @@
...
+import { DaffCartPaymentMethod } from '../../models/cart-payment';
+
+/**
+ * The interface responsible for managing a customer's cart.
+ */
+export interface DaffCartOrderServiceInterface<T extends DaffCart = DaffCart> {
+	/**
+	 * Place an order.
+	 */
+  placeOrder(id: T['id'], payment?: DaffCartPaymentMethod): Observable<T>;
```
griest024, line 14: Not sure what the return should be. If we make it a `DaffOrder` object then that imposes a dep from cart onto checkout.

damienwebdev: I think we should return a `string | number` for the return type (it would be the `id` of the resultant order).

damienwebdev, line 14: `DaffCartPaymentMethod` can be off the `DaffCart` interface, right?

griest024: @damienwebdev so return `{id: string | number}` like we did with `CartService#create`?

griest024: Yeah but it should actually be generic right?

damienwebdev: Yes, I think that makes sense.

damienwebdev: Yes, it should be.

griest024: done

griest024: done

libs/cart/src/drivers/interfaces/cart-order-service.interface.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * The interface responsible for managing a customer's cart.
+ */
+export interface DaffCartOrderServiceInterface<
+  T extends DaffCart = DaffCart,
+  V extends DaffCartPaymentMethod = DaffCartPaymentMethod
+> {
+	/**
+	 * Place an order.
```
damienwebdev, line 15: Can you update this to comment on the intent of the return type?

libs/cart/src/drivers/interfaces/cart-order-service.interface.ts
```diff
@@ -0,0 +1,22 @@
+import { InjectionToken } from '@angular/core';
+import { Observable } from 'rxjs';
+
+import { DaffCart } from '../../models/cart';
+import { DaffCartPaymentMethod } from '../../models/cart-payment';
+
+/**
+ * The interface responsible for managing a customer's cart.
```
damienwebdev, line 8: This doc is wrong.

griest024: fixed

libs/cart/src/drivers/interfaces/cart-order-service.interface.ts
griest024: done

PR #823, opened 2020-04-16
feat(cart): add place order Magento driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Adds a place order driver
- Adds a bad input error
- Changes `transformError` to use the general version in `@daffodil/driver/magento`

```
[ ] Yes
[ ] No
```

## Other information


libs/cart/src/errors/bad-input.ts
```diff
@@ -0,0 +1,8 @@
+export class DaffBadInputError extends Error {
```
lderrickable, line 1: This seems like it should go in core, but I'm not sure. Or maybe it should be named `DaffBadCartInputError`

griest024: Yeah you're right. This will prob be used in most of the Magento drivers.

griest024: I'll put it in `driver/magento`

libs/cart/src/drivers/magento/cart-order.service.ts
```diff
@@ -0,0 +1,37 @@
...
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoCartOrderService implements DaffCartOrderServiceInterface<DaffCart> {
+  constructor(
+    private apollo: Apollo,
+    public cartTransformer: DaffMagentoCartTransformer,
+  ) {}
+
+  placeOrder(cartId: DaffCart['id']) {
```
lderrickable, line 26: could use a return type

libs/cart/src/errors/bad-input.ts
griest024: done

libs/cart/src/drivers/magento/cart-order.service.ts
griest024: fixed

libs/cart/src/errors/bad-input.ts
griest024: I also changed the transform error function to use the one in `driver`.

libs/driver/magento/src/errors/public_api.ts
```diff
@@ -1 +1,2 @@
 export { daffTransformMagentoError } from './transform';
+export { DaffBadInputError } from './bad-input';
```
lderrickable, line 2: Seems to me like the `daffTransformMagentoError` should be here is `@daffodil/driver/magento`, but the actual `DaffBadInputError` should be in `@daffodil/driver`

griest024: hm that's a good point, it does seem like it'll be useful for all platform drivers.

griest024: moved it

lderrickable:
lgtm

PR #824, opened 2020-04-17
feat(cart): add cart coupon driver service interface

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
I made some of the `DaffCartCoupon` fields optional, because magento seems to only require the actual coupon code.

### DIFF
--- a/libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
+++ b/libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
@@ -5,13 +5,23 @@ import { DaffCartCoupon } from '../../models/cart-coupon';
 import { DaffCart } from '../../models/cart';
 
 /**
- * The interface responsible for apply a coupon to a cart.
+ * The interface responsible for applying a coupon to a cart.
  */
 export interface DaffCartCouponServiceInterface<T extends DaffCart = DaffCart, V extends DaffCartCoupon = DaffCartCoupon> {
 	/**
 	 * Apply a coupon to the cart and return a partial of the cart.
 	 */
-  applyCoupon(cartId: T['id'], coupon: V): Observable<Partial<T>>;
+	apply(cartId: T['id'], coupon: V): Observable<Partial<T>>;
+	
+	/**
+	 * Remove a coupon from the cart and return a partial of the cart.
+	 */
+	remove(cartId: T['id'], coupon: V): Observable<Partial<T>>;
+
+	/**
+	 * Remove all coupons from the cart and return a partial of the cart.
+	 */
+	removeAll(cartId: T['id']): Observable<Partial<T>>;
 }
 
 export const DaffCartCouponDriver = new InjectionToken<DaffCartCouponServiceInterface>(
--- a/libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
+++ b/libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
@@ -12,6 +12,11 @@ export interface DaffCartCouponServiceInterface<T extends DaffCart = DaffCart, V
 	 * Apply a coupon to the cart and return a partial of the cart.
 	 */
 	apply(cartId: T['id'], coupon: V): Observable<Partial<T>>;
+
+	/**
+	 * List coupon codes applied to a cart.
+	 */
+	list(cartId: T['id']): Observable<Partial<T>>;
 	
 	/**
 	 * Remove a coupon from the cart and return a partial of the cart.
--- a/libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
+++ b/libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
@@ -1,0 +1,19 @@
+import { InjectionToken } from '@angular/core';
+import { Observable } from 'rxjs';
+
+import { DaffCartCoupon } from '../../models/cart-coupon';
+import { DaffCart } from '../../models/cart';
+
+/**
+ * The interface responsible for apply a coupon to a cart.
+ */
+export interface DaffCartCouponServiceInterface<T extends DaffCart = DaffCart, V extends DaffCartCoupon = DaffCartCoupon> {
+	/**
+	 * Apply a coupon to the cart and return a partial of the cart.
+	 */
+  applyCoupon(cartId: T['id'], coupon: V): Observable<Partial<T>>;
+}
+
+export const DaffCartCouponDriver = new InjectionToken<DaffCartCouponServiceInterface>(
+	'DaffCartCouponDriver',
+);
--- a/libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
+++ b/libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
@@ -16,7 +16,7 @@ export interface DaffCartCouponServiceInterface<T extends DaffCart = DaffCart, V
 	/**
 	 * List coupon codes applied to a cart.
 	 */
-	list(cartId: T['id']): Observable<Partial<T>>;
+	list(cartId: T['id']): Observable<V[]>;
 	
 	/**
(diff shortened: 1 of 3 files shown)


libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
```diff
@@ -0,0 +1,19 @@
+import { InjectionToken } from '@angular/core';
+import { Observable } from 'rxjs';
+
+import { DaffCartCoupon } from '../../models/cart-coupon';
+import { DaffCart } from '../../models/cart';
+
+/**
+ * The interface responsible for apply a coupon to a cart.
```
griest024, line 8: *for applying

libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
```diff
@@ -0,0 +1,19 @@
...
+ */
+export interface DaffCartCouponServiceInterface<T extends DaffCart = DaffCart, V extends DaffCartCoupon = DaffCartCoupon> {
+	/**
+	 * Apply a coupon to the cart and return a partial of the cart.
+	 */
+  applyCoupon(cartId: T['id'], coupon: V): Observable<Partial<T>>;
```
damienwebdev, line 14: Should we be able to add/remove/list all coupons on a cart out of the gate, or should we wait for another PR?

lderrickable: I always thought of coupons as coupon codes that you only know about if you've been sent an email, or you follow someone on social media who advertises a product and provides a coupon code. Being able to add/list coupons from outside the admin panel doesn't make sense to me. Removing coupons I could see as useful, though I doubt anyone would really use it. I'll look into removing coupons

libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
lderrickable: fixed

lderrickable:
I changed things a bit. Also, magento only removes all coupons. You can't do one at a time, but I think daffodil should probably support that.

libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
```diff
@@ -0,0 +1,29 @@
+import { InjectionToken } from '@angular/core';
+import { Observable } from 'rxjs';
+
+import { DaffCartCoupon } from '../../models/cart-coupon';
+import { DaffCart } from '../../models/cart';
+
+/**
+ * The interface responsible for applying a coupon to a cart.
+ */
+export interface DaffCartCouponServiceInterface<T extends DaffCart = DaffCart, V extends DaffCartCoupon = DaffCartCoupon> {
```
damienwebdev, line 10: I think we need to be able to `list` all the coupons applied to the cart as well.

libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
xelaint: Lost Golf Balls actively advertises its coupon codes everywhere on the website. There is a section on the product page that adds relevant coupon codes to the product, so I do think it makes sense to have a section to list coupons available.

i.e. Sephora has a page that lists all of their available coupon codes, and it is referenced as a link on the cart page.

I think this is a separate PR though.

libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
lderrickable: wouldn't you just do a getCart call? To list the coupons I think you'd usually need to make a getCart query that only returns the coupons

lderrickable: I can't think of a scenario where you'd want the coupons applied to the cart without the cart itself.

griest024: Maybe if you had a checkout page just for the purpose of listing coupons and discounts? The other cart drivers have methods for getting the specific models so I think a `list` method would be consistent.

lderrickable: I don't think there's even a way to query for coupons that aren't associated with a cart. I think that's because you don't typically want everyone to know about all the coupon codes you're offering. I think that online coupons are usually used by people who are promoting your product to get people to visit the site, so that the seller can credit the promoter for bringing the customer to the site. Discounts might be a different story. Like maybe you could filter products by discounted ones.

griest024: @lderrickable I mean listing coupons on the cart. We provide `get` methods for other cart submodels and its nice to have a smaller query when you don't need the entire cart.

damienwebdev: My original point was what @griest024 is describing.

xelaint: LGB actively advertises promo codes on their website, everywhere. So it might actually be helpful to have a 'View Promo Codes' modal on the cart page by the promo code solo field that displays all the available promo codes. This is out of scope of this PR though and dependents on the client.

But I feel like you guys are discussing two different functionalities? Listing all available coupons vs listing applied coupons?

The latter, listing applied coupons is required. It's the same idea as listing applied filters. A user should be able to see and remove applied coupons.

lderrickable: ok I'll add it

xelaint: Stores often actively advertise promo codes on their website (i.e. Sephora has a separate page listing their promos). So it might actually be helpful to have a 'View Promo Codes' modal on the cart page by the promo code solo field that displays all the available promo codes. This is out of scope of this PR though.

But I feel like you guys are discussing two different functionalities? Listing all available coupons vs listing applied coupons?

The latter (listing applied coupons) is required anywhere adding a promo code is an option (cart, checkout, cart sidebar). A user should be able to see and remove applied coupons.

lderrickable: As far as I can tell, there is no way to list coupon codes by themselves apart from those actively applied to a cart. I looked at Sephora's website, and I suspect they have that page semi-hardcoded. The promo codes are actually shown as static images.

lderrickable: added

libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
```diff
@@ -0,0 +1,34 @@
...
+	/**
+	 * Apply a coupon to the cart and return a partial of the cart.
+	 */
+	apply(cartId: T['id'], coupon: V): Observable<Partial<T>>;
+
+	/**
+	 * List coupon codes applied to a cart.
+	 */
+	list(cartId: T['id']): Observable<Partial<T>>;
```
griest024, line 19: I think the return type should be `Observable<V[]>`.

libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
griest024: Yeah I think listing all coupons codes and promos that exist for a site is more like an admin panel thing than a storefront action. Definitely out of this PR's scope anyway.

libs/cart/src/drivers/interfaces/cart-coupon-service.interface.ts
lderrickable: yea, mb. Changed

PR #825, opened 2020-04-17
feat(cart): add apply/remove coupon magento mutation

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
The response of the coupon query is basically the same as cart minus all the shipping and payment info.

### DIFF
--- a/libs/cart/src/drivers/magento/queries/apply-coupon.ts
+++ b/libs/cart/src/drivers/magento/queries/apply-coupon.ts
@@ -1,0 +1,58 @@
+import gql from 'graphql-tag';
+
+import { cartItemFragment } from './fragments/cart-item';
+import { moneyFragment } from './fragments/money';
+
+export const applyCoupon = gql`
+  mutation ApplyCoupon($cartId: String!, $couponCode: String!) {
+    mutation {
+      applyCouponToCart(
+        input: {
+					cart_id: $cartId,
+					coupon_code: $couponCode
+        }
+      ) {
+				id
+				items {
+					...cartItem
+				}
+				applied_coupons {
+					code
+				}
+				prices {
+					grand_total {
+						...money
+					}
+					subtotal_excluding_tax {
+						...money
+					}
+					subtotal_including_tax {
+						...money
+					}
+					subtotal_with_discount_excluding_tax {
+						...money
+					}
+				}
+				applied_coupons {
+					code
+				}
+				items {
+					...cartItem
+				}
+				prices {
+					discounts {
+						label
+						amount {
+							...money
+						}
+					}
+					grand_total {
+						...money
+					}
+				}
+      }
+    }
+  }
+  ${cartItemFragment}
+  ${moneyFragment}
+`;
--- a/libs/cart/src/drivers/magento/queries/apply-coupon.ts
+++ b/libs/cart/src/drivers/magento/queries/apply-coupon.ts
@@ -1,7 +1,7 @@
 import gql from 'graphql-tag';
 
 import { cartItemFragment } from './fragments/cart-item';
-import { moneyFragment } from './fragments/money';
+import { pricesFragment } from './fragments/prices';
 
 export const applyCoupon = gql`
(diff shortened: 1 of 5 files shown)


lderrickable:
> The response is a cart object so I think you should use the cart fragment here:
> 
> ```graphql
> mutation ApplyCoupon($cartId: String!, $couponCode: String!) {
>     mutation {
>       applyCouponToCart(
>         input: {
> 		cart_id: $cartId,
> 		coupon_code: $couponCode
>         }
>       ) {
> 	  cart {
>              ...cart
>           }
>       }
>     }
>   }
>   ${cartFragment}
> ```

I didn't use it, because I think it's better to have a considerably smaller payload with each coupon request than to have the query code a bit cleaner.

libs/cart/src/drivers/magento/queries/apply-coupon.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export const applyCoupon = gql`
+  mutation ApplyCoupon($cartId: String!, $couponCode: String!) {
+    mutation {
+      applyCouponToCart(
+        input: {
+					cart_id: $cartId,
+					coupon_code: $couponCode
+        }
+      ) {
```
griest024, line 14: This needs to be wrapped in a `cart {}`

libs/cart/src/drivers/magento/queries/apply-coupon.ts
```diff
@@ -0,0 +1,58 @@
...
+						...money
+					}
+				}
+				applied_coupons {
+					code
+				}
+				items {
+					...cartItem
+				}
+				prices {
```
griest024, line 42: `prices` appears twice.

libs/cart/src/drivers/magento/queries/apply-coupon.ts
lderrickable: fixed

libs/cart/src/drivers/magento/queries/apply-coupon.ts
lderrickable: removed

lderrickable:
I added the removeAllCoupons mutation too

libs/cart/src/drivers/magento/queries/remove-coupon.ts
```diff
@@ -0,0 +1,42 @@
...
+						subtotal_excluding_tax {
+							...money
+						}
+						subtotal_including_tax {
+							...money
+						}
+						subtotal_with_discount_excluding_tax {
+							...money
+						}
+					}
```
griest024, line 35: I think its worth extracting the prices to a separate fragment since its now used in 3 places.

libs/cart/src/drivers/magento/queries/remove-coupon.ts
```diff
@@ -0,0 +1,42 @@
...
+						}
+						subtotal_excluding_tax {
+							...money
+						}
+						subtotal_including_tax {
+							...money
+						}
+						subtotal_with_discount_excluding_tax {
+							...money
+						}
```
griest024, line 34: Do we want to get `discounts` here in addition to the other prices? Not sure about it but we might need it for something.

libs/cart/src/drivers/magento/queries/remove-coupon.ts
lderrickable: true. done

libs/cart/src/drivers/magento/queries/remove-coupon.ts
lderrickable: I think this should be done in the cart-items fragment probably. Different PR if we need it though. It causes things to break and stuff

griest024: okay a different PR sounds good

PR #827, opened 2020-04-17
feat(cart): add place order reducers

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/src/reducers/cart-order-state.interface.ts
+++ b/libs/cart/src/reducers/cart-order-state.interface.ts
@@ -1,7 +1,7 @@
 import { DaffCartOrderResult } from '../models/cart-order-result';
 
 export interface DaffCartOrderReducerState<T extends DaffCartOrderResult = DaffCartOrderResult> {
-  order: T;
+  cartOrderResult: T;
   loading: boolean;
   errors: string[];
 }
--- a/libs/cart/src/reducers/cart-order-state.interface.ts
+++ b/libs/cart/src/reducers/cart-order-state.interface.ts
@@ -1,0 +1,5 @@
+export interface DaffCartOrderReducerState {
+  order: {id: string | number};
+  loading: boolean;
+  errors: string[];
+}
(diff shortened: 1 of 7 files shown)


libs/cart/src/reducers/cart-order-state.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface DaffCartOrderReducerState {
+  order: {id: string | number};
```
lderrickable, line 2: seems weird to make this an object, but not make it generic. I think we should go with one or the other. Probably make it generic to future proof it.

griest024: Making it generic would introduce a dependency onto checkout. Maybe thats a reason to move the place order driver into checkout instead.

griest024: changed

libs/cart/src/reducers/cart-order-state.interface.ts
```diff
@@ -0,0 +1,7 @@
+import { DaffCartOrderResult } from '../models/cart-order-result';
+
+export interface DaffCartOrderReducerState<T extends DaffCartOrderResult = DaffCartOrderResult> {
+  order: T;
```
lderrickable, line 4: I think this should be called something else to be more clear that it isn't actually an order. Maybe just cartOrderResult?

griest024: changed

libs/cart/src/reducers/cart-reducers-state.interface.ts
```diff
@@ -1,6 +1,8 @@
 import { DaffCartReducerState } from './cart-state.interface';
 import { DaffCart } from '../models/cart';
+import { DaffCartOrderReducerState } from './cart-order-state.interface';
 
 export interface DaffCartReducersState<T extends DaffCart = DaffCart> {
-  cart: DaffCartReducerState<T>
+  cart: DaffCartReducerState<T>,
+  order: DaffCartOrderReducerState
```
lderrickable, line 7: This should be generic

griest024: done

PR #840, opened 2020-04-22
feat(cart): add coupon reducer

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/src/reducers/cart-coupon/cart-coupon.reducer.ts
+++ b/libs/cart/src/reducers/cart-coupon/cart-coupon.reducer.ts
@@ -1,0 +1,80 @@
+import {
+  DaffCartCouponActionTypes,
+} from '../../actions/public_api';
+import { initialState } from '../cart-initial-state';
+import { DaffCartReducerState } from '../cart-state.interface';
+import { ActionTypes } from '../action-types.type';
+import { DaffCartErrorType } from '../cart-error-type.enum';
+import { DaffCart } from '../../models/cart';
+
+function addError<T extends DaffCart>(state: DaffCartReducerState<T>, error: string) {
+  return {
+    ...state,
+    errors: {
+      ...state.errors,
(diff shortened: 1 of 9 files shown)


libs/cart/src/reducers/cart-coupon/cart-coupon.reducer.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    case DaffCartCouponActionTypes.CartCouponApplySuccessAction:
+    case DaffCartCouponActionTypes.CartCouponRemoveSuccessAction:
+    case DaffCartCouponActionTypes.CartCouponRemoveAllSuccessAction:
+      return {
+        ...state,
+        ...resetErrors(state),
+        cart: {
+          ...state.cart,
+          ...action.payload
```
lderrickable, line 51: should this action return a `Partial<DaffCart>` instead of a `DaffCart`? Because won't this overwrite the fields not returned by the applyCoupon mutation? The applyCoupon mutation only requests a partial cart

griest024: yeah good point, they should be partials

lderrickable: That's my bad. When I was doing it originally, I thought that's how the other cart actions were done, but I must've misread them and didn't think it through this far

griest024: No its all good, I overlooked that. The other cart actions use partials so I should have known.

griest024: changed

PR #851, opened 2020-04-27
docs(all): add general guidelines for spec file import pathing

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/docs/internal/testing/testing.md
+++ b/docs/internal/testing/testing.md
@@ -31,7 +31,7 @@ The hidden problem here is that the `SomeImport` is not being accessed from the 
 
 To address this problem, we have elected to write our unit tests as if we are looking at our code from a slightly different perspective. We assume that all code that is NOT UNDER TEST comes from the public API of the package (unless the code in question is intentionally private).
 
-To achieve this we change the `@daffodil/library` `tsconfig.json`'s `compilerOptions.paths` to point to the public entrypoint of the package. As a result, our tests give us an additional guarantee: if `SomeImport` is meant to be public, we can guarantee that it is via while we're unit testing. This leads to lower maintenance overheads, and ensures better consumer-side guarantees.
+To achieve this we change the `@daffodil/library` `tsconfig.json`'s `compilerOptions.paths` to point to the public entrypoint of the package. As a result, our tests give us an additional guarantee: if `SomeImport` is meant to be public, we can guarantee that it is while we are unit testing. This leads to lower maintenance overheads, and ensures better consumer-side guarantees.
 
 ```ts
 import { SomeImport } from '@daffodil/library';
@@ -45,7 +45,7 @@ describe('My Library Function', () => {
 });
 ```
 
-### An unintentional benefit
+### Improved Import Path Readability
 Secondarily, this also makes package imports and maintenance much simpler. Instead of having to worry about silly path references like:
 
 ```ts
--- a/docs/internal/testing/testing.md
+++ b/docs/internal/testing/testing.md
@@ -1,16 +1,14 @@
 # Import Pathing Guidelines for Spec files
 
 ## The Problem
-One of the difficult problems that we deal with as library authors is defining the API that we expose to consuming application developers. This essentially means one of two things:
+One of the difficult problems that we deal with as library authors is defining the API that we expose to consuming application developers. We could handle this problem in one of two ways:
   
   1. We provide all the library exports that we use internally to consumers
-    * This is often a very bad idea for long term maintenance
   2. We provide only a subset of the libraries' exports.
-    *  We can allow our consumers to use the parts of the library we intend them to use, leading to a smaller API surface.
 
-Unfortunately, we (the authors of Daffodil) often view the world from the first perspective. We often write our internal tests depending upon `imports` that are only available to us as library authors. 
+The first scenario often is a very bad idea for long term maintenance, as the library API surface is very broad and leads to extensive backwards-compatibility maintenance, which is a massive time-sink. The second is more preferred as we can allow our consumers to use the parts of the library we intend them to use, leading to a smaller API surface.
 
-This means that the tests are from a very specific perspective, our internal view. So tests that pass for us won't necessarily pass for our consumers, as they may not have a necessary import available. Subsequently, this also means that code that works for us for example, may not work for consuming applications. 
+Unfortunately, unless we (the authors of the library) are very particular about preventing this problem, our code will often tend towards the first scenario. It would be wonderfully helpful if there were a mechanism to help us tend towards the narrow API in an automated and "by-default" way. Lucky for us, we can accomplish this by changing the perspective from which we write tests.
 
 ### An Example
 Consider the code in the file `my-library-function.spec.ts` in the fake package `@daffodil/library`:
@@ -27,7 +25,7 @@ describe('My Library Function', () => {
 });
 ```
 
-The issue at hand here is that the `SomeImport` is not being accessed from the public-facing api of `@daffodil/library`, so we aren't adequately testing a true consumer interaction. A consuming application developer could attempt to write this same code with this exact dependency set and the code wouldn't work, because `SomeImport` may not be publicly exposed.
+The issue at hand here is that the `SomeImport` is not being accessed from the public-facing api of `@daffodil/library`, so we aren't adequately testing a true consumer interaction. A consuming application developer could attempt to write this same code with this exact dependency set and the code wouldn't work if `SomeImport` isn't publicly exposed.
 
 ## The Solution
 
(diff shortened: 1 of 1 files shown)


docs/internal/testing/testing.md
```diff
@@ -0,0 +1,60 @@
...
+});
+```
+
+The hidden problem here is that the `SomeImport` is not being accessed from the public-facing api, so we aren't adequately testing a true consumer interaction. A consuming application developer could attempt to write this example same code, with this same dependency, and the code wouldn't work, because `SomeImport` may not be a publicly available token. 
+
+## The Solution
+
+To address this problem, we have elected to write our unit tests as if we are looking at our code from a slightly different perspective. We assume that all code that is NOT UNDER TEST comes from the public API of the package (unless the code in question is intentionally private).
+
+To achieve this we change the `@daffodil/library` `tsconfig.json`'s `compilerOptions.paths` to point to the public entrypoint of the package. As a result, our tests give us an additional guarantee: if `SomeImport` is meant to be public, we can guarantee that it is via while we're unit testing. This leads to lower maintenance overheads, and ensures better consumer-side guarantees.
```
griest024, line 34: `we can guarantee that it is via while we're unit testing`, `via while`. Should be just `while`.

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,60 @@
...
+import { myLibraryFunction } from './my-library-function';
+
+describe('My Library Function', () => {
+  it('should work', () => {
+    expect(myLibraryFunction(SomeImport)).toEqual("Something");
+  });
+});
+```
+
+### An unintentional benefit
```
griest024, line 48: I would name this section `Improved Import Path Readability` or something like that.

docs/internal/testing/testing.md
damienwebdev: Agreed, fixed.

docs/internal/testing/testing.md
damienwebdev: Agreed, fixed.

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,60 @@
...
+
+## The Problem
+One of the difficult problems that we deal with as library developers is defining the API that we expose to consuming application developers. This essentially means one of two things:
+  
+  1. We have to provide all the library exports to consumers
+    * This is often a very bad idea for long term maintenance
+  2. We have to provide only a subset of the libraries' exports so that we can allow our consumers to use the parts of the library we intend them to use.
+    * Smaller API surface means lower maintenance
+
+Unfortunately, when we, the authors of Daffodil, are writing tests we are often prone to depending upon `imports` that are only available to us as library authors. This means that the tests are from a very specific perspective, our internal view. So tests that pass for us, won't necessarily pass for our consumers because they may not have an import available. Subsequently, this also means that code that actually works for in `@daffodil/demo` for example, may not work for consuming applications.
```
lderrickable, line 11: > Subsequently, this also means that code that actually works for in `@daffodil/demo` for example, may not work for consuming applications. 
I don't think `@daffodil/demo` is a great example of this because we have never imported into demo directly from a lib that I know of. I think you should use `@daffodil/cart` or some other lib instead. I also think that using demo might be a little confusing for someone who isn't as familiar with daffodil as we are. In the first sentence you write "tests are often prone to depending upon imports that are only..." but the example you give is an application that depends directly upon a lib. Would probably make it so they would need to look at source code when they ask the question: "what is demo and why can't I import it?"

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,60 @@
...
+import { myLibraryFunction } from './my-library-function';
+
+describe('My Library Function', () => {
+  it('should work', () => {
+    expect(myLibraryFunction(SomeImport)).toEqual("Something");
+  });
+});
+```
+
+The hidden problem here is that the `SomeImport` is not being accessed from the public-facing api, so we aren't adequately testing a true consumer interaction. A consuming application developer could attempt to write this example same code, with this same dependency, and the code wouldn't work, because `SomeImport` may not be a publicly available token.
```
lderrickable, line 28: has a typo: "to write this example same code"

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,60 @@
...
+    expect(myLibraryFunction(SomeImport)).toEqual("Something");
+  });
+});
+```
+
+The hidden problem here is that the `SomeImport` is not being accessed from the public-facing api, so we aren't adequately testing a true consumer interaction. A consuming application developer could attempt to write this example same code, with this same dependency, and the code wouldn't work, because `SomeImport` may not be a publicly available token. 
+
+## The Solution
+
+To address this problem, we have elected to write our unit tests as if we are looking at our code from a slightly different perspective. We assume that all code that is NOT UNDER TEST comes from the public API of the package (unless the code in question is intentionally private).
```
lderrickable, line 32: I think this should be simplified to: To address this problem, we assume that all code that is NOT UNDER TEST comes from the public API of the package (unless the code in question is intentionally private).

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,60 @@
...
+
+To achieve this we change the `@daffodil/library` `tsconfig.json`'s `compilerOptions.paths` to point to the public entrypoint of the package. As a result, our tests give us an additional guarantee: if `SomeImport` is meant to be public, we can guarantee that it is while we are unit testing. This leads to lower maintenance overheads, and ensures better consumer-side guarantees.
```
lderrickable, line 34: "As a result, our tests give us an additional guarantee: if `SomeImport` is meant to be public, we can guarantee that it is while we are unit testing."

I think this should be simplified to: As a result of this change, we can guarantee that if `SomeImport` is meant to be public our unit tests will treat it as a publicly exported token.

Or something like that. I don't think it should be split up by the `:`.

damienwebdev:
@lderrickable @griest024 I adjusted the language slightly, take a look.

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,64 @@
...
+
+The issue at hand here is that the `SomeImport` is not being accessed from the public-facing api of `@daffodil/library`, so we aren't adequately testing a true consumer interaction. A consuming application developer could attempt to write this same code with this exact dependency set and the code wouldn't work, because `SomeImport` may not be publicly exposed.
+
+## The Solution
+
+To address this problem, we have decided to assume that all code that is **not under test** comes from the public API of the package (unless the code in question is intentionally private). 
+
+To achieve this we change the `@daffodil/library` `tsconfig.json`'s `compilerOptions.paths` to point to the public entrypoint of the package. 
+
+As a result, we can guarantee that if `SomeImport` is meant to be public, then our unit tests will only consume it as a public import. This leads to lower maintenance overheads, and ensures better consumer-side guarantees.
```
lderrickable, line 38: Why does this lead to lower maintenance overheads? I think you mean that we don't forget to expose things to the public api as often, but I don't think that's super obvious. I think you need to explain why it leads to lower maintenance overhead or remove "lower maintenance overheads".

lderrickable: Oh, you explain below. I think you should just remove "lower maintenance overheads" here, because you have a section about it later.

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,64 @@
+# Import Pathing Guidelines for Spec files
+
+## The Problem
+One of the difficult problems that we deal with as library authors is defining the API that we expose to consuming application developers. This essentially means one of two things:
```
lderrickable, line 4: To me, the mental transition between these two sentences and into the bullet points is jarring. The first sentence talks about something really concrete, a problem. The bullet points provide possible solutions to the problem. But the sentence in between is pretty vague and I don't think it glues the two together very well. I think "We could handle this problem in one of two ways:" flows better.

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,64 @@
...
+  
+  1. We provide all the library exports that we use internally to consumers
+    * This is often a very bad idea for long term maintenance
+  2. We provide only a subset of the libraries' exports.
+    *  We can allow our consumers to use the parts of the library we intend them to use, leading to a smaller API surface.
+
+Unfortunately, we (the authors of Daffodil) often view the world from the first perspective. We often write our internal tests depending upon `imports` that are only available to us as library authors. 
+
+This means that the tests are from a very specific perspective, our internal view. So tests that pass for us won't necessarily pass for our consumers, as they may not have a necessary import available. Subsequently, this also means that code that works for us for example, may not work for consuming applications.
```
lderrickable, line 13: Hmm...I don't know how else to say this. I really don't like lines 11-13. It feels like they're dancing around what they're trying to say instead of just saying it. And not the good kind of dancing. The bad kind.

docs/internal/testing/testing.md
```diff
@@ -0,0 +1,64 @@
...
+import { myLibraryFunction } from './my-library-function';
+
+describe('My Library Function', () => {
+  it('should work', () => {
+    expect(myLibraryFunction(SomeImport)).toEqual("Something");
+  });
+});
+```
+
+The issue at hand here is that the `SomeImport` is not being accessed from the public-facing api of `@daffodil/library`, so we aren't adequately testing a true consumer interaction. A consuming application developer could attempt to write this same code with this exact dependency set and the code wouldn't work, because `SomeImport` may not be publicly exposed.
```
lderrickable, line 30: "and the code wouldn't work, because `SomeImport` may not be publicly exposed." I think the relationship here should be stronger, like "but their code wouldn't work if `SomeImport` isn't publicly exposed."

PR #862, opened 2020-04-29
feat(cart): add Magento cart coupon driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
Adds a transformer specifically for the abridged cart coupon response so that the cart in state is not totally overwritten by a partial cart.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/src/drivers/magento/cart-coupon.service.spec.ts
+++ b/libs/cart/src/drivers/magento/cart-coupon.service.spec.ts
@@ -1,0 +1,390 @@
+import { TestBed } from '@angular/core/testing';
+import { ApolloTestingController, ApolloTestingModule } from 'apollo-angular/testing';
+import { catchError } from 'rxjs/operators';
+import { GraphQLError } from 'graphql';
+
+import {
+  DaffCart,
+  DaffCartCoupon,
+  MagentoCart
+} from '@daffodil/cart';
+import {
+  DaffCartFactory,
+  DaffCartCouponFactory,
+  MagentoCartFactory,
+} from '@daffodil/cart/testing';
+import { DaffBadInputError } from '@daffodil/driver';
+
+import { DaffMagentoCartCouponService } from './cart-coupon.service';
+import {
+  applyCoupon,
+  listCartCoupons,
+  removeAllCoupons
+} from './queries/public_api';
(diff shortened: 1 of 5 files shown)


libs/cart/src/drivers/magento/cart-coupon.service.spec.ts
```diff
@@ -0,0 +1,390 @@
...
+
+        const op = controller.expectOne(removeAllCoupons);
+
+        op.flush({
+          data: mockRemoveAllCouponsResponse
+        });
+      });
+    });
+
+    describe('when the call to the Magento API is unsuccessful', () => {
```
lderrickable, line 336: all of the groups of two tests for error can be just one test, because the transforming of the error should be tested in the transformCartMagentoError. Can just have one, "when the call to the Magento API fails, should throw an error".

lderrickable: actually, I guess I could kind of go either way with this. It's fine as is

libs/cart/src/drivers/magento/transforms/outputs/cart-coupon-response.service.ts
```diff
@@ -0,0 +1,84 @@
...
+import { DaffCart } from '../../../../models/cart';
+import { daffMagentoCouponTransform } from './cart-coupon';
+
+/**
+ * Transforms magento carts into an object usable by daffodil.
+ */
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffMagentoCartCouponResponseTransformer {
```
lderrickable, line 14: I think we've been trying to move toward these being pure functions rather than injectable services, since they no longer need/can be paired with an injection token. Could you change this?

griest024: @lderrickable I need to inject the cart item transformer. Is there a way to do this with a pure function?

lderrickable: oh right. No, I think needs to be done after the item transformer is a pure function too.

PR #875, opened 2020-05-06
feat(product): add query fragments for configurable products

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/product/src/drivers/magento/queries/fragments/configurable-product.ts
+++ b/libs/product/src/drivers/magento/queries/fragments/configurable-product.ts
@@ -1,0 +1,46 @@
+import gql from 'graphql-tag';
+
+export const magentoConfigurableProductFragment = gql`
+  fragment magentoConfigurableProduct on ConfigurableProduct {
+		items {
+			configurable_options {
+				attribute_code
+				attribute_id
+				id
+				label
+				position
+				product_id
+				values {
+					label
+					value_index
+					swatch_data {
+						value
+						thumbnail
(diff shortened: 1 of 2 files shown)


libs/product/src/drivers/magento/queries/fragments/configurable-product.ts
```diff
@@ -0,0 +1,45 @@
...
+					}
+				}
+			}
+			variants {
+				attributes {
+					code
+					label
+					value_index
+				}
+				product {
```
lderrickable, line 28: I made product only retrieve price and an image, because I think that's all that will be needed and I'm trying to cut down on the query size

damienwebdev: I think you need the `sku` for the subsequent add to cart call.

I also think you won't have enough info to meet the `DaffProduct` type guarantee. Maybe consider shifting the type on the DaffConfigurableProduct down to a Partial<DaffProduct> or a special type with explicit type guarantees.

lderrickable: Oh, I didn't know the sku worked that way. 

And yea, I realized this when I was working on the next PR. I made that change there, but I'll make it here since you bring it up

PR #881, opened 2020-05-11
feat(order): add reducers

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Adds a reducer for tracking loading and errors related to order operations
- Adds an entity reducer for storing and managing loaded orders

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/order/src/reducers/order-entities/order-entities-adapter.ts
+++ b/libs/order/src/reducers/order-entities/order-entities-adapter.ts
@@ -1,0 +1,12 @@
+import { EntityAdapter, createEntityAdapter } from '@ngrx/entity';
+
+import { DaffOrder } from '../../models/order';
+
+/**
+ * Order Adapter for changing/overwriting entity state.
+ */
+export const getOrderAdapter = (() => {
+  let cache;
+  return <T extends DaffOrder = DaffOrder>(): EntityAdapter<T> =>
+    cache = cache || createEntityAdapter<T>();
+})();
--- a/libs/order/src/reducers/order-entities/order-entities-adapter.ts
+++ b/libs/order/src/reducers/order-entities/order-entities-adapter.ts
@@ -5,7 +5,7 @@ import { DaffOrder } from '../../models/order';
 /**
  * Order Adapter for changing/overwriting entity state.
  */
-export const getOrderAdapter = (() => {
+export const daffGetOrderAdapter = (() => {
   let cache;
   return <T extends DaffOrder = DaffOrder>(): EntityAdapter<T> =>
     cache = cache || createEntityAdapter<T>();
--- a/libs/order/src/reducers/order-entities/order-entities.reducer.spec.ts
+++ b/libs/order/src/reducers/order-entities/order-entities.reducer.spec.ts
@@ -1,0 +1,62 @@
+import { DaffOrderFactory } from '@daffodil/order/testing';
+
+import {
+  DaffOrderLoadSuccess,
+  DaffOrderListSuccess,
+} from '../../actions/order.actions';
+import {
+  daffOrderEntitiesReducer as reducer,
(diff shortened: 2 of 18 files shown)


libs/order/src/reducers/order/order-reducer.interface.ts
```diff
@@ -0,0 +1,4 @@
+export interface DaffOrderReducerState {
```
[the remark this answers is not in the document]
griest024: The point of this state is just to keep track of the loading state and errors that arise from order list and load.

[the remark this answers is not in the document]
damienwebdev: This doesn't seem terribly useful at the moment, I'd just remove it.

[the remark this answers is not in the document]
griest024: Lets say a customer wants to view their recent orders. Knowing the loading state of these actions would allow the app dev to implement a ghost state or show a spinner to the customer. Its something that we have in pretty much all other state modules. I think we should keep it but I'll remove it if you really don't like it.

libs/order/src/reducers/order-entities/order-entities.reducer.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  describe('when OrderListSuccessAction is triggered', () => {
+    let result;
+
+    beforeEach(() => {
+      const orderListSuccess = new DaffOrderListSuccess([order]);
+
+      result = reducer(initialState, orderListSuccess);
+    });
+
+    it('should set countries from action.payload', () => {
```
lderrickable, line 58: copy paste error

libs/order/src/reducers/order-entities/order-entities-adapter.ts
```diff
@@ -0,0 +1,12 @@
+import { EntityAdapter, createEntityAdapter } from '@ngrx/entity';
+
+import { DaffOrder } from '../../models/order';
+
+/**
+ * Order Adapter for changing/overwriting entity state.
+ */
+export const getOrderAdapter = (() => {
```
lderrickable, line 8: should have a "daff" prefix

libs/order/src/reducers/order-store-feature-key.ts
```diff
@@ -0,0 +1 @@
+export const DAFF_ORDER_STORE_FEATURE_KEY = 'daffOrder';
```
lderrickable, line 1: what is the purpose of this key?

libs/order/src/reducers/order/order-reducer.interface.ts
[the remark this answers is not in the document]
lderrickable: I'm going to vote we keep it because we'll need it for the customer's profile in the future anyway.

libs/order/src/reducers/order-entities/order-entities.reducer.spec.ts
griest024: good catch, fixed

libs/order/src/reducers/order-entities/order-entities-adapter.ts
griest024: changed

libs/order/src/reducers/order-store-feature-key.ts
griest024: So that the key that the state is stored in doesn't have to be hardcoded, kinda the same rationale for using an enum instead of just hardcoded strings. It will be used in the feature state selector.

PR #884, opened 2020-05-13
feat(product): add configurable product actions and reducers

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
Configurable product actions are for users selecting an attribute of a configurable product, like the color blue for a T-shirt. The selectors, which are the complicated part, will be handled in another PR.

### DIFF
--- a/libs/product/src/actions/configurable-product.actions.ts
+++ b/libs/product/src/actions/configurable-product.actions.ts
@@ -37,7 +37,9 @@ export class DaffConfigurableProductRemoveAttribute<T extends DaffConfigurablePr
 }
 
 /**
- * Toggles an attribute of a particular configurable product.
+ * Toggles an attribute of a particular configurable product. If the attribute type of the configurable product already has 
+ * a different value than the one provided in the action, the attribute value in state will be overwritten by the value provided 
+ * in the action.
  * 
  * @param id - Id of the Configurable Product
  * @param attributeId - Id of the attribute to be toggled
--- a/libs/product/src/actions/configurable-product.actions.ts
+++ b/libs/product/src/actions/configurable-product.actions.ts
@@ -1,0 +1,55 @@
+import { Action } from '@ngrx/store';
+
+import { DaffConfigurableProduct } from '../models/configurable-product';
+
+/**
+ * Action types for Configurable Product Actions.
+ */
+export enum DaffConfigurableProductActionTypes {
+	ConfigurableProductApplyAttributeAction = '[Configurable Product] Configurable Product Apply Attribute Action',
+	ConfigurableProductRemoveAttributeAction = '[Configurable Product] Configurable Product Remove Attribute Action',
+	ConfigurableProductToggleAttributeAction = '[Configurable Product] Configurable Product Toggle Attribute Action',
+}
+
+/**
+ * Applies an attribute to a particular configurable product.
+ * 
+ * @param id - Id of the Configurable Product
+ * @param attributeId - Id of the attribute to be applied
+ * @param attributeValue - Value of the attribute to be applied
+ */
+export class DaffConfigurableProductApplyAttribute<T extends DaffConfigurableProduct> implements Action {
+  readonly type = DaffConfigurableProductActionTypes.ConfigurableProductApplyAttributeAction;
+
+  constructor(public id: T['id'], public attributeId: string, public attributeValue: string) {}
+}
+
+/**
+ * Removes an applied attribute from a particular configurable product.
+ * 
+ * @param id - Id of the Configurable Product
+ * @param attributeId - Id of the attribute to be removed
+ */
+export class DaffConfigurableProductRemoveAttribute<T extends DaffConfigurableProduct> implements Action {
+  readonly type = DaffConfigurableProductActionTypes.ConfigurableProductRemoveAttributeAction;
+
+  constructor(public id: T['id'], public attributeId: string) {}
+}
+
+/**
+ * Toggles an attribute of a particular configurable product.
+ * 
+ * @param id - Id of the Configurable Product
+ * @param attributeId - Id of the attribute to be toggled
+ * @param attributeValue - Value of the attribute to be toggled
+ */
+export class DaffConfigurableProductToggleAttribute<T extends DaffConfigurableProduct> implements Action {
+  readonly type = DaffConfigurableProductActionTypes.ConfigurableProductToggleAttributeAction;
+
+  constructor(public id: T['id'], public attributeId: string, public attributeValue: string) {}
+}
+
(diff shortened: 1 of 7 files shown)


libs/product/src/reducers/configurable-product-entities/configurable-product-entities.reducer.ts
```diff
@@ -0,0 +1,83 @@
...
+ */
+export function daffConfigurableProductEntitiesReducer<T extends DaffProduct>(
+  state = daffConfigurableProductAppliedAttributesEntitiesAdapter().getInitialState(), 
+  action: DaffProductGridActions<T> | DaffBestSellersActions<T> | DaffProductActions<T> | DaffConfigurableProductActions): EntityState<DaffProductVariantAttributesDictionary> {
+	const adapter = daffConfigurableProductAppliedAttributesEntitiesAdapter();
+  switch (action.type) {
+    case DaffProductGridActionTypes.ProductGridLoadSuccessAction:
+			return adapter.upsertMany(buildConfigurableProductAppliedAttributesEntities(action.payload), state);
+    case DaffBestSellersActionTypes.BestSellersLoadSuccessAction:
+      return adapter.upsertMany(buildConfigurableProductAppliedAttributesEntities(action.payload), state);
```
griest024, line 26: it seems that these 2 cases can be combined.

libs/product/src/reducers/configurable-product-entities/configurable-product-entities-reducer-adapter.ts
```diff
@@ -0,0 +1,12 @@
...
+
+import { DaffProductVariantAttributesDictionary } from '../../models/configurable-product';
+
+/**
+ * Configurable Product Applied Attributes Adapter for changing/overwriting entity state.
+ */
+export const daffConfigurableProductAppliedAttributesEntitiesAdapter = (() => {
+	let cache;
+  return (): EntityAdapter<DaffProductVariantAttributesDictionary> =>
+    cache = cache || createEntityAdapter<DaffProductVariantAttributesDictionary>();
```
griest024, line 11: Does this need to be generic?

libs/product/src/reducers/configurable-product-entities/configurable-product-entities.reducer.ts
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Reducer function that catches actions and changes/overwrites product entities state.
+ * 
+ * @param state current State of the redux store
+ * @param action ProductGrid, BestSellers, or Product actions
+ * @returns Product entities state
+ */
+export function daffConfigurableProductEntitiesReducer<T extends DaffProduct>(
+  state = daffConfigurableProductAppliedAttributesEntitiesAdapter().getInitialState(), 
+  action: DaffProductGridActions<T> | DaffBestSellersActions<T> | DaffProductActions<T> | DaffConfigurableProductActions): EntityState<DaffProductVariantAttributesDictionary> {
```
griest024, line 20: I think you need to add a `DaffConfigurableProduct` generic and pass it to `DaffConfigurableProductActions`.

libs/product/src/reducers/configurable-product-entities/configurable-product-entities-reducer-adapter.ts
lderrickable: I don't think so. These "entities" aren't really entities. They're really just a way to keep track of a bunch of on/off flags for each configurable product, and it wouldn't make sense to add to the model.

libs/product/src/reducers/configurable-product-entities/configurable-product-entities.reducer.ts
lderrickable: fixed

libs/product/src/reducers/configurable-product-entities/configurable-product-entities.reducer.ts
lderrickable: done

libs/product/src/reducers/configurable-product-entities/configurable-product-entities-reducer-adapter.ts
griest024: Well the attribute values have significance beyond on/off, right? Like one of these entites might look like
```ts
{
  color: 'blue',
  size: 'M'
}
```

I agree that it wouldn't really be that useful to customize the model though so a generic would not be appropriate here.

libs/product/src/actions/configurable-product.actions.ts
```diff
@@ -0,0 +1,55 @@
...
+ * 
+ * @param id - Id of the Configurable Product
+ * @param attributeId - Id of the attribute to be toggled
+ * @param attributeValue - Value of the attribute to be toggled
+ */
+export class DaffConfigurableProductToggleAttribute<T extends DaffConfigurableProduct> implements Action {
+  readonly type = DaffConfigurableProductActionTypes.ConfigurableProductToggleAttributeAction;
+
+  constructor(public id: T['id'], public attributeId: string, public attributeValue: string) {}
+}
```
griest024, line 50: I'm not sure that a toggle action makes sense for values that aren't boolean. LMK what you think; I might be misunderstanding something.

lderrickable: oh, you're right. I'll take another look at it

lderrickable: I changed how the reducer handles the action. Basically, I still want to allow for a toggle-type action, so that the app dev can just use that instead of the apply/remove actions. But I hadn't thought of the usecase (which would happen all the time) for when the color blue was selected and the new selection is the color red. So now the app dev can just use the toggle action for any time the user clicks on a new configurable attribute. So to be clear on the use cases:
1. Configurable T-shirt has the attribute color set to blue. User clicks on the blue swatch again to deselect the attribute color. 
2. Configurable T-shirt has no attribute color selected. User clicks on the blue swatch to select the attribute color.
3. Configurable T-shirt has the attribute color set to blue. User clicks on the red swatch to change their selection of the color attribute to red. (this is what previously wouldn't have worked)

libs/product/src/reducers/configurable-product-entities/configurable-product-entities-reducer-adapter.ts
lderrickable: yea, that's right

libs/product/src/actions/configurable-product.actions.ts
griest024: Alright lgtm. I'm convinced that this is a use case that we want to support.

PR #890, opened 2020-05-14
test(cart): improve in-memory backend coupon spec

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There is a race condition in the setup of one of the in-memory backend coupon tests.

## What is the new behavior?
The race condition is no longer present.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/testing/src/in-memory-backend/cart-root.service.integration.spec.ts
+++ b/libs/cart/testing/src/in-memory-backend/cart-root.service.integration.spec.ts
@@ -218,19 +218,19 @@ describe('DaffInMemoryBackendCartRootService | Integration', () => {
 
   describe('processing an apply coupon request', () => {
     let result;
+    let newCoupon: DaffCartCoupon;
 
     beforeEach(done => {
-      mockCart.coupons = [];
-      httpClient.post<any>('commands/resetDb', {carts: [mockCart]});
+      newCoupon = cartCouponFactory.create();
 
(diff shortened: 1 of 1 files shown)


lderrickable:
does this mean that sometimes the test would fail and sometimes it wouldn't?

griest024:
@lderrickable yeah, I never actually noticed it failing but it theorectically could run the expectation before the result was initialized.

libs/cart/testing/src/in-memory-backend/cart-root.service.integration.spec.ts
```diff
@@ -221,7 +221,7 @@ describe('DaffInMemoryBackendCartRootService | Integration', () => {
 
     beforeEach(done => {
       mockCart.coupons = [];
-      httpClient.post<any>('commands/resetDb', {carts: [mockCart]}).subscribe(() => done());
+      httpClient.post<any>('commands/resetDb', {carts: [mockCart]});
```
lderrickable, line 224: I still can't tell for sure if the test is really accomplishing what it's supposed to. Maybe it would make more sense to try adding a new coupon instead of removing and adding the old one, because it isn't easy to verify that the removal actually happens the way you'd expect?

griest024: fair enough, I have changed that spec to add a brand new coupon. I also changed another expectation to be more explicit.

PR #900, opened 2020-05-18
feat(cart): add selector for cart item discounted total

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
A selector was added for getting the discounted total of a cart item.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/cart/src/facades/cart/cart.facade.ts
+++ b/libs/cart/src/facades/cart/cart.facade.ts
@@ -50,6 +50,8 @@ export class DaffCartFacade<
 	orderResult$: Observable<V>;
 	orderResultId$: Observable<V['id']>;
 
+	private selectCartItemDiscountedTotal = getDaffCartSelectors<T, V>().selectCartItemDiscountedRowTotal;
+
   constructor(private store: Store<DaffCartReducersState<T, V>>) {
 		const {
 			selectCartLoading,
@@ -116,7 +118,11 @@ export class DaffCartFacade<
     this.orderResultErrors$ = this.store.pipe(select(selectCartOrderErrors));
     this.orderResult$ = this.store.pipe(select(selectCartOrderValue));
     this.orderResultId$ = this.store.pipe(select(selectCartOrderId));
-  }
+	}
+	
+	getCartItemDiscountedTotal(itemId: string | number): Observable<number> {
+		return this.store.pipe(select(this.selectCartItemDiscountedTotal, { id: itemId }));
+	}
 
   dispatch(action: Action) {
     this.store.dispatch(action);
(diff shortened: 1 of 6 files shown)


libs/cart/src/facades/cart/cart.facade.ts
```diff
@@ -50,6 +50,8 @@ export class DaffCartFacade<
 	orderResult$: Observable<V>;
 	orderResultId$: Observable<V['id']>;
 
+	private selectCartItemDiscountedTotal = getDaffCartSelectors<T, V>().selectCartItemDiscountedRowTotal;
```
griest024, line 53: I would initialize this field in the constructor since we're already calling `getDaffCartSelectors` in there.

libs/cart/src/selectors/cart/cart.selector.ts
```diff
@@ -152,6 +155,17 @@ const createCartSelectors = <
...
   );
+	const selectCartItemDiscountedRowTotal = createSelector(
+		selectCartItems,
+		(cartItems: DaffCartItem[], props) => {
+			return cartItems.reduce((acc: number, item: DaffCartItem) => {
+				if(item.item_id === props.id) {
+					return item.row_total - item.total_discount
+				}
+				return acc;
+			}, null)
```
griest024, line 166: This shouldn't be implemented with `reduce`. I would just find the cart item in the array and do the calculation.

libs/cart/src/facades/cart/cart.facade.ts
lderrickable: I can't reference it outside of the constructor if I do that

libs/cart/src/selectors/cart/cart.selector.ts
lderrickable: > This shouldn't be implemented with reduce.
why?

griest024: Its not a reduce operation. It doesn't depend on all the elements of the array, only one. There's no point iterating through every array element when you might find the element you need at the very beginning.

libs/cart/src/facades/cart/cart.facade.ts
griest024: hm that doesn't sound right...the only change is where you initialize the value of the field, not the field declaration itself. It seems to work for me, doing `this.selectCartItemDiscountedTotal = selectCartItemDiscountedRowTotal` in the constructor.

libs/cart/src/selectors/cart/cart.selector.ts
lderrickable: makes sense. changed

libs/cart/src/facades/cart/cart.facade.ts
lderrickable: like that?

griest024: yeah that looks good

PR #931, opened 2020-06-04
feat(geography): track the fully loaded countries in state

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Adds state for tracking the fully loaded countries
- Adds selectors for checking if a country is fully loaded
- Adds some generic default values

```
[ ] Yes
[x] No
```

## Other information
This is needed because a country's subdivision should be not a required value only when the country has been fully loaded _and_ no subdivions are present. Otherwise there is no way to determine whether or not a country has subdivisions without actually trying to load them all.

### DIFF
--- a/libs/geography/src/facades/geography/geography.facade.spec.ts
+++ b/libs/geography/src/facades/geography/geography.facade.spec.ts
@@ -17,8 +17,8 @@ import { DaffCountryFactory, DaffSubdivisionFactory } from '@daffodil/geography/
 import { DaffGeographyFacade } from './geography.facade';
 
 describe('DaffGeographyFacade', () => {
-  let store: MockStore<{ [DAFF_GEOGRAPHY_STORE_FEATURE_KEY]: Partial<DaffGeographyFeatureState<DaffCountry>> }>;
-  let facade: DaffGeographyFacade<DaffCountry>;
+  let store: MockStore<{ [DAFF_GEOGRAPHY_STORE_FEATURE_KEY]: Partial<DaffGeographyFeatureState> }>;
+  let facade: DaffGeographyFacade;
   let countryFactory: DaffCountryFactory;
   let subdivisionFactory: DaffSubdivisionFactory;
 
@@ -96,9 +96,10 @@ describe('DaffGeographyFacade', () => {
     });
 
     it('should be the countries upon a successful load', () => {
-      const expected = cold('a', { a: [mockCountry] });
       store.dispatch(new DaffCountryLoadSuccess(mockCountry));
-      expect(facade.countries$).toBeObservable(expected);
+      facade.countries$.subscribe(countries => {
+        expect(countries).toEqual([jasmine.objectContaining(mockCountry)])
+      });
     });
   });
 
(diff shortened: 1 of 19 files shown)


griest024:
@damienwebdev You mean store that info as a subfield of `DaffCountry`? So set it in the entities reducer?

damienwebdev:
@griest024 Yes. I think that makes more sense the storing a map of props elsewhere and having to go search for the prop in another piece of the state tree.

griest024:
@damienwebdev changed

libs/geography/src/facades/geography/geography.facade.spec.ts
```diff
@@ -17,8 +17,8 @@ import { DaffCountryFactory, DaffSubdivisionFactory } from '@daffodil/geography/
 import { DaffGeographyFacade } from './geography.facade';
 
 describe('DaffGeographyFacade', () => {
```
lderrickable, line 19: These tests should continue to use jasmine marbles instead of subscribe just to avoid possible subscription memory leaks. You wouldn't think it would happen but I've found testing with rxjs and angular to be a little unpredictable in terms of what carries over between tests.

libs/geography/src/reducers/country-entities/country-entities.reducer.ts
```diff
@@ -7,15 +7,18 @@ import { daffCountryEntitiesInitialState } from './country-entities-initial-stat
...
   switch (action.type) {
     case DaffGeographyActionTypes.CountryLoadSuccessAction:
-      return adapter.upsertOne(action.payload, state);
+      return adapter.upsertOne({
+        ...action.payload,
+        loaded: true
+      }, state);
 
     case DaffGeographyActionTypes.CountryListSuccessAction:
       return adapter.upsertMany(action.payload, state);
```
lderrickable, line 24: I think it might be a little more accurate to do something like:
```
case DaffGeographyActionTypes.CountryListSuccessAction:
      return adapter.upsertMany(action.payload.map(country => (
          { ...country, loaded: false}
      )), state);
```
Just so the `loaded` field isn't true or null. It'll be true or false.

libs/geography/src/reducers/geography/geography.reducer.ts
```diff
@@ -13,6 +13,12 @@ export function daffGeographyReducer<T extends DaffCountry>(
       return { ...state, loading: true };
 
     case DaffGeographyActionTypes.CountryLoadSuccessAction:
+      return {
```
lderrickable, line 16: It looks like this change is unneeded

libs/geography/src/facades/geography/geography.facade.spec.ts
griest024: I was using it so I could take advantage of different matchers like `toBeFalsy()` but with those new changes you requested, that shouldn't be necessary.

libs/geography/src/reducers/geography/geography.reducer.ts
griest024: yeah it was leftover from a previous implementation, good catch. removed.

libs/geography/src/reducers/country-entities/country-entities.reducer.ts
griest024: changed

libs/geography/src/facades/geography/geography.facade.spec.ts
griest024: changed

libs/geography/src/selectors/country-entities.selector.ts
```diff
@@ -67,7 +76,11 @@ const createCountryEntitySelectors = <T extends DaffCountry>() => {
     /**
      * Selector for a specific country's subdivisions.
      */
-    selectCountrySubdivisions
+    selectCountrySubdivisions,
+    /**
+     * Selector for checking if a country has been fully loaded.
```
lderrickable, line 81: I think you should mention subdivisions in this comment, since loading the subdivisions is what it means to be "fully loaded"

libs/geography/src/selectors/geography.selector.ts
```diff
@@ -10,9 +10,10 @@ export interface DaffGeographySelectors {
   selectGeographyState: MemoizedSelector<object, DaffGeographyReducerState>;
   selectGeographyLoading: MemoizedSelector<object, boolean>;
   selectGeographyErrors: MemoizedSelector<object, string[]>;
+
```
lderrickable, line 13: remove blank space

libs/geography/src/reducers/country-entities/country-entities.reducer.ts
griest024: actually I'm not sure if this is a good idea. If the country has already been fully loaded, then a list success would overwrite this. I think it should be modified to
```ts
return adapter.upsertMany(action.payload.map(country => (
          { ...country, loaded: country.loaded || false}
      )), state);
```

LMK what you think

griest024: Actually the problem is broader, any listed countries will overwrite the subdivisions of already loaded countries. I'm going to rework this reducer case.

libs/geography/src/selectors/country-entities.selector.ts
griest024: added

libs/geography/src/selectors/geography.selector.ts
griest024: removed

libs/geography/src/reducers/country-entities/country-entities.reducer.ts
griest024: Okay I just pushed a new implementation that doesn't overwrite fully loaded countries.

libs/geography/src/reducers/country-entities/country-entities.reducer.spec.ts
```diff
@@ -54,13 +58,35 @@ describe('Geography | Reducer | CountryEntities', () => {
...
 
       result = reducer(initialState, countryListSuccess);
     });
 
     it('should set countries from action.payload', () => {
-      expect(result.entities).toEqual({[countryId]: jasmine.objectContaining(country)})
+      expect(result.entities).toEqual({[countryId]: jasmine.objectContaining(mockCountry)})
+    });
+
+    describe('when a country has previously been fully loaded', () => {
```
lderrickable, line 70: I think you should add one more test for the case where the list request gives subdivisions

griest024: done

lderrickable:
lgtm

PR #935, opened 2020-06-08
feat(cart, geography): deprecate region_id and transform region to string for cart address input

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Deprecates `DaffAddress#region_id`
- Changes the Magento cart address input transformer to transform the region to a string.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/geography/src/models/address.ts
+++ b/libs/geography/src/models/address.ts
@@ -5,7 +5,12 @@ export interface DaffAddress {
 	street: string;
 	street2?: string;
   city: string;
-  region: string;
+  region: string | number;
+  /**
+   * Use DaffAddress#region instead.
+   *
+   * @deprecated
+   */
   region_id?: string | number;
   country?: string;
   country_id?: string | number;
--- a/libs/cart/src/drivers/magento/transforms/inputs/cart-address.service.spec.ts
+++ b/libs/cart/src/drivers/magento/transforms/inputs/cart-address.service.spec.ts
@@ -36,15 +36,18 @@ describe('Driver | Magento | Cart | Transformer | MagentoCartAddress', () => {
     let street;
     let city;
     let firstname;
+    let region;
 
     beforeEach(() => {
       street = 'street';
       city = 'city';
       firstname = 'firstname';
+      region = 'region';
 
       mockDaffCartAddress.street = street;
       mockDaffCartAddress.city = city;
       mockDaffCartAddress.firstname = firstname;
+      mockDaffCartAddress.region = region;
 
       transformedCartAddress = service.transform(mockDaffCartAddress);
     });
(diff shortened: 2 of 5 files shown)


libs/geography/src/models/address.ts
```diff
@@ -1,13 +1,40 @@
...
+   * Required if region_id is not set.
+   */
+  region?: string;
+}
+
+/**
+ * A model representing an address in a country with unknown regions.
+ * The region here should be user-supplied.
+ */
+export interface DaffUnknownRegionAddress extends Omit<DaffAddress, 'region_id'> {
```
lderrickable, line 30: shouldn't you omit the `region_id` and the `region` for the UnknownRegionAddress? Then keep both for the KnownRegionAddress?

libs/geography/src/models/address.ts
```diff
@@ -1,13 +1,40 @@
...
+  country_id: string | number;
+  /**
+   * A field for a region_id supplied by the platform.
+   * Needed for countries that have a list of known regions.
+   * Required if region is not set.
+   */
+  region_id?: string | number;
+  /**
+   * A field for user-supplied regions.
+   * Needed for countries that don't have any regions supplied by the platform.
```
lderrickable, line 20: I think this comment could be a little more precise. "Needed for countries that have regions not supplied by the platform."

libs/geography/src/models/address.ts
```diff
@@ -1,13 +1,40 @@
...
+ */
+export interface DaffUnknownRegionAddress extends Omit<DaffAddress, 'region_id'> {
+  region: string;
+}
+
+/**
+ * A model representing an address in a country with a known set of regions with unique IDs.
+ * The region_id should be set to one of those IDs.
+ */
+export interface DaffKnownRegionAddress extends Omit<DaffAddress, 'region'> {
```
griest024, line 38: @lderrickable if you have suggestions for better names for these models, they are gladly welcome. Damien and I couldn't think of anything better.

lderrickable, line 38: I think maybe region should still remain in the case that there are some regions defined by the platform, but not all regions are defined.

griest024: @damienwebdev feel free to chime in but I think thats not a scenario that can happen.

libs/geography/src/models/address.ts
griest024: ah nice, that is better. Changed

libs/geography/src/models/address.ts
lderrickable: Ok, I'm too concerned about it. It would be an obscure and easily fixable bug if it ended up being the case in the future.

lderrickable: another question. Can't we just treat the region as the region_id whenever the region_id isn't given by the platform?

lderrickable: That way we wouldn't need the region field, and when the user needs to supply a region, we would just write into the region_id field. It would also simplify a lot of this

griest024: uh haha yes that would be way simpler. But if we do that I vote that we use the `region` field for everything instead of `region_id`.

@damienwebdev Thoughts?

PR #946, opened 2020-06-14
feat(cart): add testing drivers

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[x] No
```

## Other information


lderrickable:
I think the tests should use helper methods like `isCart` to verify that the object is actually a cart. Or you could do:
```
let expectedValue = new Factory().create();
let stubFactory = jasmine.createSpyObject('factory', ['create']);
...
{ provide: "Factory", useValue: stubFactory }
...
stubFactory.create.and.returnValue(expectedValue);
...
const expected = cold('(a|)', {a: expectedValue});

expect(service.get(id)).toBeObservable(expected);
```

libs/cart/testing/src/drivers/testing/cart-shipping-information/cart-shipping-information.service.ts
```diff
@@ -0,0 +1,32 @@
...
+  DaffCartShippingInformationServiceInterface,
+  DaffCartShippingRate
+} from '@daffodil/cart';
+
+import { DaffCartFactory, DaffCartShippingRateFactory } from '../../../factories/public_api';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffTestingCartShippingInformationService implements DaffCartShippingInformationServiceInterface<DaffCartShippingRate, DaffCart> {
```
lderrickable, line 15: You could remove the type definitions for the Interface here and elsewhere, since they all have default values now.

griest024: changed

griest024:
@lderrickable how would I use the helper functions with jasmine marbles? I think it might be better to just mock the factories.

lderrickable:
@griest024 Oh, yea you're right. You'd need to mock the factories. I think that's better than having a describe that says "should return a cart" when all that is tested is that it is an Object type.

lderrickable:
Imo, I don't think we even really need tests for these files, because all the typing should be enforced by the interfaces

griest024:
@lderrickable I played around with trying to test it like `jasmine.any(MockDaffCart)` but that doesn't work because of the way that the factory spreads in the partials.

The tests are a bit superfluous but we gotta have that test coverage 👀

PR #950, opened 2020-06-17
feat(product): add facade fields for configurable product discounts

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
This adds selectors and facade fields for minimum and maximum discounted prices for configurable products.

### DIFF
--- a/libs/product/src/facades/configurable-product/configurable-product-facade.interface.ts
+++ b/libs/product/src/facades/configurable-product/configurable-product-facade.interface.ts
@@ -32,12 +32,6 @@ export interface DaffConfigurableProductFacadeInterface extends DaffStoreFacade<
 	getMaximumPrice(id: string): Observable<number>;
 
 	/**
-	 * Get the current maximum price possible based on the applied attributes and remaining variants.
-	 * @param id the id of the configurable product.
-	 */
-	getMaximumPrice(id: string): Observable<number>;
-
-	/**
 	 * Get the current minimum discounted price possible based on the applied attributes and remaining variants.
 	 * @param id the id of the configurable product.
 	 */
--- a/libs/product/src/facades/configurable-product/configurable-product-facade.interface.ts
+++ b/libs/product/src/facades/configurable-product/configurable-product-facade.interface.ts
@@ -32,10 +32,34 @@ export interface DaffConfigurableProductFacadeInterface extends DaffStoreFacade<
 	getMaximumPrice(id: string): Observable<number>;
 
 	/**
+	 * Get the current maximum price possible based on the applied attributes and remaining variants.
+	 * @param id the id of the configurable product.
+	 */
+	getMaximumPrice(id: string): Observable<number>;
+
+	/**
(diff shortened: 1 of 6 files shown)


libs/product/src/selectors/configurable-product/configurable-product.selectors.ts
```diff
@@ -57,6 +61,31 @@ const createConfigurableProductSelectors = (): DaffConfigurableProductMemoizedSe
...
+	/**
+	 * Selector that determines whether any of the variants for the current configuration of the configurable product has a discount.
+	 */
+	const selectConfigurableProductHasDiscount = createSelector(
+		selectProductEntities,
+		selectConfigurableProductAppliedAttributesEntitiesState,
+		(products, appliedAttributesEntities, props) => {
+			return selectMatchingConfigurableProductVariants.projector(products, appliedAttributesEntities, { id: props.id })
+				.filter(variant => variant.discount ? variant.discount.amount > 0 : false)
+				.length > 0;
```
griest024, line 85: This should be implemented with `reduce`.

libs/product/src/facades/configurable-product/configurable-product-facade.interface.ts
```diff
@@ -31,12 +31,36 @@ export interface DaffConfigurableProductFacadeInterface extends DaffStoreFacade<
 	 */
 	getMaximumPrice(id: string): Observable<number>;
 
+	/**
+	 * Get the current maximum price possible based on the applied attributes and remaining variants.
+	 * @param id the id of the configurable product.
+	 */
+	getMaximumPrice(id: string): Observable<number>;
```
griest024, line 38: This duplicates the above interface key.

libs/product/src/selectors/configurable-product/configurable-product.selectors.ts
lderrickable: why? this is shorter, right?

libs/product/src/facades/configurable-product/configurable-product-facade.interface.ts
lderrickable: oops. removed

libs/product/src/selectors/configurable-product/configurable-product.selectors.ts
griest024: You're computing a single value that depends on every value in a list. That is a reduce operation so I think its more clear to express it as such.

I'm not sure which implementation would actually contain fewer characters, but a `reduce` implementation is just a single operation instead of 2: `filter` and then a comparison on `length`.

```ts
reduce((variant, acc) => acc || (variant.discount && variant.discount.amount > 0), false)
```

lderrickable: ok done

PR #954, opened 2020-06-23
feat(cart): add cart address update state

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
Adds actions for updating both the shipping and the billing address at once. This is useful because it difficult to deal with checking multiple success actions when dispatching the update actions separately.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/src/actions/cart-address.actions.ts
+++ b/libs/cart/src/actions/cart-address.actions.ts
@@ -1,0 +1,45 @@
+import { Action } from '@ngrx/store';
+
+import { DaffCart } from '../models/cart';
+import { DaffCartAddress } from '../models/cart-address';
+
+export enum DaffCartAddressActionTypes {
+  CartAddressUpdateAction = '[DaffCart] Cart Address Update Action',
+  CartAddressUpdateSuccessAction = '[DaffCart] Cart Address Update Success Action',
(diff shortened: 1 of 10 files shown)


lderrickable:
I think that the loading and errors will be a little bit user unfriendly. It makes sense that you have a different set of effects for the CartAddressUpdate, but I think it might make more sense not to have CartAddress reducers/selectors/state. I think since it is basically the same feature area, it's ok to have the Billing/ShippingAddress reducers listen for the CartAddressUpdate actions. Since the billing address and shipping address are always the same in this case, you can just directly set the address too through the Billing/ShippingAddress reducers. This way, the app dev doesn't need to worry about accessing different states for errors or loading. Like, imagine now, needing access to the Billing/Shipping/Address loading states all in the same shipping form. The only downside that I see in my proposal is that if one of the update driver calls fail, it would put an error in both Billing and Shipping address states. But I think that's ok, since we can customize the error message to just say "There was an error in updating the Billing/shipping address" and the solo shipping address update failure action could pass a more specific error message. What do you think?

griest024:
@lderrickable okay yeah I think thats better,

griest024:
@lderrickable done

lderrickable:
lgtm

PR #960, opened 2020-06-26
feat(core): add substream operator

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
Adds the substream operator for listening for a sequence of multiple actions.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/core/src/operators/substream.pipe.ts
+++ b/libs/core/src/operators/substream.pipe.ts
@@ -1,56 +1,6 @@
 import { Action } from '@ngrx/store';
-import { pipe, Observable, MonoTypeOperatorFunction, Operator, TeardownLogic, Subscriber } from 'rxjs';
-import { scan } from 'rxjs/operators';
-
-class SkipWhileSubscriber<T> extends Subscriber<T> {
-  private skipping = true;
-  private index = 0;
-
-  constructor(
-    destination: Subscriber<T>,
-    private predicate: (value: T, index: number) => boolean
-  ) {
-    super(destination);
-  }
-
-  protected _next(value: T): void {
-    const destination = this.destination;
-
-    // call the predicate every time
-    // this is the only change from the vanilla rxjs operator
-    this.tryCallPredicate(value);
-
-    if (!this.skipping) {
-      destination.next(value);
-    }
-  }
-
-  private tryCallPredicate(value: T): void {
-    try {
-      const result = this.predicate(value, this.index++);
-      this.skipping = Boolean(result);
-    } catch (err) {
-      this.destination.error(err);
-    }
-  }
-}
-
-class SkipWhileOperator<T> implements Operator<T, T> {
-  constructor(private predicate: (value: T, index: number) => boolean) {
-  }
-
-  call(subscriber: Subscriber<T>, source: any): TeardownLogic {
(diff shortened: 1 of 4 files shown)


lderrickable:
This looks really good to me. It's actually a bit different than how I was imagining it, but I don't think it could've been a pure function the way I was picturing it. Nice work!

libs/core/src/operators/substream.pipe.ts
```diff
@@ -0,0 +1,106 @@
...
+export type ActionType = Action['type'];
+export type ActionSequenceStep = ActionType | ActionType[];
+export type ActionSequence = ActionSequenceStep[];
+
+const compareActionSequenceStep = (step: ActionSequenceStep, action: Action): boolean => Array.isArray(step)
+  ? step.reduce((acc, type) => acc || action.type === type, false)
+  : action.type === step
+
+/**
+ * Finds and emits a substream of actions.
```
lderrickable, line 64: I think this should be something like: "Watches for a particular stream of actions, and emits a list of those actions once all actions have occurred"

libs/core/src/operators/substream.pipe.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * You can optionally pass a list of terminators.
+ * If a terminator is encountered in the action stream, any partially matched substream is reset.
+ * For example, if substream([A, B], C), an action stream of [A1, C1, B1] would cause substream to not emit.
+ *
+ * @param sequence The sequence of action types that define the substream to listen for.
+ * @param terminators A list of terminators, any of which interrupt and reset a partially matched substream.
+ */
+export const substream = (sequence: ActionSequence, ...terminators: ActionType[]) => {
+  const buildMatchedSequence = (matchedSequence, action) =>
```
lderrickable, line 84: I think this function should be named something else like "accumulateStream" or "accumulateSequence" or something like that.

griest024: How about `accumulateMatchedStream`? yeah, you're right what I call the sequence and the stream are different data structures. I shouldn't mix the terms.

libs/core/src/operators/substream.pipe.ts
griest024: That's better. I think `Watches for a particular stream of actions in the specified order and emits a list of those actions once all sequence steps have been matched.` is even a little more precise.

libs/core/src/operators/substream.pipe.ts
lderrickable: Oh, good point. I was actually just talking about the word "build". `accumulateMatchedSubstream` sounds good

griest024: oh right I don't think I read your comment fully. not sure what I was thinking. changed.

griest024:
@lderrickable don't merge yet

griest024:
@damienwebdev `bufferToggle` doesn't work that well. I vote either using this current implementation or  writing a better `buffer` operator.

griest024:
@damienwebdev @lderrickable I have removed comparator support and added Damien's fix for not using `skipWhile` (I'm dumb).

PR #965, opened 2020-07-01
feat(authorizenet): add action and state to check whether acceptJs ha…

lderrickable:
…s loaded

## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There is currently no way to verify that the AcceptJs library has loaded.

## What is the new behavior?
It isn't pretty, but there is now a way to verify that the library has loaded via a manually triggered action and a selector.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/authorizenet/src/actions/authorizenet.actions.ts
+++ b/libs/authorizenet/src/actions/authorizenet.actions.ts
@@ -1,13 +1,16 @@
 import { Action } from '@ngrx/store';
 
-import { DaffAuthorizeNetTokenRequest } from '../models/request/authorize-net-token-request';
 import { DaffCartAddress } from '@daffodil/cart';
+
+import { DaffAuthorizeNetTokenRequest } from '../models/request/authorize-net-token-request';
 
 export enum DaffAuthorizeNetActionTypes {
   UpdatePaymentAction = '[Daff-Authorize-Net] Update Payment',
   UpdatePaymentSuccessAction = '[Daff-Authorize-Net] Update Payment Success',
 	UpdatePaymentFailureAction = '[Daff-Authorize-Net] Update Payment Failure',
-	LoadAcceptJsAction = '[Daff-Authorize-Net] Load Accept Js'
+	LoadAcceptJsAction = '[Daff-Authorize-Net] Load Accept Js',
+	LoadAcceptJsSuccessAction = '[Daff-Authorize-Net] Load Accept Js Success',
+	LoadAcceptJsFailureAction = '[Daff-Authorize-Net] Load Accept Js Failure'
 }
 
 /**
@@ -48,10 +51,28 @@ export class DaffLoadAcceptJs implements Action {
 	readonly type = DaffAuthorizeNetActionTypes.LoadAcceptJsAction;
 }
 
+/**
+ * Indicates that the AcceptJs library has loaded successfully.
+ */
+export class DaffLoadAcceptJsSuccess implements Action {
+	readonly type = DaffAuthorizeNetActionTypes.LoadAcceptJsSuccessAction;
+}
+
+/**
+ * Indicates that the AcceptJs library has failed to load
+ */
+export class DaffLoadAcceptJsFailure implements Action {
+	readonly type = DaffAuthorizeNetActionTypes.LoadAcceptJsFailureAction;
+
+	constructor(public payload: string) {};
+}
+
 export type DaffAuthorizeNetActions<
 	T extends DaffAuthorizeNetTokenRequest = DaffAuthorizeNetTokenRequest
 > =
 	| DaffAuthorizeNetUpdatePayment<T>
 	| DaffAuthorizeNetUpdatePaymentSuccess
 	| DaffAuthorizeNetUpdatePaymentFailure
+	| DaffLoadAcceptJsSuccess
+	| DaffLoadAcceptJsFailure
 	| DaffLoadAcceptJs;
--- a/libs/authorizenet/src/effects/authorize-net.effects.ts
+++ b/libs/authorizenet/src/effects/authorize-net.effects.ts
@@ -1,25 +1,24 @@
 import { Injectable, Inject } from '@angular/core';
 import { Actions, Effect, ofType } from '@ngrx/effects';
-import { tap, switchMap, catchError, map } from 'rxjs/operators';
+import { switchMap, catchError, map, tap } from 'rxjs/operators';
 import { Observable, of } from 'rxjs';
 
 import { DaffCartPaymentActionTypes, DaffCartPaymentUpdateWithBilling } from '@daffodil/cart';
-import { substream } from '@daffodil/core';
+import { substream, backoff } from '@daffodil/core';
 
 import { 
 	DaffAuthorizeNetActionTypes, 
 	DaffAuthorizeNetUpdatePayment, 
 	DaffAuthorizeNetUpdatePaymentFailure, 
 	DaffLoadAcceptJs,
-	DaffAuthorizeNetUpdatePaymentSuccess
+	DaffAuthorizeNetUpdatePaymentSuccess,
+	DaffLoadAcceptJsFailure,
+	DaffLoadAcceptJsSuccess
 } from '../actions/authorizenet.actions';
 import { DaffAuthorizeNetTokenRequest } from '../models/request/authorize-net-token-request';
 import { DaffAuthorizeNetService, DaffAuthorizeNetDriver } from '../drivers/interfaces/authorize-net-service.interface';
 import { DaffAuthorizeNetPaymentId } from '../drivers/interfaces/authorize-net-payment-id.token';
-import { DaffAuthorizeNetConfig, DaffAuthorizeNetConfigToken } from '../drivers/public_api';
-
-const ACCEPT_JS_SANDBOX_URL = 'https://jstest.authorize.net/v1/Accept.js';
-const ACCEPT_JS_PRODUCTION_URL = 'https://js.authorize.net/v1/Accept.js';
+import { DaffAcceptJsLoadingService } from '../services/accept-js-loading.service';
 
 @Injectable()
 export class DaffAuthorizeNetEffects<T extends DaffAuthorizeNetTokenRequest = DaffAuthorizeNetTokenRequest> {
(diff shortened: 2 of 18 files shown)


libs/authorizenet/src/reducers/authorize-net/authorize-net.reducer.ts
```diff
@@ -27,6 +31,19 @@ export function daffAuthorizeNetReducer <T extends DaffAuthorizeNetTokenRequest>
 				loading: false,
 				error: action.payload
 			};
+		case DaffAuthorizeNetActionTypes.CheckAcceptJsAction:
+			return {
```
damienwebdev, line 35: This looks like effect behavior that can be exponentially backedoff and retried, not reducer code..

lderrickable: sure, I can do that

lderrickable: done

libs/authorizenet/src/effects/authorize-net.effects.ts
```diff
@@ -66,16 +73,33 @@ export class DaffAuthorizeNetEffects<T extends DaffAuthorizeNetTokenRequest = Da
...
+	)
+	
+	@Effect()
+	checkAcceptJs$ : Observable<any> = this.actions$.pipe(
+		ofType(DaffAuthorizeNetActionTypes.CheckAcceptJsAction),
+		debounce(() => interval(this.checkAcceptJsRetryTimer)),
+		map(() => {
+			try {
+				// tslint:disable-next-line: no-unused-expression
+				Accept;
```
damienwebdev, line 97: typeof Accept === "undefined"

libs/authorizenet/src/effects/authorize-net.effects.ts
```diff
@@ -66,16 +73,33 @@ export class DaffAuthorizeNetEffects<T extends DaffAuthorizeNetTokenRequest = Da
...
     ofType(DaffAuthorizeNetActionTypes.LoadAcceptJsAction),
-		tap((action: DaffLoadAcceptJs) => {
+		map((action: DaffLoadAcceptJs) => {
 			const acceptJsScript = document.createElement('script');
 			acceptJsScript.async = true;
 			acceptJsScript.setAttribute('type', 'text/javascript');
 			acceptJsScript.setAttribute('src', this.config.production ? ACCEPT_JS_PRODUCTION_URL : ACCEPT_JS_SANDBOX_URL);
 			acceptJsScript.setAttribute('charset', 'utf-8');
 			document.body.appendChild(acceptJsScript);
+			return new DaffCheckAcceptJs();
```
damienwebdev, line 86: I don't think we need this action, I think we can simply start the `checkAcceptJs$` when the `DaffLoadAcceptJs` triggers.

libs/authorizenet/src/effects/authorize-net.effects.ts
```diff
@@ -66,16 +73,33 @@ export class DaffAuthorizeNetEffects<T extends DaffAuthorizeNetTokenRequest = Da
...
+		})
+	)
+	
+	@Effect()
+	checkAcceptJs$ : Observable<any> = this.actions$.pipe(
+		ofType(DaffAuthorizeNetActionTypes.CheckAcceptJsAction),
+		debounce(() => interval(this.checkAcceptJsRetryTimer)),
+		map(() => {
```
damienwebdev, line 94: Taken from the [Angular docs:](https://angular.io/guide/practical-observable-usage#exponential-backoff)

```bash
function backoff(maxTries, ms) {
 return pipe(
   retryWhen(attempts => zip(range(1, maxTries), attempts)
     .pipe(
       map(([i]) => i * i),
       mergeMap(i =>  timer(i * ms))
     )
   )
 );
}
```

This way, there's no class prop.

This should also fail.

libs/authorizenet/src/effects/authorize-net.effects.ts
lderrickable: This would have worked if I could've gotten backoff to work, but I don't think it makes sense without it.

libs/authorizenet/src/effects/authorize-net.effects.ts
lderrickable: changed

libs/authorizenet/src/effects/authorize-net.effects.ts
lderrickable: Try as I might (for hours), I couldn't get this to work. I converted the Accept check into an observable that would throw an error when it failed, but backoff just let everything through and the check never gets triggered again.

lderrickable: added backoff operator

libs/authorizenet/src/effects/authorize-net.effects.ts
lderrickable: done

lderrickable:
@damienwebdev This is ready for a re-review

libs/authorizenet/src/effects/authorize-net.effects.ts
```diff
@@ -66,16 +72,23 @@ export class DaffAuthorizeNetEffects<T extends DaffAuthorizeNetTokenRequest = Da
...
-			acceptJsScript.async = true;
-			acceptJsScript.setAttribute('type', 'text/javascript');
-			acceptJsScript.setAttribute('src', this.config.production ? ACCEPT_JS_PRODUCTION_URL : ACCEPT_JS_SANDBOX_URL);
-			acceptJsScript.setAttribute('charset', 'utf-8');
-			document.body.appendChild(acceptJsScript);
-		})
-  )
+		map((action: DaffLoadAcceptJs) => {
+			if(typeof Accept === 'undefined') {
+				const acceptJsScript = document.createElement('script');
```
damienwebdev, line 80: `document` isn't available server-side. You should probably extract most of this code to a separate service and inject it here.

libs/authorizenet/src/actions/authorizenet.actions.ts
```diff
@@ -1,13 +1,16 @@
...
 
+import { DaffAuthorizeNetTokenRequest } from '../models/request/authorize-net-token-request';
+
 export enum DaffAuthorizeNetActionTypes {
   UpdatePaymentAction = '[Daff-Authorize-Net] Update Payment',
   UpdatePaymentSuccessAction = '[Daff-Authorize-Net] Update Payment Success',
 	UpdatePaymentFailureAction = '[Daff-Authorize-Net] Update Payment Failure',
-	LoadAcceptJsAction = '[Daff-Authorize-Net] Load Accept Js'
+	LoadAcceptJsAction = '[Daff-Authorize-Net] Load Accept Js',
+	CheckAcceptJsAction = '[Daff-Authorize-Net] Check Accept Js',
```
damienwebdev, line 12: Check isn't necessary any longer, I think?

libs/authorizenet/src/actions/authorizenet.actions.ts
```diff
@@ -1,13 +1,16 @@
...
+	AcceptJsLoadedAction = '[Daff-Authorize-Net] Accept Js Loaded'
```
damienwebdev, line 13: Needs a failure action.

libs/core/src/operators/backoff.pipe.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+
+import { backoff } from './backoff.pipe';
+declare var somethingThrowsError;
+
+describe('Core | Operators | Backoff', () => {
+  let actions$: Observable<Action>;
+	const action1: Action = {type: 'type1'};
+
+	it(`should retry the observable when the observable throws an error
+	    should pass through when the observable does not throw an error`, () => {
```
damienwebdev, line 14: Is this test name a mistake?

libs/core/src/operators/backoff.pipe.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+		actions$ = hot('a', { a: action1 });
+		const streamThatThrowsAnError4Times$ = actions$.pipe(
+			map(() => 0),
+			switchMap((count) => of(null).pipe(
+				map(() => count++),
+				map((count1) => count1 < 4 ? somethingThrowsError : count1),
```
damienwebdev, line 20: can "Something that throws an error" just be a throw?

libs/core/src/operators/backoff.pipe.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+				backoff(10, 0),
+				map((count2) => expect(count2).toEqual(4))
+			))
+		);
+		
+		streamThatThrowsAnError4Times$.subscribe();
+
+		expect(true).toBeTruthy();
+	});
+
```
damienwebdev, line 30: This needs a test around the backoff behavior of multiple increments, e.g. 1,2,3,4, etc.

libs/core/src/operators/backoff.pipe.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+	it('should retry the observable until the max number of retries is reached', () => {
+		actions$ = hot('a', { a: action1 });
+		const streamThatThrowsAnErrorInfinitely$ = actions$.pipe(
+			switchMap(() => of(null).pipe(
+				map(() => somethingThrowsError),
+				backoff(10, 0),
+				map(() => fail('This point in the test should not be reached'))
+			)
+		));
+
```
damienwebdev, line 40: This needs a test about the length of frames backed off per backoff step.

libs/core/src/operators/backoff.pipe.spec.ts
lderrickable: It might be, but I'm not aware that it is

libs/core/src/operators/backoff.pipe.spec.ts
lderrickable: I don't think so

lderrickable:
> needs PR split into appropriate commits, but otherwise, lgtm. We can do the reducer for failure handling in another PR where we refactor the state model.

separated commits

damienwebdev:
@lderrickable can you add the pr # into the commit message, and then this looks good for rebase merge.

lderrickable:
> @lderrickable can you add the pr # into the commit message, and then this looks good for rebase merge.

Done

PR #972, opened 2020-07-02
feat(order): add Magento guest orders driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Adds error map and transform
- Adds validators for guest orders response
- Adds order transformer
- Adds guest order driver service

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/order/src/drivers/src/magento/errors/codes.ts
+++ b/libs/order/src/drivers/src/magento/errors/codes.ts
@@ -1,0 +1,3 @@
+export enum MagentoOrderGraphQlErrorCode {
+	CART_NOT_FOUND = 'graphql-no-such-entity',
+}
--- a/libs/order/src/drivers/src/magento/order.service.spec.ts
+++ b/libs/order/src/drivers/src/magento/order.service.spec.ts
@@ -1,0 +1,455 @@
+import { TestBed } from '@angular/core/testing';
+import { ApolloTestingController, ApolloTestingModule } from 'apollo-angular/testing';
+import { GraphQLError } from 'graphql';
+import { catchError } from 'rxjs/operators';
+
+import { DaffCartNotFoundError } from '@daffodil/cart';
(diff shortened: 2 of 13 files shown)


libs/order/src/drivers/src/magento/errors/codes.ts
```diff
@@ -0,0 +1,3 @@
+export enum MagentoOrderGraphQlErrorCode {
+	CART_NOT_FOUND = 'graphql-no-such-entity',
```
lderrickable, line 2: Could we reuse the error code `@daffodil/cart`? It's the same code

libs/order/src/drivers/src/magento/order.service.spec.ts
```diff
@@ -0,0 +1,455 @@
...
+
+          op.flush({
+            data: mockGetOrdersResponse
+          });
+        });
+      });
+    });
+
+    describe('when the call to the Magento API is unsuccessful', () => {
+      it('should throw a DaffCartNotFoundError', done => {
```
lderrickable, line 428: I don't think you need to test the error transform logic here, but I think this "it" block should just be "should throw an error" or something more general. Just to be clear that this won't necessarily throw a DaffCartNotFoundError.

libs/order/src/drivers/src/magento/errors/codes.ts
griest024: I don't think it gets exported.

libs/order/src/drivers/src/magento/order.service.spec.ts
griest024: changed

PR #979, opened 2020-07-03
fix(all): helper function return false on ID of 0

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The helper functions would return `false` when the factory generated a field with `0`.

## What is the new behavior?
All `faker.random.number` calls have a minimum value of `1`, which will evaluate to `true` in the helper functions.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/checkout/testing/src/order/factories/order-payment.factory.ts
+++ b/libs/checkout/testing/src/order/factories/order-payment.factory.ts
@@ -15,8 +15,8 @@ export class MockOrderPayment implements DaffOrderPayment {
     updated_at = new Date();
     method = 'card';
     cc_type = 'visa';
-    cc_number_enc = faker.random.number({min: 1, max: 1000}).toString();
-    cc_last4 = faker.random.number({min: 1, max: 1000}).toString();
+    cc_number_enc = faker.random.number({min: 1000, max: 9999}).toString();
+    cc_last4 = faker.random.number({min: 1000, max: 9999}).toString();
     cc_cid_enc = faker.random.number({min: 1, max: 1000}).toString();
     cc_owner = 'owner';
     cc_exp_month = 'month';
--- a/libs/checkout/testing/src/order/factories/order-payment.factory.ts
+++ b/libs/checkout/testing/src/order/factories/order-payment.factory.ts
@@ -9,15 +9,15 @@ import { DaffModelFactory } from '@daffodil/core/testing';
  * @deprecated
  */
 export class MockOrderPayment implements DaffOrderPayment {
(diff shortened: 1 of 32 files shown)


griest024:
@lderrickable I went ahead and changed every single random number generation to have a min value of 1. Can you make sure that won't break anything, especially in category and product?

griest024:
Would it be better to change the helper functions instead?

libs/checkout/testing/src/order/factories/order-payment.factory.ts
```diff
@@ -9,15 +9,15 @@ import { DaffModelFactory } from '@daffodil/core/testing';
...
+    quote_id = faker.random.number({min: 1, max: 1000});
     created_at = new Date();
     updated_at = new Date();
     method = 'card';
     cc_type = 'visa';
-    cc_number_enc = faker.random.number(1000).toString();
-    cc_last4 = faker.random.number(1000).toString();
-    cc_cid_enc = faker.random.number(1000).toString();
+    cc_number_enc = faker.random.number({min: 1, max: 1000}).toString();
+    cc_last4 = faker.random.number({min: 1, max: 1000}).toString();
```
lderrickable, line 19: this should be `{min: 1000, max: 9999}`

libs/order/testing/src/factories/order-payment.factory.ts
```diff
@@ -6,13 +6,13 @@ import * as faker from 'faker/locale/en_US';
...
-    payment_id = faker.random.number(1000);
-    order_id = faker.random.number(1000);
+    payment_id = faker.random.number({min: 1, max: 1000});
+    order_id = faker.random.number({min: 1, max: 1000});
     created_at = faker.date.past().toString();
     updated_at = faker.date.past().toString();
     method = 'card';
     cc_type = 'visa';
-    cc_last4 = faker.random.number(1000).toString();
+    cc_last4 = faker.random.number({min: 1, max: 1000}).toString();
```
lderrickable, line 15: same here

libs/checkout/testing/src/order/factories/order-payment.factory.ts
griest024: done

libs/order/testing/src/factories/order-payment.factory.ts
griest024: done

PR #980, opened 2020-07-04
feat(product): add is_default to the composite product model

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
When I originally wrote the transformers for the composite product, I didn't think it was necessary to have the `is_default` field. Turns out I need it to give the app dev the full flexibility that the magento admin panel provides.


griest024:
I think its worth adding a selector for the default item option of a bundled product.

lderrickable:
I would like to have it so that the app dev doesn't need to know anything about the defaults. I think ideally, all the app dev worries about is what options are selected and daffodil takes care of all the default logic. Then if there is no default, daffodil defaults to the first option if it is required. What do you think?

griest024:
That seems good.

PR #995, opened 2020-07-12
feat(cart): store cartId in order result

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The cart ID corresponding to the order result will be lost when a new cart is created.

## What is the new behavior?
Stores the cart ID used to place the order in the order result object, which allows the cart ID to be used for order fetching even after a new cart is created.

```
[x] Yes
[ ] No
```

Deprecates `DaffCartOrderResult#id` in favor of `DaffCartOrderResult#orderId`.

## Other information

### DIFF
--- a/libs/cart/src/facades/cart/cart-facade.interface.ts
+++ b/libs/cart/src/facades/cart/cart-facade.interface.ts
@@ -48,7 +48,8 @@ export interface DaffCartFacadeInterface<
   orderResultLoading$: Observable<boolean>;
 	orderResultErrors$: Observable<string[]>;
 	orderResult$: Observable<V>;
-	orderResultId$: Observable<V['id']>;
+	orderResultId$: Observable<V['orderId']>;
+	orderResultCartId$: Observable<V['cartId']>;
 
 	getCartItemDiscountedTotal(itemId: string | number): Observable<number>;
 }
--- a/libs/cart/src/reducers/cart-order/cart-order.reducer.spec.ts
+++ b/libs/cart/src/reducers/cart-order/cart-order.reducer.spec.ts
@@ -31,9 +31,11 @@ describe('Cart | Reducer | CartOrder', () => {
     let result;
     let state: DaffCartOrderReducerState;
     let orderId;
+    let cartId;
 
     beforeEach(() => {
-      orderId = 'id';
+      orderId = 'orderId';
+      cartId = 'cartId';
       state = {
         ...initialState,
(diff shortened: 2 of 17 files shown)


libs/cart/src/facades/cart/cart-facade.interface.ts
```diff
@@ -48,7 +48,8 @@ export interface DaffCartFacadeInterface<
   orderResultLoading$: Observable<boolean>;
 	orderResultErrors$: Observable<string[]>;
 	orderResult$: Observable<V>;
-	orderResultId$: Observable<V['id']>;
+	orderResultId$: Observable<V['orderId']>;
+	orderResultCartId$: Observable<V['cartId']>;
```
lderrickable, line 52: I don't think orderResultCartId should be on the facade. The app dev should never need it. If he/she does, it can be accessed with the selector.

libs/cart/src/reducers/cart-order/cart-order.reducer.spec.ts
```diff
@@ -39,13 +39,16 @@ describe('Cart | Reducer | CartOrder', () => {
...
-      const cartPlaceOrderSuccess = new DaffCartPlaceOrderSuccess({id: orderId});
+      const cartPlaceOrderSuccess = new DaffCartPlaceOrderSuccess({
+        orderId,
+        cartId: 'cartId',
+      });
 
       result = reducer(state, cartPlaceOrderSuccess);
     });
 
     it('should set the order ID from action.payload', () => {
```
lderrickable, line 50: This should probably change to "should set the order result from action.payload"

libs/cart/src/facades/cart/cart-facade.interface.ts
griest024: The app dev would need the cart ID to fetch guest orders. I have a PR out for that functionality.

lderrickable: Oh, I'd thought that was being handled internally by daffodil, but I guess it makes sense they would need to pass it.

libs/cart/src/reducers/cart-order/cart-order.reducer.spec.ts
griest024: changed

PR #996, opened 2020-07-12
feat(order): pass cart ID into state layer

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The effects layer would get the cart ID from the storage service for loading orders.

## What is the new behavior?
Since the cart ID in the storage service will not match the one needed for the order load, the cart ID must now be explicitly passed.

```
[x] Yes
[ ] No
```

Changes the payload format for `DaffOrderLoad` and adds an optional payload for `DaffOrderList`.

## Other information

### DIFF
--- a/libs/order/src/actions/order.actions.ts
+++ b/libs/order/src/actions/order.actions.ts
@@ -1,5 +1,7 @@
 import { Action } from '@ngrx/store';
 
+import { DaffCart } from '@daffodil/cart';
+
 import { DaffOrder } from '../models/public_api';
 
 export enum DaffOrderActionTypes {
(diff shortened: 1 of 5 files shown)


libs/order/src/actions/order.actions.ts
```diff
@@ -11,10 +13,21 @@ export enum DaffOrderActionTypes {
...
+  T extends DaffOrder = DaffOrder,
+  V extends DaffCart = DaffCart
+> implements Action {
   readonly type = DaffOrderActionTypes.OrderLoadAction;
 
-  constructor(public payload: T['id']) {}
+  constructor(public payload: {
+    orderId: T['id'],
+    cartId?: V['id'],
+  }) {}
```
lderrickable, line 30: what do you think about this sort of format vs:
```
constructor(public orderId: T['id'], public cartId?: V['id']) {}
```

griest024: I don't have any strong preference either way. I guess yours has the benefit of not being a breaking change. Plus it exactly matches the driver call so its probably better that way.

griest024: changed

lderrickable: I was thinking that the usage looks a little cleaner: 
`new Action({ value1: value1, value2: value2})` vs `new Action(value1, value2)`

PR #1005, opened 2020-07-17
feat(product): add qty to composite product option state

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Composite Product Option state does not hold qty, which is a feature needed for composite products.

## What is the new behavior?
Composite Product Option state now has qty.

```
[ ] Yes
[x] No
```

## Other information
I'd like to make a minor (though breaking) change to the add to cart action payload to reflect the same naming convention of this PR. I'll do it separately.

### DIFF
--- a/libs/product/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
+++ b/libs/product/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
@@ -7,7 +7,7 @@ import { daffCompositeProductAppliedOptionsEntitiesAdapter } from './composite-p
 import { DaffProduct, DaffProductTypeEnum } from '../../models/product';
 import { DaffCompositeProductActions, DaffCompositeProductActionTypes } from '../../actions/composite-product.actions';
 import { DaffCompositeProduct } from '../../models/composite-product';
-import { DaffCompositeProductEntity } from './composite-product-entity';
+import { DaffCompositeProductEntity, DaffCompositeProductEntityItem } from './composite-product-entity';
 import { DaffCompositeProductItem } from '../../models/composite-product-item';
 
 /**
@@ -44,7 +44,10 @@ export function daffCompositeProductEntitiesReducer<T extends DaffProduct, V ext
 					id: action.id,
 					items: {
 						...state.entities[action.id].items,
-						[action.itemId]: action.optionId
+						[action.itemId]: {
+							value: action.optionId,
+							qty: action.qty ? action.qty : 1
+						}
 					}
 				},
 				state
@@ -64,12 +67,17 @@ function buildCompositeProductAppliedOptionsEntity(product: DaffCompositeProduct
 	}
 }
 
-function getDefaultOption(item: DaffCompositeProductItem): string {
+function getDefaultOption(item: DaffCompositeProductItem): DaffCompositeProductEntityItem {
 	const defaultOptionIndex = item.options.findIndex(option => option.is_default);
 
 	if(defaultOptionIndex > -1) {
-		return item.options[defaultOptionIndex].id;
+		return {
+			value: item.options[defaultOptionIndex].id,
+			qty: 1
+		}
 	} else {
-		return item.required ? item.options[0].id : null;
+		return item.required ? 
+			{ value: item.options[0].id, qty: 1 } :
+			{ value: null, qty: null }
 	}
 }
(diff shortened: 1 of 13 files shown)


libs/product/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
```diff
@@ -57,13 +54,24 @@ export function daffCompositeProductEntitiesReducer<T extends DaffProduct, V ext
...
 
+function applyItem(currentItems: DaffCompositeProductEntityItem[], appliedItemCode: string, appliedItemValue: string, quantity: number): DaffCompositeProductEntityItem[] {
+	const itemIndex = currentItems.findIndex(item => item.code === appliedItemCode);
+	currentItems[itemIndex] = {
+		code: appliedItemCode,
+		value: appliedItemValue,
+		qty: quantity
+	}
+	return currentItems;
+}
```
griest024, line 73: This should not mutate `currentItems`.

libs/product/src/reducers/composite-product-entities/composite-product-entity.ts
```diff
@@ -3,5 +3,11 @@ import { DaffCompositeProductItemOption } from '../../models/composite-product-i
 
 export interface DaffCompositeProductEntity {
 	id: string;
-	items: Dictionary<DaffCompositeProductItemOption['id']>;
+	items: DaffCompositeProductEntityItem[];
```
griest024, line 6: I don't understand why this needs to be an array. Is order significant?

lderrickable: oh, I guess it could be:
```
{
    itemIdValue: {
      value: optionId,
      qty: quantity
   }
}
```
I'll change it to that. Order doesn't matter.

griest024: It seems that its much simpler to deal with it as a dictionary. If you need the selector to be an array I would just convert it to array in the selector.

libs/product/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
lderrickable: fixed because I'm not working with arrays anymore.

libs/product/src/reducers/composite-product-entities/composite-product-entity.ts
lderrickable: changed back to dictionary. I force pushed to the branch. The changes are closer to what it was before the original commit.

libs/product/src/reducers/composite-product-entities/composite-product-entity.ts
```diff
@@ -1,7 +1,13 @@
 import { Dictionary } from '@ngrx/entity';
+
 import { DaffCompositeProductItemOption } from '../../models/composite-product-item';
```
griest024, line 3: unused import

libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -117,13 +121,13 @@ export const getDaffCompositeProductSelectors = (() => {
...
-function getOptionsDiscountAmount(product: DaffCompositeProduct, appliedOptions: Dictionary<string>): number {
+function getOptionsDiscountAmount(product: DaffCompositeProduct, appliedOptions: Dictionary<DaffCompositeProductEntityItem>): number {
 	return product.items.reduce((acc, item) => {
-		const itemOptionDiscount = item.options.find(option => option.id === appliedOptions[item.id]).discount;
+		const itemOptionDiscount = item.options.find(option => option.id === appliedOptions[item.id].value).discount;
 
 		return daffAdd(
 			acc, 
-			(itemOptionDiscount && itemOptionDiscount.amount > 0 ? itemOptionDiscount.amount : 0)
+			daffMultiply((itemOptionDiscount && itemOptionDiscount.amount > 0 ? itemOptionDiscount.amount : 0), appliedOptions[item.id].qty)
```
griest024, line 130: I would do the `itemOptionDiscount && itemOptionDiscount.amount > 0` check outside of the `daffMultiply` call so that `daffMultiply` is only called when its needed.

libs/product/src/reducers/composite-product-entities/composite-product-entity.ts
lderrickable: removed

libs/product/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: fixed

PR #1012, opened 2020-07-25
feat(product): add ranged price selectors for composite products

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
Adds ranged prices for composite products. Two kinds of ranged price selectors were needed:

One represents the full possible range of prices for the composite product, regardless of what kinds of items have been chosen by default behind the scenes or by the user. This would be used in an area of the application where the customer cannot see what items and options are available, such as which a product quick-view or a product card on a category page. Think of this as the total possible range of what a user could configure for a composite product.

The other kind of ranged price is for a product page and will typically only be needed for composite products that have required items which also have no default option. This range is dependent on the applied options for the composite product. If all required items are chosen, this range will not exist, because the final price of the product can be represented by a single number. If a composite product has a required item that has no option chosen, then the final price of the product cannot be represented by a single number, because that single number cannot be known until all required items are decided.

### DIFF
--- a/libs/product/src/selectors/composite-product/composite-product.selectors.ts
+++ b/libs/product/src/selectors/composite-product/composite-product.selectors.ts
@@ -101,7 +101,7 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
 
 			return (<DaffCompositeProduct>product).items.reduce((acc, item) => daffAdd(
 				acc, 
-				appliedOptions[item.id].value ? item.options.find(option => option.id === appliedOptions[item.id].value).price : getMinimumCompositeItemPrice(item)
+				appliedOptions[item.id].value ? findAppliedCompositeOptionPrice(item, appliedOptions) : getMinimumCompositeItemPrice(item)
 			), product.price);
 		}
 	);
@@ -123,7 +123,7 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
 			return (<DaffCompositeProduct>product).items.reduce((acc, item) => 
 				daffAdd(
 					acc, 
-					appliedOptions[item.id].value ? item.options.find(option => option.id === appliedOptions[item.id].value).price : getMaximumRequiredItemPrice(item)
+					appliedOptions[item.id].value ? findAppliedCompositeOptionPrice(item, appliedOptions) : getMaximumRequiredItemPrice(item)
 				), product.price);
 		}
 	);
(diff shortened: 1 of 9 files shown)


libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -29,6 +36,108 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
...
+		(products, appliedOptionsEntities, props) => {
+			const product = selectProduct.projector(products, { id: props.id });
+			if(product.type !== DaffProductTypeEnum.Composite) {
+				return undefined;
+			}
+			const appliedOptions = selectCompositeProductAppliedOptions.projector(appliedOptionsEntities, { id: props.id });
+
+			return (<DaffCompositeProduct>product).items.reduce((acc, item) => daffAdd(
+				acc, 
+				appliedOptions[item.id].value ? item.options.find(option => option.id === appliedOptions[item.id].value).price : getMinimumCompositeItemPrice(item)
```
griest024, line 104: I think you should extract this into:

```ts
function findAppliedCompositeOptionPrice(item: DaffCompositeProductItem, appliedOptions: Dictionary<DaffCompositeProductEntityItem>): number {
  return item.options.find(option => option.id === appliedOptions[item.id].value).price
}
```

and use it in `selectCompositeProductPrice`, `selectCompositeProductMinPrice`, and `selectCompositeProductMaxPrice`.

libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -29,6 +36,108 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
...
+		(products, props) => {
+			const product = selectProduct.projector(products, { id: props.id });
+			if(product.type !== DaffProductTypeEnum.Composite) {
+				return undefined;
+			}
+			
+			return (<DaffCompositeProduct>product).items.reduce((acc, item) => 
+				daffAdd(
+					acc, 
+					Math.max(...item.options.map(option => option.price))
```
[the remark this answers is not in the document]
lderrickable: No, because I want to include the optional items to get the maximum price possible

libs/product/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: done

PR #1013, opened 2020-07-27
fix(cart): driver failures cause effect obsevables to complete

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When a driver call fails, certain effects handle those errors and then complete. This causes the effects to never run again.

## What is the new behavior?
Effects will not complete as a result of driver or storage failures.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/src/effects/cart-coupon.effects.ts
+++ b/libs/cart/src/effects/cart-coupon.effects.ts
@@ -41,62 +41,57 @@ export class DaffCartCouponEffects<
 
   @Effect()
   apply$ = this.actions$.pipe(
-    tap(action => console.log('coupon', action)),
     ofType(DaffCartCouponActionTypes.CartCouponApplyAction),
-    switchMap((action: DaffCartCouponApply<V>) =>
-      this.driver.apply(this.storage.getCartId(), action.payload).pipe(
-        map(resp => new DaffCartCouponApplySuccess(resp)),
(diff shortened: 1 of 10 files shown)


libs/cart/src/effects/cart-coupon.effects.ts
```diff
@@ -41,16 +41,18 @@ export class DaffCartCouponEffects<
 
   @Effect()
   apply$ = this.actions$.pipe(
+    tap(action => console.log('coupon', action)),
```
lderrickable, line 44: console.log needs removed

griest024: removed

griest024:
@lderrickable I changed the implementation so that storage errors will also not cause the effects to complete

libs/cart/src/effects/cart-resolver.effects.ts
```diff
@@ -18,14 +18,7 @@ import { DaffCartNotFoundError } from '../errors/not-found';
 import { DaffStorageServiceError } from '@daffodil/core';
 
 function catchResolveErrors(error: Error) {
```
lderrickable, line 20: Looks like this should have stuff in it, or should be removed

griest024: yeah it was leftover. removed

PR #1024, opened 2020-08-07
Cart item types

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Cart items now have a `type` field for what kind of cart item it is. 

```
[ ] Yes
[x] No
```

## Other information
I plan on doing another PR to move the `DaffCartItemInputType` enum into the `DaffCartItem` and rename it to `DaffCartItemType`, because that seems like the most basic model.

### DIFF
--- a/libs/cart/src/models/composite-cart-item.ts
+++ b/libs/cart/src/models/composite-cart-item.ts
@@ -1,0 +1,12 @@
+import { DaffCartItem } from './cart-item';
+import { DaffCartItemInputType } from './cart-item-input';
+
+export interface DaffCompositeCartItem extends DaffCartItem {
+	type: DaffCartItemInputType.Composite;
+	options: DaffCompositeCartItemOption[];
+}
+
+export interface DaffCompositeCartItemOption {
+	option_label: string;
+	value_label: string;
+}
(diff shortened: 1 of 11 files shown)


griest024:
@lderrickable can you rebase to clean up the diff?

libs/cart/src/models/composite-cart-item.ts
```diff
@@ -0,0 +1,12 @@
+import { DaffCartItem } from './cart-item';
+import { DaffCartItemInputType } from './cart-item-input';
+
+export interface DaffCompositeCartItem extends DaffCartItem {
+	type: DaffCartItemInputType.Composite;
+	options: DaffCompositeCartItemOption[];
+}
+
+export interface DaffCompositeCartItemOption {
+	option_label: string;
```
griest024, line 10: Why are these fields suffixed with `_label`?

lderrickable: On the composite product, there is an option id, value, and label. It was just for clarity

lderrickable:
> @lderrickable can you rebase to clean up the diff?

done

libs/cart/src/models/composite-cart-item.ts
griest024: okay so all Daffodil cares about for the cart item is the label, not the actual value?

lderrickable: Nothing else comes back from magento anyway, but I think we should just need the labels for the cart item, since you shouldn't be able to edit the item options in the cart.

griest024: okay that makes sense.

PR #1028, opened 2020-08-11
feat(cart): listen for resolve success and failure in cart guards

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Cart guards will proceed on resolve success, but hang indefinitely on cart resolve failure.

## What is the new behavior?
- Adds state for storing the resolved state of the cart. False initially and when `DaffResolveCart` is dispatched. True when cart resolution succeeds or fails.
- Cart guards will check the above state to determine when to proceed with the cart field check. This results in the behavior of cart guards proceeding on both cart resolution success _and_ failure.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/src/actions/cart.actions.ts
+++ b/libs/cart/src/actions/cart.actions.ts
@@ -97,10 +97,10 @@ export class DaffResolveCart implements Action {
 	readonly type = DaffCartActionTypes.ResolveCartAction;
 }
 
-export class DaffResolveCartSuccess<T extends DaffCart = DaffCart> implements Action {
+export class DaffResolveCartSuccess implements Action {
   readonly type = DaffCartActionTypes.ResolveCartSuccessAction;
 
-  constructor(public payload: Partial<T>) {}
+  constructor() {}
 }
 
 export class DaffResolveCartFailure implements Action {
@@ -124,5 +124,5 @@ export type DaffCartActions<T extends DaffCart = DaffCart> =
   | DaffCartClearSuccess<T>
   | DaffCartClearFailure
   | DaffResolveCart
-  | DaffResolveCartSuccess<T>
+  | DaffResolveCartSuccess
   | DaffResolveCartFailure;
--- a/libs/cart/src/actions/cart.actions.ts
+++ b/libs/cart/src/actions/cart.actions.ts
@@ -16,7 +16,9 @@ export enum DaffCartActionTypes {
   CartClearAction = '[DaffCart] Cart Reset Action',
   CartClearSuccessAction = '[DaffCart] Cart Reset Success Action',
   CartClearFailureAction = '[DaffCart] Cart Reset Failure Action',
-  ResolveCartAction = '[DaffCart] Resolve Cart Action'
+  ResolveCartAction = '[DaffCart] Resolve Cart Action',
+  ResolveCartSuccessAction = '[DaffCart] Resolve Cart Success Action',
+  ResolveCartFailureAction = '[DaffCart] Resolve Cart Failure Action',
 }
 
 export class DaffCartStorageFailure implements Action {
@@ -95,6 +97,18 @@ export class DaffResolveCart implements Action {
 	readonly type = DaffCartActionTypes.ResolveCartAction;
 }
 
+export class DaffResolveCartSuccess<T extends DaffCart = DaffCart> implements Action {
+  readonly type = DaffCartActionTypes.ResolveCartSuccessAction;
+
+  constructor(public payload: Partial<T>) {}
+}
+
+export class DaffResolveCartFailure implements Action {
+  readonly type = DaffCartActionTypes.ResolveCartFailureAction;
+
+  constructor(public payload: string) {}
+}
+
 export type DaffCartActions<T extends DaffCart = DaffCart> =
   | DaffCartStorageFailure
   | DaffCartLoad
@@ -109,4 +123,6 @@ export type DaffCartActions<T extends DaffCart = DaffCart> =
   | DaffCartClear
   | DaffCartClearSuccess<T>
   | DaffCartClearFailure
-  | DaffResolveCart;
+  | DaffResolveCart
+  | DaffResolveCartSuccess<T>
+  | DaffResolveCartFailure;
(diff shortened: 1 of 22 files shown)


libs/cart/src/effects/cart-resolver.effects.spec.ts
```diff
@@ -56,161 +59,285 @@ describe('DaffCartResolverEffects', () => {
...
+
+        expect(effects.onResolveCart$).toBeObservable(expected);
+      });
+    });
+
+    describe('handling XHR errors when retrieving a cart', () => {
+      it('should dispatch a DaffCartLoadFailure action', () => {
+        const response = cold('#', {});
+        driver.get.and.returnValue(response);
+        driver.create.and.returnValue(response);
```
lderrickable, line 136: I feel like this line shouldn't be required for the test to pass.

lderrickable: If it is required, then maybe the test should be "should create a new cart", but that doesn't seem right either.

libs/cart/src/effects/cart-resolver.effects.spec.ts
```diff
@@ -56,161 +59,285 @@ describe('DaffCartResolverEffects', () => {
...
+        expect(effects.onResolveCart$).toBeObservable(expected);
+      });
+    });
+
+    describe('when the error thrown is a DaffStorageServiceError', () => {
+
+      it('should indicate that the storage service has failed', () => {
+        const response = cold('#', {}, new DaffStorageServiceError());
+        driver.get.and.returnValue(response);
+        driver.create.and.returnValue(response);
```
lderrickable, line 208: These services shouldn't throw this error, right?

libs/cart/src/effects/cart-resolver.effects.spec.ts
```diff
@@ -56,161 +59,285 @@ describe('DaffCartResolverEffects', () => {
...
+          b: cartStorageFailureAction
+        });
+
+        expect(effects.onResolveCart$).toBeObservable(expected);
+      });
+    });
+  });
+
+  describe('onResolveCartSuccess$', () => {
+    describe('when cart resolution is followed by a successful cart load', () => {
```
lderrickable, line 223: I think this should be "when a cart resolve action is followed by a successful cart load", since "resolution" sounds like you're already at ResolveSuccess

libs/cart/src/effects/cart-resolver.effects.spec.ts
```diff
@@ -56,161 +59,285 @@ describe('DaffCartResolverEffects', () => {
...
+        const resolveCartAction = new DaffResolveCart();
+        const createCartFailureAction = new DaffCartCreateFailure(error);
+
+        actions$ = hot('--a--b', {
+          a: resolveCartAction,
+          b: createCartFailureAction
+        });
+      })
+
+      it('should indicate cart resolution success', () => {
```
lderrickable, line 317: wording is wrong here

libs/cart/src/effects/cart-resolver.effects.spec.ts
```diff
@@ -56,161 +59,285 @@ describe('DaffCartResolverEffects', () => {
...
+        const resolveCartAction = new DaffResolveCart();
+        const resolveCartSuccessAction = new DaffResolveCartSuccess(stubCart);
+
+        actions$ = hot('--a--b', {
+          a: resolveCartAction,
+          b: resolveCartSuccessAction
+        });
+      })
+
+      it('should not indicate cart resolution success', () => {
```
lderrickable, line 336: wording is wrong here

libs/cart/src/reducers/cart-resolve/cart-resolve.reducer.ts
```diff
@@ -0,0 +1,39 @@
...
+        resolved: false
+      };
+
+    case DaffCartActionTypes.ResolveCartSuccessAction:
+      return {
+        ...state,
+        cart: {
+          ...state.cart,
+          ...action.payload
+        },
```
lderrickable, line 26: I don't think this reducer should need to track the cart. That should be handled by the actions that end up triggering the ResolveCartSuccessAction like DaffCartCreateSuccess.

libs/cart/src/actions/cart.actions.ts
```diff
@@ -95,6 +97,18 @@ export class DaffResolveCart implements Action {
 	readonly type = DaffCartActionTypes.ResolveCartAction;
 }
 
+export class DaffResolveCartSuccess<T extends DaffCart = DaffCart> implements Action {
+  readonly type = DaffCartActionTypes.ResolveCartSuccessAction;
+
+  constructor(public payload: Partial<T>) {}
```
lderrickable, line 103: I don't think this should have a payload. The Actions that trigger the ResolveCartSuccess should handle setting the cart on state. This action should just set the resolve boolean

libs/cart/src/effects/cart-resolver.effects.spec.ts
griest024: yeah its not, removed

libs/cart/src/effects/cart-resolver.effects.spec.ts
griest024: I didn't write any of these tests but you're right, this test is set up incorrectly. fixed

libs/cart/src/effects/cart-resolver.effects.spec.ts
griest024: changed

libs/cart/src/effects/cart-resolver.effects.spec.ts
griest024: good catch, fixed

libs/cart/src/effects/cart-resolver.effects.spec.ts
griest024: fixed

libs/cart/src/reducers/cart-resolve/cart-resolve.reducer.ts
griest024: good point, changed

libs/cart/src/actions/cart.actions.ts
griest024: changed

libs/cart/src/effects/cart-resolver.effects.spec.ts
lderrickable: Yea, I think I did. Oops

PR #1032, opened 2020-08-13
feat(core): create noop storage service

xelaint:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
+++ b/libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
@@ -30,7 +30,7 @@ describe('DaffNoopStorageService', () => {
 		expect(service.getItem('testkey')).toEqual(undefined);
 	});
 	
-	it('should be able to delete all items from memory', () => {
+	it('should not be able to delete all items from memory', () => {
 		service.setItem('testkey', 'testvalue');
 		service.setItem('testkey2', 'testvalue2');
 		 
--- a/libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
+++ b/libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
@@ -1,6 +1,6 @@
 import { DaffNoopStorageService } from './noopstorage.service';
 
-fdescribe('DaffNoopStorageService', () => {
+describe('DaffNoopStorageService', () => {
   let service: DaffNoopStorageService;
 
   beforeEach(() => {
@@ -23,7 +23,7 @@ fdescribe('DaffNoopStorageService', () => {
 		expect(service.getItem('testkey')).toEqual(undefined);
 	});
 	
-	it('should be able to delete an item from memory', () => {
+	it('should not be able to delete an item from memory', () => {
 		service.setItem('testkey', 'testvalue');
 		service.removeItem('testkey');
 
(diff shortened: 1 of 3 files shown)


libs/core/src/storage/noopstorage/noopstorage.service.ts
```diff
@@ -0,0 +1,8 @@
+import { DaffPersistenceService } from "../persistence.interface";
+
+export class DaffNoopStorageService implements DaffPersistenceService {
```
griest024, line 3: This should be decorated with `@Injectable`.

xelaint: done

libs/core/src/storage/noopstorage/noopstorage.service.ts
```diff
@@ -0,0 +1,12 @@
+import { Injectable } from '@angular/core';
+import { DaffPersistenceService } from '../persistence.interface';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class DaffNoopStorageService implements DaffPersistenceService {
+  setItem(key : string, value: any): void {};
+  getItem(key: string): any {};
```
griest024, line 9: test that this method actually return nothing

libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+		expect(service.getItem('testkey')).toEqual(undefined);
+	});
+	
+	it('should not be able to retrieve a value from memory', () => {
+    service.setItem('testkey', 'testvalue');
+
+		expect(service.getItem('testkey')).toEqual(undefined);
+	});
+	
+	it('should be able to delete an item from memory', () => {
```
griest024, line 26: I don't understand this behavior. I would think that a noop service should not be interacting with memory at all.

libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
```diff
@@ -1,11 +1,50 @@
...
```
damienwebdev, line 26: `not` I think is a mistype.

libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
xelaint: @damienwebdev

libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
xelaint: updated

libs/core/src/storage/noopstorage/noopstorage.service.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+	});
+	
+	it('should not be able to delete an item from memory', () => {
+		service.setItem('testkey', 'testvalue');
+		service.removeItem('testkey');
+
+		expect(service.getItem('testkey')).toEqual(undefined);
+	});
+	
+	it('should be able to delete all items from memory', () => {
```
griest024, line 33: you need a `not` here as well

xelaint: updated

xelaint: updated

PR #1039, opened 2020-08-19
feat(product): add stock_status field

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
This adds a stock_status field to the product and configurable product variants to determine if the product is out of stock. There is no stock_status field for magento bundle product item options for some reason, even though they're just products.


libs/product/src/facades/product-grid/product-grid.facade.spec.ts
```diff
@@ -61,7 +63,7 @@ describe('DaffProductGridFacade', () => {
     });
 
     it('should be an observable of the list of products', () => {
-      const products = [{id: '1', name: 'Some Name'}];
```
lderrickable, line 64: Originally, I was going to make the `stock_status` field required, so I fixed these tests. But then I realized that the MagentoBundleProductItemOptions (which extend DaffProduct) wouldn't have a stock status, so `stock_status` needed to be optional. And I just kept this fix instead of reverting it.

libs/product/src/drivers/magento/models/magento-product.ts
```diff
@@ -33,6 +38,7 @@ export interface MagentoProduct {
 			}
 		}
 	};
+	stock_status?: string;
```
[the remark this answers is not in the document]
lderrickable: yea, it should. I had to make some changes to some tests to get it to work, because typescript complains about getting a string from a json object instead of the enum.

libs/product/src/drivers/magento/transforms/bundled-product-transformers.spec.ts
```diff
@@ -12,9 +13,13 @@ describe('DaffMagentoBundledProductTransformers', () => {
   });
 
 	describe('transform', () => {
+		const magentoBundledProduct = {
+			...magentoBundledProductData,
+			stock_status: MagentoProductStockStatusEnum.InStock
```
griest024, line 18: I don't think these test changes are necessary.

libs/product/src/drivers/magento/models/magento-product.ts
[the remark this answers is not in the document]
griest024: oh it doesn't do that for me. It builds and passes the tests without the changes though, is that also what you see?

libs/product/src/drivers/magento/transforms/bundled-product-transformers.spec.ts
lderrickable: like that?

griest024: whoa thats odd, it didn't fail on my machine

griest024: my bad, I guess those test changes are necessary

lderrickable: np, changed back

PR #1045, opened 2020-08-25
feat(core): add GraphQL fragment utils

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Adds a `daffGetFragmentNames` function for reading fragment names from a graphql document node.
- Adds a `daffBuildFragmentNameSpread` for formatting fragment names in such a way that they can easily stuck into a GraphQL query.
- Adds a `daffBuildFragmentDefinition` for formatting fragment bodies in such a way that they can easily stuck into a GraphQL query.

```
[ ] Yes
[x] No
```

## Other information


griest024:
Do you think its worth doing any type checking on the fragments in `daffGetFragmentNames`? We could filter out any defined fragments that don't match the specified type. That way we can guarantee that all fragments being applied to the `Cart` object are actually on type `Cart`, for example.

libs/core/src/graphql/get-fragment-names.spec.ts
```diff
@@ -0,0 +1,63 @@
+import { DocumentNode } from 'graphql';
+import gql from 'graphql-tag';
+import { daffGetFragmentNames } from './get-fragment-names';
+
+describe('Core | GraphQL | daffGetFragmentNames', () => {
```
damienwebdev, line 5: Can you provide an integration test of how this would be used to merge fragments into queries?

griest024: done

lderrickable:
> Do you think its worth doing any type checking on the fragments in `daffGetFragmentNames`? We could filter out any defined fragments that don't match the specified type. That way we can guarantee that all fragments being applied to the `Cart` object are actually on type `Cart`, for example.

I don't think we should. If we check types, then we should probably throw an error instead of just silently filtering the bad ones out. And graphql throws errors anyway if the type isn't right, so I think we're ok not to

griest024:
@lderrickable 

> I don't think we should. If we check types, then we should probably throw an error instead of just silently filtering the bad ones out. And graphql throws errors anyway if the type isn't right, so I think we're ok not to

That seems reasonable.

libs/core/src/graphql/fragment.integration.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+  });
+
+  describe('adding fields to a query', () => {
+    let query;
+
+    beforeEach(() => {
+      query = gql`
+        query {
+          id
+          ${daffBuildFragmentSpread(daffGetFragmentNames(mockFragment))}
```
damienwebdev, line 49: I think this API is a bit clunky, I'd like to only have to pass a mockFragment, not a nested function.

libs/core/src/graphql/build-fragment-spread.ts
```diff
@@ -0,0 +1,10 @@
+import { FragmentDefinitionNode } from 'graphql'
+
+/**
+ * Builds a string of fragment names that can be interpolated into a GraphQL query.
+ * Each name is separated by a newline character: '\n'.
+ * If an empty array is passed, an empty string is returned.
+ * @param names A list of fragment names.
+ */
+export const daffBuildFragmentSpread = (names: FragmentDefinitionNode['name']['value'][]): string =>
```
damienwebdev, line 9: This arg type should likely be expanded to `FragmentDefinitionNode[]`

griest024: okay but I think I'll make it a rest param so that single fragments can be passed without having to create an array.

griest024: changed

libs/core/src/graphql/fragment.integration.spec.ts
griest024: changed, see integration tests for final usage

griest024:
@damienwebdev removed spread in integration spec

damienwebdev:
@lderrickable can you give this a once over before merge?

libs/core/src/graphql/get-fragment-names.ts
```diff
@@ -0,0 +1,17 @@
...
+  ).map(def =>
+    (def as FragmentDefinitionNode).name.value
+  )
+
+/**
+ * Builds a list of fragment names present inside the specified GraphQL document nodes.
+ * Returns an empty array if no fragments have been defined or if null is passed.
+ * @param fragments The created fragments.
+ */
+export const daffGetFragmentNames = (...fragments: DocumentNode[]): string[] =>
```
lderrickable, line 15: Will this ever be used outside of the `daffBuildFragmentNameSpread` function? I can't think of when it would be, and if that's the case, I don't think it should be a part of the public api. I think it should be moved into the same file as `daffBuildFragmentNameSpread`

lderrickable: lgtm otherwise

griest024: I can't think of any use case at the moment. I feel like getting the names of the fragments present in a document definition is a useful utility function to have but we can always easily export it again if we need it. YAGNI for now.

griest024: changed

lderrickable: Probably move the tests too, right? Save time later if it needs moved out

griest024: nah we don't need to test a function that isn't exported. It'll be part of this PR's diff if we really need it.

PR #1057, opened 2020-09-04
fix(cart): not all available shipping methods returned by update

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`DaffMagentoCartShippingInformationService#update` only returns the selected shipping method for the available shipping methods.

## What is the new behavior?
`DaffMagentoCartShippingInformationService#update` manually requests the available shipping methods while bypassing the cache after performing the update mutation. This returns all of the available shipping methods.

```
[ ] Yes
[x] No
```

## Other information
This is a Magento specific bug and fix.

### DIFF
--- a/libs/cart/src/drivers/magento/cart-shipping-information.service.spec.ts
+++ b/libs/cart/src/drivers/magento/cart-shipping-information.service.spec.ts
@@ -230,7 +230,7 @@ describe('Driver | Magento | Cart | CartShippingInformationService', () => {
         data: mockSetSelectedShippingMethodResponse
       });
 
-      // set timeout because the requests here are made in serial
+      // set timeout because the requests here are made in series
       setTimeout(() => {
         const listShippingMethodsOp = controller.expectOne(addTypenameToDocument(listShippingMethods([daffMagentoNoopCartFragment])));
 
--- a/libs/cart/src/drivers/magento/cart-shipping-information.service.spec.ts
+++ b/libs/cart/src/drivers/magento/cart-shipping-information.service.spec.ts
@@ -23,11 +23,12 @@ import { MagentoCartShippingMethod } from './models/outputs/cart-shipping-method
 import { DaffMagentoCartTransformer } from './transforms/outputs/cart.service';
 import { MagentoGetSelectedShippingMethodResponse } from './models/responses/get-selected-shipping-method';
 import { MagentoSetSelectedShippingMethodResponse } from './models/responses/set-selected-shipping-method';
-import { getSelectedShippingMethod, setSelectedShippingMethod } from './queries/public_api';
+import { getSelectedShippingMethod, setSelectedShippingMethod, listShippingMethods } from './queries/public_api';
 import { DaffMagentoCartShippingRateTransformer } from './transforms/outputs/cart-shipping-rate.service';
 import { DaffMagentoShippingMethodInputTransformer } from './transforms/inputs/shipping-method.service';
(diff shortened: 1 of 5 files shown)


libs/cart/src/drivers/magento/cart-shipping-information.service.spec.ts
```diff
@@ -212,11 +224,40 @@ describe('Driver | Magento | Cart | CartShippingInformationService', () => {
...
+      });
+
+      // set timeout because the requests here are made in serial
+      setTimeout(() => {
+        const listShippingMethodsOp = controller.expectOne(addTypenameToDocument(listShippingMethods([daffMagentoNoopCartFragment])));
+
+        listShippingMethodsOp.flush({
+          data: mockListCartShippingMethodsResponse
+        });
+        console.log(listShippingMethodsOp);
```
lderrickable, line 259: console.log

libs/cart/src/drivers/magento/cart-shipping-information.service.spec.ts
```diff
@@ -212,11 +224,40 @@ describe('Driver | Magento | Cart | CartShippingInformationService', () => {
...
+      setTimeout(() => {
+        const listShippingMethodsOp = controller.expectOne(addTypenameToDocument(listShippingMethods([daffMagentoNoopCartFragment])));
+
+        listShippingMethodsOp.flush({
+          data: mockListCartShippingMethodsResponse
+        });
+      })
+    });
+
+    it('should manually refetch the shipping methods and force cache bypass', () => {
```
lderrickable, line 243: This test doesn't actually register that there are any expectations, so it counts as a skipped test. Probably because of setTimeout

griest024: Its because the actual checking happens in `afterEach`. Unfortunately there's not really a good way to check that the request bypasses the cache so I removed that spec since its not doing anything more than the other one.

libs/cart/src/drivers/magento/cart-shipping-information.service.spec.ts
griest024: good catch, removed

libs/cart/src/drivers/magento/queries/list-shipping-methods.ts
```diff
@@ -8,6 +8,7 @@ import { availableShippingMethodFragment } from './fragments/public_api';
 export const listShippingMethods = (extraCartFragments: DocumentNode[] = []) => gql`
   query ListShippingMethods($cartId: String!) {
     cart(cart_id: $cartId) {
+      id
```
lderrickable, line 11: just curious. Why was the id added to the query?

libs/cart/src/drivers/magento/cart-shipping-information.service.spec.ts
lderrickable: that's fine with me

libs/cart/src/drivers/magento/queries/list-shipping-methods.ts
griest024: Apollo would refuse to write the query result to the store without an ID. I'm not sure why it doesn't do this under any other circumstances.

lderrickable (approved):

PR #1061, opened 2020-09-06
In Stock field for products and cart items

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The stock status fields are currently enums.

## What is the new behavior?
The stock status fields are now in stock boolean fields.

```
[x] Yes
[ ] No
```

## Other information
I just realized that there's no reason for the enum, since there are only two states.

### DIFF
--- a/libs/cart/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
+++ b/libs/cart/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
@@ -113,7 +113,9 @@ const createCartItemEntitiesSelectors = <
 	const selectIsCartItemOutOfStock = createSelector(
 		selectCartItemEntities,
 		(cartItems, props) => {
-			return !selectCartItem.projector(cartItems, { id: props.id }).in_stock;
+			const cartItem = selectCartItem.projector(cartItems, { id: props.id });
+
+			return cartItem ? !cartItem.in_stock : null;
 		}
 	);
 
--- a/libs/cart/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
+++ b/libs/cart/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
@@ -1,7 +1,7 @@
 import { createSelector, MemoizedSelector, MemoizedSelectorWithProps } from '@ngrx/store';
 import { EntityState } from '@ngrx/entity';
 
-import { DaffCartItem, DaffCartItemStockEnum } from '../../models/cart-item';
+import { DaffCartItem } from '../../models/cart-item';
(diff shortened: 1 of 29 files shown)


griest024:
@lderrickable I wonder if `stock_status` really needs to be an enum. Do you think its worth treating back orders as a separate stock status?

lderrickable:
I think that would only be on orders? I don't think back orders would ever touch products or carts. You'd need to submit an order for products that are out of stock.

griest024:
Well sometimes when online shopping I'll see a product listed as on back-order on the product page. So you can add it to your cart and place an order for it but it won't ship right away.

I think we might need it in the far future but I guess YAGNI for now.

lderrickable:
oh, that's true. Might be good to get @damienwebdev into this conversation. He might be more familiar with how magento 1 handled this, which will probably help

griest024:
@lderrickable We should treat it as YAGNI as per the twist thread.

libs/product/src/selectors/product-entities/product-entities.selectors.ts
```diff
@@ -97,8 +97,8 @@ const createProductEntitiesSelectors = <T extends DaffProduct>(): DaffProductEnt
 	const selectIsProductOutOfStock = createSelector(
 		selectProductEntities,
 		(products, props) => {
-			return selectProduct.projector(products, { id: props.id })
-				.stock_status === DaffProductStockEnum.OutOfStock;
+			return !selectProduct.projector(products, { id: props.id })
```
griest024, line 100: this can be a lambda

griest024 (changes requested):

libs/cart/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
```diff
@@ -113,8 +113,7 @@ const createCartItemEntitiesSelectors = <
 	const selectIsCartItemOutOfStock = createSelector(
 		selectCartItemEntities,
 		(cartItems, props) => {
-			return selectCartItem.projector(cartItems, { id: props.id })
-				.stock_status === DaffCartItemStockEnum.OutOfStock;
+			return !selectCartItem.projector(cartItems, { id: props.id }).in_stock;
```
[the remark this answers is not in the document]
lderrickable: I actually think this should be null checked, so I added that instead

libs/product/src/selectors/product-entities/product-entities.selectors.ts
lderrickable: same here

griest024 (approved):

PR #1065, opened 2020-09-14
feat(cart): add platform agnostic payment method ID facade field

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Adds an injection token for provides payment method maps.
- Adds a `paymentId$` facade field for mapping the cart payment method to the user provided payment ID.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/src/facades/cart/cart.facade.spec.ts
+++ b/libs/cart/src/facades/cart/cart.facade.spec.ts
@@ -49,7 +49,7 @@ import { DaffCartOrderResult } from '../../models/cart-order-result';
 import { DaffConfigurableCartItem } from '../../models/configurable-cart-item';
 import { DaffCompositeCartItem } from '../../models/composite-cart-item';
 import { DaffResolveCartSuccess } from '../../actions/public_api';
-import { DaffCartPaymentMethodMap } from '../../injection-tokens/public_api';
+import { DaffCartPaymentMethodIdMap } from '../../injection-tokens/public_api';
 
 describe('DaffCartFacade', () => {
   let store: MockStore<{ product: Partial<DaffCartReducersState> }>;
@@ -77,7 +77,7 @@ describe('DaffCartFacade', () => {
       providers: [
         DaffCartFacade,
         {
-          provide: DaffCartPaymentMethodMap,
+          provide: DaffCartPaymentMethodIdMap,
           useValue: {
             [paymentMethod]: paymentId
           }
--- a/libs/cart/src/facades/cart/cart.facade.spec.ts
+++ b/libs/cart/src/facades/cart/cart.facade.spec.ts
@@ -27,7 +27,9 @@ import {
   DaffCartPlaceOrderFailure,
   DaffCartPlaceOrderSuccess,
   DaffCartCouponListFailure,
-	DaffCartTotalTypeEnum
+  DaffCartTotalTypeEnum,
+  DaffCartPaymentMethod,
+  DaffCart,
 } from '@daffodil/cart';
 import {
   DaffCartFactory,
@@ -43,11 +45,11 @@ import { DaffCartFacade } from './cart.facade';
 import { DaffCartReducersState, daffCartReducers, initialState } from '../../reducers/public_api';
 import { DaffCartErrors } from '../../reducers/errors/cart-errors.type';
 import { DaffCartErrorType } from '../../reducers/errors/cart-error-type.enum';
-import { DaffCart } from '../../models/cart';
 import { DaffCartOrderResult } from '../../models/cart-order-result';
 import { DaffConfigurableCartItem } from '../../models/configurable-cart-item';
 import { DaffCompositeCartItem } from '../../models/composite-cart-item';
 import { DaffResolveCartSuccess } from '../../actions/public_api';
+import { DaffCartPaymentMethodMap } from '../../injection-tokens/public_api';
 
 describe('DaffCartFacade', () => {
   let store: MockStore<{ product: Partial<DaffCartReducersState> }>;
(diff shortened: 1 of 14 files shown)


libs/cart/src/facades/cart/cart.facade.ts
```diff
@@ -179,6 +185,14 @@ export class DaffCartFacade<
     this.shippingInformation$ = this.store.pipe(select(selectCartShippingInformation));
     this.availableShippingMethods$ = this.store.pipe(select(selectCartAvailableShippingMethods));
     this.availablePaymentMethods$ = this.store.pipe(select(selectCartAvailablePaymentMethods));
+    this.paymentId$ = this.payment$.pipe(
+      filter(payment => !!(payment && payment.method)),
+      map(payment => {
+        console.log(this.paymentMethodMap, payment.method);
```
lderrickable, line 191: you've got some extra stuff here

libs/cart/src/injection-tokens/payment-method-map.ts
```diff
@@ -0,0 +1,11 @@
+import { InjectionToken } from '@angular/core';
+
+/**
+ * An injection token for a mapping from the platform-specific cart payment method
+ * to a user-defined platform-agnostic payment ID.
+ * It should be an object whose keys are the cart payment's method.
+ * Defaults to an empty object.
+ */
+export const DaffCartPaymentMethodMap = new InjectionToken<Object>('DaffCartPaymentMethodMap', {
```
lderrickable, line 9: This should be `DaffCartPaymentMethodIdMap`, right? To be a little more clear to the user

lderrickable (changes requested):

libs/cart/src/facades/cart/cart.facade.ts
griest024: ah good catch, thanks

griest024: changed

libs/cart/src/injection-tokens/payment-method-map.ts
griest024: sounds good, changed

libs/cart/src/facades/cart/cart.facade.spec.ts
```diff
@@ -559,6 +569,81 @@ describe('DaffCartFacade', () => {
...
+        store.dispatch(new DaffCartLoadSuccess({
+          ...cart,
+          payment: {
+            ...mockPayment,
+            method: paymentMethod
+          }
+        }));
+      });
+
+      it('should return the mapped payment ID', () => {
```
lderrickable, line 640: I think this should be "should return the platform agnostic payment ID"

libs/cart/src/facades/cart/cart.facade.ts
```diff
@@ -81,7 +84,10 @@ export class DaffCartFacade<
 	private _selectCartItemCompositeOptions;
 	private _selectIsCartItemOutOfStock;
 
-  constructor(private store: Store<DaffCartReducersState<T, V, U>>) {
+  constructor(
+    private store: Store<DaffCartReducersState<T, V, U>>,
+    @Inject(DaffCartPaymentMethodIdMap) private paymentMethodMap: Object
```
lderrickable, line 89: This should be typed as `Dictionary<string>`

libs/cart/src/facades/cart/cart.facade.ts
```diff
@@ -179,6 +185,10 @@ export class DaffCartFacade<
     this.shippingInformation$ = this.store.pipe(select(selectCartShippingInformation));
     this.availableShippingMethods$ = this.store.pipe(select(selectCartAvailableShippingMethods));
     this.availablePaymentMethods$ = this.store.pipe(select(selectCartAvailablePaymentMethods));
+    this.paymentId$ = this.payment$.pipe(
```
lderrickable, line 188: I think this should return null if `!!(payment && payment.method)`. Is there a reason it doesn't return anything instead?

lderrickable (changes requested):

libs/cart/src/facades/cart/cart.facade.ts
griest024: I don't think we can guarantee that the user will map the methods to strings.

libs/cart/src/facades/cart/cart.facade.ts
griest024: Nah no particular reason. changed

libs/cart/src/facades/cart/cart.facade.spec.ts
griest024: changed

libs/cart/src/facades/cart/cart.facade.ts
lderrickable: That's true, but only because we can't strongly type the injection token. Maybe we should type the actual injection token as `Dictionary<string>` in case someone looks at the type there so that they are encouraged to map to payment ids, even though that typing will not be enforced. I'm just thinking that this feature is pretty complicated and we'll probably need a doc just to explain why this is necessary, so I think it's best to give as many helpful hints as possible. And I think typing the injection token as `Dictionary<string>` instead of `Object` is one such hint. But I agree that here in the facade it should probably just remain an object, since we can't guarantee it yet.

griest024: hm okay I can retype the token. The only thing that is odd to me is that `Dictionary` is specifically an `@ngrx/entity` type, I wonder if it might be better to instead type it as `{[key: string]: string}`.

lderrickable: I think that's fine, since we're dependent on @ngrx/entity anyway, but `{[ket: string]: string}` seems good too

griest024: changed

lderrickable (approved):

PR #1070, opened 2020-09-18
feat(product): update the composite product option price and add disc…

lderrickable:
…ount

## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
The magento composite product option prices need to come from `option.product.price_range` instead of `option.price`.

### DIFF
--- a/libs/product/src/drivers/magento/transforms/bundled-product-transformers.ts
+++ b/libs/product/src/drivers/magento/transforms/bundled-product-transformers.ts
@@ -1,5 +1,5 @@
 import { MagentoBundledProduct, MagentoBundledProductItem, MagentoBundledProductItemOption } from '../models/bundled-product';
-import { DaffProductTypeEnum } from '../../../models/product';
+import { DaffProductDiscount, DaffProductTypeEnum } from '../../../models/product';
 import { DaffCompositeProduct } from '../../../models/composite-product';
 import { 
 	DaffCompositeProductItemOption, 
@@ -7,7 +7,7 @@ import {
 	DaffCompositeProductItemInputEnum 
 } from '../../../models/composite-product-item';
 import { transformMagentoSimpleProduct } from './simple-product-transformers';
-import { MagentoProductStockStatusEnum } from '../models/magento-product';
+import { MagentoProduct, MagentoProductStockStatusEnum } from '../models/magento-product';
 
 /**
  * Transforms the magento MagentoProduct from the magento product query into a DaffProduct. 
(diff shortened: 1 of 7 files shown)


libs/product/src/drivers/magento/transforms/bundled-product-transformers.ts
```diff
@@ -35,9 +35,31 @@ function transformMagentoBundledProductItemOption(option: MagentoBundledProductI
...
+function getPrice(product: MagentoProduct): number {
+	return product.price_range && 
+		product.price_range.maximum_price && 
+		product.price_range.maximum_price.regular_price && 
+		product.price_range.maximum_price.regular_price.value !== null
+	? product.price_range.maximum_price.regular_price.value : null;
+}
+
+function getDiscount(product: MagentoProduct): DaffProductDiscount {
+	return product.price_range &&
```
griest024, line 58: Can you add `TODO: use optional chaining` comments to these functions? That will make it easier to find this code for refactoring once we upgrade to Angular 9.

libs/product/src/drivers/magento/transforms/bundled-product-transformers.ts
```diff
@@ -35,9 +35,31 @@ function transformMagentoBundledProductItemOption(option: MagentoBundledProductI
...
+}
+
+function getDiscount(product: MagentoProduct): DaffProductDiscount {
+	return product.price_range && 
+		product.price_range.maximum_price && 
+		product.price_range.maximum_price.discount 
+		? { 
+			amount: product.price_range.maximum_price.discount.amount_off,
+			percent: product.price_range.maximum_price.discount.percent_off
+		} : { amount: null, percent: null }
```
griest024, line 64: Do you think these values should be initialized to 0? I feel like they should, but that is also probably handled in the state layer anyway.

griest024 (changes requested):

libs/product/src/drivers/magento/transforms/bundled-product-transformers.ts
lderrickable: I think it should be null, since it is passed to daffodil as null and not 0

libs/product/src/drivers/magento/transforms/bundled-product-transformers.ts
lderrickable: added

griest024 (approved):

PR #1071, opened 2020-09-21
fix(cart): a null shipping total should be null instead of zero

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The shipping total in daffodil is set to zero even if the total from magento is null. This is a problem, because the shipping total needs to be shown as "FREE" when the shipping total is zero. So the shipping will be shown as "FREE" even though a shipping method hasn't been chosen yet.

## What is the new behavior?
The shipping total will be null when the shipping method hasn't been chosen yet.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/cart/src/drivers/magento/transforms/outputs/cart-totals-transformer.spec.ts
+++ b/libs/cart/src/drivers/magento/transforms/outputs/cart-totals-transformer.spec.ts
@@ -172,7 +172,7 @@ describe('transformCartTotals', () => {
 				{
 					name: DaffCartTotalTypeEnum.shipping,
 					label: 'Shipping',
-					value: 0
+					value: null
 				}
 			]
 		};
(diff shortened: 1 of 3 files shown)


griest024 (dismissed):

damienwebdev (changes requested):
Would it be better to instead return an empty object for the total, and then throw an error here? A null total is a bit odd, I'd anticipate it being numeric.

lderrickable:
I don't think so. Because we'd be throwing an error any time that there is no shipping method defined. I think it makes sense for the shipping total to be numeric or null, because a value of zero is not the same as it not being provided. In other words, we don't have a way to differentiate between no shipping method selected and "Free Shipping". Actually, I think tax should be handled this way too, because we don't currently have a way to differentiate between no tax and zero tax either.

damienwebdev:
Needs docs, but I'm fine with the behavior.

lderrickable:
added docs

griest024 (approved):

damienwebdev (approved):
lgtm

PR #1080, opened 2020-09-30
feat(cart, magento): removes shipping methods from main cart fragement

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Magento's GraphQl query for Shipping Methods doesn't cache the results of rate requests for providers. This means that Magento cart requests are obnoxiously slow. See https://github.com/magento/magento2/issues/28337

Fixes: N/A

## What is the new behavior?
This removes the shipping methods from a typical cart load, they must now be loaded explicitly.

```
[x] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/cart/src/drivers/magento/queries/fragments/cart.ts
+++ b/libs/cart/src/drivers/magento/queries/fragments/cart.ts
@@ -5,7 +5,6 @@ import { availablePaymentMethodFragment } from './available-payment-method';
 import { selectedPaymentMethodFragment } from './selected-payment-method';
 import { cartItemFragment } from './cart-item';
 import { cartCouponFragment } from './cart-coupon';
-import { availableShippingMethodFragment } from './available-shipping-method';
 import { selectedShippingMethodFragment } from './selected-shipping-method';
 import { pricesFragment } from './prices';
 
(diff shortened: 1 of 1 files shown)


lderrickable:
this is a breaking change, right?

damienwebdev:
@lderrickable yes. I'll sign the commit as such.

lderrickable:
I think this will cause errors in the cart transformers. There are places where `cart.available_payment_methods.map` is called, and this will blow up when `cart.available_payment_methods` is null/undefined

damienwebdev:
@lderrickable I think the [transformers are ok](https://github.com/graycoreio/daffodil/blob/develop/libs/cart/src/drivers/magento/transforms/outputs/cart.service.ts#L90). We will simply return an empty array.

But, this will lead to an issue with the state where unless you're explicitly requesting methods, we won't load them and subsequent calls will remove them. Though, perhaps this is a feature.

lderrickable:
oops, yea you're right. I was looking at payment stuff.

lderrickable (approved):

damienwebdev:
Awaiting @griest024 review as well.

griest024:
Do you think its worth adding these fields back for the queries that we _know_ will affect the shipping methods, for example, set shipping address?

damienwebdev:
@griest024 yes. that's a good idea, I'll make the change.

damienwebdev:
In retrospect @griest024 I don't think we should. The performance of this request is too variable and I think needs to be siloed into its own mechanism.

griest024 (approved):

Issue #1082, opened 2020-10-01
[FEAT] Improve Loading Behaviors on Cart State


xelaint:
# :bulb: Feature request

## Feature Name
 Improve Loading Behaviors on Cart State

## The Desired Behavior
Resolution vs. Mutation vs. Loading
- Item
- Address
- Billing Address
- Shipping Address
- Shipping Methods
- Currently selected shipping method
- Payment methods
- Currently selected payment
- Promo code

## Your Use Case
As a designer, I'd like to be able to specify the loading UI more precisely.

Some Examples:
- Updating the quantity of a cart item should not trigger the loading state for the promo code container.
- Selecting a different shipping option does not need to trigger a load state on the whole cart summary container. The shipping line and totals line should be the only parts that show the 'ghost' state or a variation of it.
- Saving your address does not need to trigger a load state on the whole cart summary container.
- Removing an item from cart does not need to trigger load state on whole cart (only on item being removed).
- Adding a promo code to cart does not need to trigger a cart item's load state.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

griest024:
I'll take this one.

@xelaint When you say
> Resolution vs. Mutation vs. Loading

Does that mean that you want different loading states for different operations? So updating the shipping address would be different than just fetching the shipping address?

Does the cart as a whole still have a loading state? Would that cause all the cart subfield loading states to also be loading? For example, triggering an entire cart load will fetch the shipping address; would that cause the shipping address loading state to be true in addition to the main cart loading state?

PR #1084, opened 2020-10-03
feat(core, cart): segregate loading state into cart subfields

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #1082 

```
[x] Yes
[ ] No
```

The `loading$` facade field and the `selectCartLoading` selector have slightly different behavior than before. They now only track the loading state for the actions that specifically refer to the cart as a whole.

## Other information
This is a pretty big PR. LMK if I should break it up.

Note that the mutating selectors and facade fields were not added for the sections that don't have mutations such as shipping or payment methods.


lderrickable:
could you explain what the use for Resolve vs Mutate loading states would be for an app?

griest024:
@lderrickable I'm not exactly sure, maybe @xelaint could elaborate? The issue specifically requested `Resolution vs. Mutation vs. Loading`.

lderrickable:
@griest024 @xelaint Oh I saw the examples in the issue. 

@griest024 I think the code here makes sense (still reviewing), but it won't work in all cases. For example, the grand total will change on almost any cart state change. So in the app, we would need to have a really long "or" check for any of the cart loading states that are applicable. Like `cartLoading$ || shippingMethodLoading$ || cartItemLoading$ || cartShippingAddressLoading$ || cartCouponLoading$ || cartBillingAddressLoading$`. To cover this, I think we need a selector that is in a resolving state if anything in the feature area is resolving or mutating.

libs/cart/src/selectors/cart/cart.selector.ts
```diff
@@ -20,9 +20,132 @@ export interface DaffCartStateMemoizedSelectors<
...
-  selectCartLoading: MemoizedSelector<object, boolean>;
   selectCartResolved: MemoizedSelector<object, boolean>;
 
+  /**
+   * The object that holds all the loading states for cart operations.
+   */
+  selectCartLoadingObject: MemoizedSelector<object, DaffCartReducerState<T>['loading']>;
+  /**
+   * Selects whether there is a cart operation in progress.
+   * This does not include operations specifally for cart subfields.
```
lderrickable, line 31: Maybe this should cover all operations for cart? Like it returns true if any subfields of cart are Resolving or Mutating. I can't think of a use case where the entire cart is loading, and we wouldn't want a subfield of the cart to be loading.

libs/cart/src/facades/cart/cart-facade.interface.ts
```diff
@@ -6,31 +6,156 @@ import { DaffStoreFacade } from '@daffodil/core';
...
   resolved$: Observable<boolean>;
   cart$: Observable<T>;
+
+  /**
+   * The object that holds all the loading states for cart operations.
+   */
+  loadingObject$: Observable<DaffCartLoading>;
+  /**
+   * Whether there is a cart operation in progress.
+   * This does not include operations specifally for cart subfields.
```
lderrickable, line 31: "specifically" typo

libs/cart/src/selectors/cart/cart.selector.ts
```diff
@@ -20,9 +20,132 @@ export interface DaffCartStateMemoizedSelectors<
...
+   */
+  selectCartLoading: MemoizedSelector<object, boolean>;
+  /**
+   * Selects whether there is a cart resolve operation in progress.
+   * This does not include operations specifally for cart subfields.
```
lderrickable, line 36: "specifically" typo

libs/cart/src/selectors/cart/cart.selector.ts
```diff
@@ -20,9 +20,132 @@ export interface DaffCartStateMemoizedSelectors<
...
+  /**
+   * Selects whether there is a cart billing address resolve operation in progress.
+   * This pertains only to requests that do not mutate data such as "load" or "list".
+   */
+  selectBillingAddressResolving: MemoizedSelector<object, boolean>;
+  /**
+   * Selects whether there is a cart billing address mutate operation in progress.
+   * This pertains only to requests that mutate data such as "update".
+   */
+  selectBillingAddressMutating: MemoizedSelector<object, boolean>;
```
lderrickable, line 59: I think that these selectors should probably be named `selectCartBillingAddressMutating` instead of `selectBillingAddressMutating`, because of how they are sometimes imported into a class. Example:
```
const {
    selectBillingAddressMutating
} = getDaffCartSelectors();
```
This could create confusion and conflicts in the class

griest024: The local var could always be aliased during destructuring to avoid conflicts:
```ts
const {
    selectBillingAddressMutating: selectCartBillingAddressMutating
} = getDaffCartSelectors();
```

As far as clarity, I think its probably fine either way. I was just following the pattern for the error selectors. If you think this should still be changed then the error selectors should also get the `Cart` prefix.

libs/cart/src/facades/cart/cart-facade.interface.ts
griest024: whoops. fixed

libs/cart/src/selectors/cart/cart.selector.ts
griest024: fixed

libs/cart/src/selectors/cart/cart.selector.ts
griest024: Either that or we add another selector that tracks if any of the feature areas are loading. I kinda like having a disctinction between "are any of the cart fields loading" and "are all of the cart fields loading".

libs/cart/src/selectors/cart/cart.selector.ts
lderrickable: No the aliasing sounds good to me

libs/cart/src/selectors/cart/cart.selector.ts
lderrickable: That sounds fine to me, as long as the names makes sense

griest024: How about I use `selectCartFeatureLoading` to refer to the "are any of the cart fields loading"?

lderrickable: sure, sounds good

griest024: added

libs/cart/src/facades/cart/cart.facade.spec.ts
```diff
@@ -182,6 +182,336 @@ describe('DaffCartFacade', () => {
...
+      });
+
+      it('should return true', () => {
+        const expected = cold('a', {a: true});
+
+        expect(facade.featureLoading$).toBeObservable(expected);
+      });
+    });
+
+    describe('when the cart item operations have not completed', () => {
```
lderrickable, line 206: Could you remove all of the subfield tests? It is already tested in the selectors, so it's all just extra here.

griest024: It also tests the "are any" logic but I'll remove it if you really want me to.

libs/cart/src/selectors/cart/cart.selector.ts
```diff
@@ -10,11 +10,21 @@ import { daffComparePersonalAddresses } from '@daffodil/geography';
...
-import { DaffCartReducerState, DaffCartReducersState, DaffCartOperationType } from '../../reducers/public_api';
+import { DaffCartReducerState, DaffCartReducersState, DaffCartOperationType, DaffCartLoading } from '../../reducers/public_api';
 import { DaffCartOrderResult } from '../../models/cart-order-result';
 import { DaffCartItem } from '../../models/cart-item';
 import { DaffCartTotalTypeEnum, DaffCartTotal } from '../../models/cart-total';
 
+/**
+ * Util function to project loading selectors
+ */
+function projectAllLoadingSelectors(
```
lderrickable, line 21: was this going to be for the selectors below, then you ended up not using it?

lderrickable: put another way, I'm just not sure what this is for

griest024: whoops yeah it was

griest024: removed

libs/cart/src/facades/cart/cart.facade.spec.ts
lderrickable: Yea, but that logic is actually done and tested in the selectors, not the facade. Really, we just want to verify that the facade is giving the value of the selector we're setting it to. We don't need to retest all of the selector logic.

griest024: ah my bad you were specifically talking about the facade tests. I should have understood. removed the extra facade test cases.

lderrickable (approved):

PR #1092, opened 2020-10-13
docs(cart): add cart guides

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[x] No
```

## Other information


docs/advanced/fragment-introspection.md
```diff
@@ -0,0 +1,99 @@
+# GraphQL Fragment Introspection
```
damienwebdev, line 1: This doc should be in the guides of the builders.

libs/cart/README.md
```diff
@@ -3,9 +3,30 @@
 Building and maintaining a model and code for an ecommerce store is complex and mentally taxing. `@daffodil/cart`
 provides clear interfaces, models, and factories for the frontend of an ecommerce store so that you don't have to.
 
-
 ## Installation
 
-```
-npm install @daffodil/cart
-```
+Refer to [the installation guide](./guides/installation).
```
damienwebdev, line 8: These links don't seem to work. https://github.com/griest024/daffodil/blob/docs/cart/guides/libs/cart/README.md

libs/cart/README.md
```diff
@@ -3,9 +3,30 @@
...
+
+## Extension
+
+Refer to [the extension guide](./guides/extension).
+
+## Testing
+
+Refer to [the testing guide](./guides/testing).
+
+## Live Demo
```
damienwebdev, line 26: As we don't have this yet, please remove.

libs/cart/README.md
```diff
@@ -3,9 +3,30 @@
...
+
+[Checkout a live example of the `DaffCart` library in action!]()
+
+## Supported Drivers
```
damienwebdev, line 30: Move to a `drivers` doc.

libs/cart/guides/installation.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Installing with `yarn`
+
+To install the cart library and its dependencies, use the following command in your terminal.
+
+```bash
+yarn add @daffodil/cart @daffodil/core @ngrx/store @ngrx/effects
+```
+
+## Post-Install
```
damienwebdev, line 32: We should clarify what you're supposed to actually do post-install. Some people may only want `state` some people may only want `driver` and some people may only want a specific driver.

libs/cart/guides/installation.md
```diff
@@ -0,0 +1,46 @@
...
+
+After installing, an ecommerce platform driver needs to be set-up. We highly recommend installing the [In-Memory Web API](#in-memory-web-api) for fast, out-of-the-box development.
+
+## In-Memory Web API
```
damienwebdev, line 36: The docs below here doc feel out of place, consider moving to the driver's step.

libs/cart/guides/getting-started.md
```diff
@@ -0,0 +1,30 @@
...
+The `DaffCartModule` provides a `DaffCartFacade` that wraps the complexities of the state library into one place. This facade will handle updating the user's cart and can also be utilized to build your UI with behaviors common to a cart.
+
+To inject the facade inside your component, include an instance of `DaffCartFacade` in your component's constructor.
+
+```typescript
+@Component({})
+export class CartComponent {
+  constructor(public cartFacade: DaffCartFacade) {}
+}
+```
```
damienwebdev, line 30: Now what, we should make sure people know where to go next.

damienwebdev (changes requested):
I think a general driver's doc would be helpful as well.

docs/advanced/storage.md
```diff
@@ -0,0 +1,51 @@
+# Storage
```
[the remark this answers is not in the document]
lderrickable: I think it makes the most sense to have it as an overarching doc for all of daffodil rather than have it reside in core. People might not find it easily if it is in core.

libs/cart/README.md
griest024: hm, they worked in the VSCode markdown preview. I might need to add the file extension

libs/cart/guides/drivers/in-memory.md
```diff
@@ -0,0 +1,23 @@
+# Cart In-Memory Web API Driver
+
+The In-Memory driver is for rapid development without the need to set up a magento/shopify/etc backend. It will mock out the management of a cart and operate like a functional backend. It is intended for development and testing purposes; it is not meant to be used in production.
+
+To set up, import the `DaffCartInMemoryDriverModule` from the `@daffodil/cart/testing` library and the `HttpClientInMemoryWebApiModule` from `angular-in-memory-web-api`.
+ Include `DaffCartInMemoryDriverModule.forRoot()` and `HttpClientInMemoryWebApiModule` in the imports section of your `app.module`.
+
+```typescript
+import { HttpClientInMemoryWebApiModule } from 'angular-in-memory-web-api';
+import { DaffCartInMemoryDriverModule } from '@daffodil/Cart/testing';
```
lderrickable, line 10: typo "Cart" -> "cart"

libs/cart/guides/drivers/magento.md
```diff
@@ -0,0 +1,28 @@
+# Cart Magento Driver
+
+The In-Memory driver is for rapid development without the need to set up a magento/shopify/etc backend. It will mock out the submission of a cart form and operate like a functional backend.
```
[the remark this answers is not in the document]
lderrickable: Also, probably change the example to an add to cart action. I think this is from the contact docs.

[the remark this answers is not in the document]
griest024: This is actually erroneously copied from the in-memory docs

libs/cart/guides/drivers/magento.md
```diff
@@ -0,0 +1,28 @@
...
+
+To set up, import the `DaffCartMagentoDriverModule` from the `@daffodil/cart/testing` library and the `ApolloModule` from `angular-in-memory-web-api`.
```
lderrickable, line 5: This have in memory docs mixed in

libs/cart/guides/extension.md
```diff
@@ -0,0 +1,106 @@
...
+  }
+}
+```
+
+## Extensible GraphQL Fragments
+
+Arbitrary additional fields can be requested on the cart object. Inject a GraphQL document node containing fragments on the platform's cart type to define extra fields.
+
+The following cart drivers support extensible fragments:
+- Magento
```
lderrickable, line 37: I think we don't need to say what the supported platforms here. I think for all new drivers, we'll just have them all support extensible fragments. So as long as we say somewhere that magento is the only platform supported by cart, we should be good.

griest024: Well only the drivers that use GraphQL support this. So its not guaranteed that all future drivers will support this. Also, the In-Memory driver does not.

libs/cart/guides/extension.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Magento
+
+Provide the `DaffMagentoExtraCartFragments` to query additional fields on a Magento cart query. This applies to all of the driver calls that return a `DaffCart`, which is most of them.
```
lderrickable, line 41: This can work with all queries, right?

docs/advanced/storage.md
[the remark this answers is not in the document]
damienwebdev: This doc is about storage, which is a part of the API of Daffodil core. I would expect that docs provided at this level are more like "integration" guides, which describe tutorials on how to actually use the Persistence APIs, not simply the list of available storage APIs and their purposes.

libs/cart/guides/extension.md
damienwebdev: I think we should absolutely explicitly state which drivers support this mechanism of extension.

libs/cart/guides/extension.md
griest024: Not queries that don't return the cart, like `placeOrder`

libs/cart/guides/getting-started.md
```diff
@@ -0,0 +1,30 @@
...
+@ngModule({
+  imports:[
+    StoreModule.forRoot({}),
+    DaffCartModule
+  ]
+})
+export class AppModule {}
+```
+
+## Utilizing inside your component
```
lderrickable, line 19: Can we use the word "use" instead of "utilize"? Utilize means to use for something other than its intended purpose, so it's almost always correct to use "use" instead of "utilize"

libs/cart/guides/usage.md
```diff
@@ -0,0 +1,135 @@
...
+      type: DaffCartItemInputType.Simple,
+      productId,
+      qty
+    };
+    this.cartFacade.dispatch(new DaffCartItemAdd(input));
+  }
+}
+```
+
+> In this example, three observable streams are assigned from `cartFacade`. Then when `addSimpleItem` is called, the `cartFacade`'s  `dispatch` function is called with the appropriately formed input. The effect layer will pick up the `DaffCartItemAdd` and invoke the injected driver with the item input info and the stored cart ID. The driver layer then sends your data off to the backend and updates the three observable streams when a response is received.
```
lderrickable, line 50: `The driver layer then sends your data off to the backend and updates the three observable streams when a response is received.` This is a little incomplete. The driver layer goes back to the effects layer, which fires a success action that goes to the reducer layer to update state. Then the observable streams are automatically updated through the cart selectors, which watch state changes.

docs/advanced/storage.md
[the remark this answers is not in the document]
griest024: I agree with Damien, most of this guide is specific to core. maybe `Environment-Specific Storage Services` should be in the main docs folder but everything else belongs in core. Moved to core.

docs/advanced/fragment-introspection.md
griest024: moved

libs/cart/README.md
griest024: added file extensions, hopefully they work now?

libs/cart/README.md
griest024: removed

libs/cart/README.md
griest024: moved and added a driver usage example

libs/cart/guides/installation.md
griest024: I directed them to the getting started guide where I talk about the different ways to use Daffodil

libs/cart/guides/installation.md
griest024: moved to `drivers.md`

libs/cart/guides/getting-started.md
griest024: moved to `state.md` and more examples follow this one

libs/cart/guides/drivers/in-memory.md
griest024: fixed

libs/cart/guides/drivers/magento.md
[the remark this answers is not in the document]
griest024: fixed

libs/cart/guides/drivers/magento.md
griest024: fixed

libs/cart/guides/getting-started.md
griest024: changed

libs/cart/guides/usage.md
griest024: hm I'm actually thinking about removing a lot of the description of the internal workings here. I don't think we need to describe how it works internally in the context of this example. We can have an architecture doc which describes that stuff

libs/cart/guides/extension.md
lderrickable: Ok, maybe we should also say: "Only drivers that use GraphQL support extensible fragments because fragments are specific to GraphQL."

libs/cart/guides/extension.md
lderrickable: why can't we do this:

```typescript
import gql from 'graphql-tag';
import {
  DaffMagentoExtraOrderFragments,
  DaffCartFacade,
  DaffCartLoad,
  DaffCart
} from '@daffodil/cart';
const extraOrderFragment = gql`
  fragment ExtraCartFields on Order {
    field1 {
      amount
    }
    shipping_addresses {
      field2 {
        value
      }
    }
  }
`;
@NgModule({
  ...,
  providers: [
    {
      provide: DaffMagentoExtraOrderFragments,
      useValue: extraOrderFragment,
      multi: true
    }
  ]
})
class AppModule {}
```

or something like that

libs/cart/guides/usage.md
lderrickable: that sounds good

libs/cart/guides/extension.md
griest024: I mean, we could and should for the order package, but that doesn't belong here. `placeOrder` returns a `DaffCartOrderResult` that is neither a cart nor an order.

libs/cart/guides/getting-started.md
griest024: changes

libs/cart/guides/usage.md
griest024: changed

libs/cart/guides/extension.md
lderrickable: I was just using the example you provided. My point was just that I think we can do this with all models and all queries, which is at odds with "This applies to all of the driver calls that return a `DaffCart`, which is most of them."

libs/core/guides/advanced/storage.md
```diff
@@ -0,0 +1,59 @@
...
+  - [Environment-Specific Storage Services](#environment-specific-storage-services)
+
+Daffodil abstracts the storage layer behind the `DaffPersistenceService` interface. A number of different services implement this interface and suit different use cases.
+
+## Services
+<!-- TODO: clarify use cases -->
+
+### `DaffLocalStorageService`
+
+The `DaffLocalStorageService` utilizes `localStorage` to implement persistence. It is appropriate for browser environments.
```
lderrickable, line 18: "utilizes" => "uses"

docs/advanced/storage.md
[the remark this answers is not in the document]
lderrickable: I still disagree, because I think this is more of an "integration" guide that happens to have api docs in it rather than api docs that happen to have tutorials. I agree that the API docs should probably go in core as regular, autogenerated docs that would be handled by dgeni. I disagree that this is essentially an api doc; I think it is a tutorial guide on how to use the persistence APIs.

libs/cart/guides/state.md
```diff
@@ -0,0 +1,86 @@
+# State
```
lderrickable, line 1: I feel like all of this could go in a general daffodil doc, because all packages will need it

libs/cart/guides/state.md
```diff
@@ -0,0 +1,86 @@
...
+@ngModule({
+  imports:[
+    StoreModule.forRoot({}),
+    DaffCartModule
+  ]
+})
+export class AppModule {}
+```
+
+## Using the Facade
```
lderrickable, line 25: I think this should be "Using the Facade to Access State" and it could be contrasted with another heading of "Accessing State Directly Through Daffodil Selectors". I think that could be useful. What do you think? Maybe it would make more sense for that to go under the advanced guides?

docs/advanced/storage.md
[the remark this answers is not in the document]
lderrickable: I'm fine with resolving this though. No need for this to hold up the PR, I don't think.

libs/cart/guides/extension.md
griest024: We can do this with any fragment but currently the only one that we have implemented is the Cart one. I'm not trying to document general extension mechanisms but only the ones that currently exist in the cart package.

libs/cart/guides/extension.md
griest024: added

libs/cart/guides/extension.md
lderrickable: ok sounds good

libs/cart/guides/state.md
griest024: I think it makes to have some docs that talk about using the selectors, especially since we have a somewhat non-standard way of instantiating them. Since its not the recommended way to access state though, I think it should be an advanced guide or separate section that's presented as an alternative to using the facade.

I think this section header should stay as it is because the section also documents dispatching actions through the facade.

libs/cart/guides/state.md
griest024: I think we should have a general doc that talks about Daffodil state architecture but this guide is talking about cart-specific stuff. This guide will be somewhat redundant but I don't think thats bad.

libs/core/guides/advanced/storage.md
griest024: fixed

libs/cart/guides/state.md
griest024: added selector doc

libs/cart/guides/state.md
```diff
@@ -82,5 +83,56 @@ export class CartItemComponent implements OnInit {
 
 > In this example, three observable streams are assigned from `cartFacade`. Then when `addSimpleItem` is called, the `cartFacade`'s  `dispatch` function is called with the appropriately formed input. The input data is then sent off to the backend and the three observable streams are updated when a response is received.
 
+## Accessing State with Selectors
+
+Accessing state with the facade is recommended but for greater flexibility the redux store's state can be directly accessed with Daffodil's selectors.
+
+To properly maintain memoization of selectors, Daffodil selectors are hidden behind an exported getter function. This function returns an object that contains the selectors.
```
lderrickable, line 90: "To properly maintain memoization of selectors that are compatible with generic models" to give the full reason why.

griest024: changed

lderrickable:
lgtm

lderrickable (approved):

docs/advanced/storage.md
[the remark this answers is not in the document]
griest024: @damienwebdev bump. This guide has been moved to core

damienwebdev (approved):

PR #1098, opened 2020-10-15
feat(tools-dgeni): remove internal properties from generated docs

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fields that are not part of the public API are shown in the generated docs.

## What is the new behavior?
Fields that are prefixed with an underscore and the angular lifecycle hooks are removed during docs generation.

```
[ ] Yes
[x] No
```


damienwebdev:
I like this idea, but maybe instead of trying to maintain a discrete set/matching underscores, we have a flag like `@private` somewhere in the method docs. I've noticed something similar when reading Angular internal docs and I think the premise is sound.

lderrickable:
> I like this idea, but maybe instead of trying to maintain a discrete set/matching underscores, we have a flag like `@private` somewhere in the method docs. I've noticed something similar when reading Angular internal docs and I think the premise is sound.

Supposedly (according to the dgeni docs), there is a tag, `@private`, that can mark a field as private and filter it out with provided processors, but I couldn't get it to work. So I made it filter on `@docs-private`; i.e. any field that is labeled as `@docs-private` will not be added to the docs.

lderrickable:
When looking into another issue, I learned that this is exactly how angular tags properties as private. They tag it with the custom tag "@docs-private" and add this tag to the list of supported tag definitions. I added this tag to our list of supported tag definitions.

damienwebdev:
Can we then check the tag instead as opposed to searching the entire content for the string?

lderrickable:
> Can we then check the tag instead as opposed to searching the entire content for the string?

yep, done

damienwebdev:
Lgtm.

damienwebdev (approved):

PR #1099, opened 2020-10-15
docs(product): add some api docs for models and facades

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
Added some API docs for models and facades.

### DIFF
--- a/libs/product/src/facades/composite-product/composite-product-facade.interface.ts
+++ b/libs/product/src/facades/composite-product/composite-product-facade.interface.ts
@@ -7,6 +7,9 @@ import { DaffStoreFacade } from '@daffodil/core/state';
 import { DaffCompositeProductItemOption, DaffCompositeProductItem } from '../../models/composite-product-item';
 import { DaffCompositeProduct } from '../../models/composite-product';
 
+/**
+ * A facade for accessing composite product state from an application component.
+ */
 export interface DaffCompositeProductFacadeInterface extends DaffStoreFacade<Action> {
 
 	/**
--- a/libs/product/src/facades/composite-product/composite-product-facade.interface.ts
+++ b/libs/product/src/facades/composite-product/composite-product-facade.interface.ts
@@ -8,7 +8,8 @@ import { DaffCompositeProductItemOption, DaffCompositeProductItem } from '../../
 import { DaffCompositeProduct } from '../../models/composite-product';
 
 /**
- * A facade for accessing composite product state from an application component.
+ * A facade for interacting with the composite product state.
+ * Exposes many parts of the state for easy access and allows dispatching of actions.
  */
 export interface DaffCompositeProductFacadeInterface extends DaffStoreFacade<Action> {
 
--- a/libs/product/src/facades/composite-product/composite-product.facade.ts
+++ b/libs/product/src/facades/composite-product/composite-product.facade.ts
@@ -22,67 +22,156 @@ export class DaffCompositeProductFacade<T extends DaffProduct = DaffProduct> imp
 	selectors = getDaffProductSelectors<T>();
 	
 	constructor(private store: Store<DaffProductReducersState<T>>) {}
-	
+
+	/**
+	 * Get the minimum price for a composite product, excluding optional items and regardless of applied options.
+	 * This would be useful for a quick preview of a product.
+	 * @param id the id of the composite product.
+	 */
 	getMinPossiblePrice(id: string): Observable<number> {
 		return this.store.pipe(select(this.selectors.selectCompositeProductMinPossiblePrice, { id }));
 	}
 
+	/**
+   * Get the maximum price for a composite product, including optional items and regardless of applied options.
+	 * This would be useful for a quick preview of a product.
+	 * @param id the id of the composite product.
+	 */
 	getMaxPossiblePrice(id: string): Observable<number> {
 		return this.store.pipe(select(this.selectors.selectCompositeProductMaxPossiblePrice, { id }));
 	}
 
+	/**
+	 * Returns whether the composite product could have a price range in any configuration,
+	 * including the optional items and regardless of the current item option selection.
+	 * This could be useful for a quick preview of the product.
+	 * @param id the id of the composite product.
+	 */
 	possiblyHasPriceRange(id: string): Observable<boolean> {
 		return this.store.pipe(select(this.selectors.selectCompositeProductPossiblyHasPriceRange, { id }));
 	}
 
+	/**
+	 * Get the minimum discounted price for a composite product, excluding optional items and regardless of the current selection of item options.
+	 * This could be useful for a quick preview of the product.
+	 * @param id the id of the composite product.
+	 */
 	getMinPossibleDiscountedPrice(id: string): Observable<number> {
 		return this.store.pipe(select(this.selectors.selectCompositeProductMinPossibleDiscountedPrice, { id }));
 	}
 
+	/**
+	 * Get the maximum discounted price for a composite product, including optional items and regardless of the current selection of item options.
+	 * This could be useful for a quick preview of the product.
+	 * @param id the id of the composite product.
+	 */
 	getMaxPossibleDiscountedPrice(id: string): Observable<number> {
 		return this.store.pipe(select(this.selectors.selectCompositeProductMaxPossibleDiscountedPrice, { id }));
 	}
 
+	/**
+	 * Selector for whether the composite product could have a discounted price range in any configuration,
+	 * including the optional items and regardless of the current item option selection.
+	 * This could be useful for a quick preview of the product.
+	 * @param id the id of the composite product.
+	 */
 	possiblyHasDiscountedPriceRange(id: string): Observable<boolean> {
 		return this.store.pipe(select(this.selectors.selectCompositeProductPossiblyHasDiscountedPriceRange, { id }));
 	}
 
+	/**
+	 * Returns whether the minimum and maximum composite product discounted prices equal the minimum and maximum
+	 * pre discount prices, including optional items. Note: this intentionally misses a usecase where there is a discount
+	 * in between the minimum and maximum prices. This is because the resulting UI would be non-sensical to the customer for
+	 * this usecase. For example, if we have a set of item options' prices: (price, discount), (10, 0), (20, 0), (15, 4).
+	 * While it is true that there is a discounted price (15, 4), the user would see the original price range (10-20) followed
+	 * by the discounted price range (10-20). This would happen, because the 15-4=11 price would fall between the two extremes
+	 * of the discounted prices and it wouldn't make sense to the customer what is happening.
+	 * @param id the id of the composite product.
+	 */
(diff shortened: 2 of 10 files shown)


libs/product/src/facades/composite-product/composite-product-facade.interface.ts
```diff
@@ -7,6 +7,9 @@ import { DaffStoreFacade } from '@daffodil/core/state';
 import { DaffCompositeProductItemOption, DaffCompositeProductItem } from '../../models/composite-product-item';
 import { DaffCompositeProduct } from '../../models/composite-product';
 
+/**
+ * A facade for accessing composite product state from an application component.
```
griest024, line 11: The facade can be used to dispatch actions as well and its not really restricted just to components. I would say something like:
```ts
/**
 * A facade for interacting with the composite product state.
 * Exposes many parts of the state for easy access and allows dispatching of actions.
 */
```

libs/product/src/facades/composite-product/composite-product.facade.ts
```diff
@@ -22,67 +22,156 @@ export class DaffCompositeProductFacade<T extends DaffProduct = DaffProduct> imp
 	selectors = getDaffProductSelectors<T>();
 	
 	constructor(private store: Store<DaffProductReducersState<T>>) {}
-	
+
+	/**
+	 * Get the minimum price for a composite product, excluding optional items and regardless of applied options.
```
griest024, line 27: Aren't these documented on the interface? I say its better to configure dgeni to inherit the docs from the interface or just direct people to the interface.

libs/product/src/facades/product/product-facade.interface.ts
```diff
@@ -5,17 +5,46 @@ import { DaffStoreFacade } from '@daffodil/core/state';
...
 	hasDiscount(id: string): Observable<boolean>;
+
+	/**
+	 * Get the discount amount of a product.
+	 * @param id a product id
+	 */
 	getDiscountAmount(id: string): Observable<number>;
+
+	/**
+	 * Get the discount percent of a product
```
griest024, line 40: Missing period.

libs/product/src/facades/product/product.facade.ts
```diff
@@ -33,23 +32,43 @@ export class DaffProductFacade<T extends DaffProduct = DaffProduct> implements D
...
+	/**
+	 * Get the discount amount of a product.
+	 * @param id a product id
+	 */
 	getDiscountAmount(id: string): Observable<number> {
 		return this.store.pipe(select(this.selectors.selectProductDiscountAmount, { id }));
 	}
 
+	/**
+	 * Get the discount percent of a product
```
griest024, line 61: period

libs/product/src/models/composite-product-item.ts
```diff
@@ -1,10 +1,19 @@
...
 	select = 'select',
 	radio = 'radio'
 }
 
+/**
+ * The composite product item describes one set of product options that the user can add to the composite product.
+ * If a composite product is a toolbox bundle, a composite product item might be a screw driver, and the options contained in that
+ * item might be a phillips head and a flathead. The customer would choose to add either a phillips head or a flathead to the
+ * composite product, or neither if the item is optional.
+ */
```
griest024, line 16: I think this is a slightly better:

```ts
/**
 * The composite product item describes one set of product options that the user can add to the composite product.
 * A composite product item can be required or optional. If it is required, an option _must_ be chosed in order to add the product to the cart.
 * If the item is optional, the product can be added to the cart without an option having been chosen.
 * For example, if the composite product in question is a toolbox bundle, a composite product item could be a screw driver.
 * The options contained in that item could therefore be a phillips head and a flathead.
 * The customer could choose to add either a phillips head or a flathead to the composite product.
 */
```

Double check to make sure my descriptions are accurate.

libs/product/src/models/composite-product.ts
```diff
@@ -1,6 +1,14 @@
 import { DaffProduct } from './product';
 import { DaffCompositeProductItem } from './composite-product-item';
 
+/**
+ * A composite product is a group of products sold together. It includes one primary product and many accessory "items". 
+ * The composite product items are additional products the customer might want to purchase with the primary product, 
+ * and each item has a number of options from which the user can choose. Composite product items can be optional or required.
+ * For example, a toolbox bundle might have a primary product of the toolbox. The items could be a hammer and a screw driver, 
+ * and the options for these items could be two different hammers and two different screw drivers the customer could choose between. 
+ * All of these items could have different prices/discounts/etc, which means the composite products can have ranged prices.
```
griest024, line 10: IMO
```
 * and the options for these items could be two different hammers and two different screwdrivers between which the customer could choose. 
 * All of these items could have different prices/discounts/etc which would cause the composite products to have ranged prices.
```

libs/product/src/models/configurable-product.ts
```diff
@@ -3,17 +3,33 @@ import { DaffSortable } from '@daffodil/core';
...
 }
 
+/**
+ * A variant is one version of the configurable product with all attributes chosen. Variants exist because you might not have
+ * all versions of every configuration of the product in stock. For example, you might have a medium, red shirt and a small,
+ * green shirt, but no small, red shirts. So you could have two variants (mediumRed, smallGreen) rather than the maximum 4 variants 
+ * (smallRed, mediumRed, smallGreen, mediumGreen). This ensures the customer can't add a configurable product to the cart that is not
+ * in stock. However, variants don't usually need to be considered by the frontend dev, because daffodil abstacts the concept of variants away into 
+ * an "available attributes" selector.
+ */
```
griest024, line 32: I think we want to avoid speaking in the 2nd person. So instead of `Variants exist because you might not have all versions of every configuration of the product in stock` you would say `Variants exist because its possible that not all versions of every configuration of the product are in stock`.

FWIW I need to go back and fix this in my cart docs.

griest024 (changes requested):

libs/product/src/models/configurable-product.ts
griest024: Actually @damienwebdev what is your opinion on this? Is `you` and `your` okay or not?

lderrickable: I think you're right. Probably better to go more formal with these

libs/product/src/facades/composite-product/composite-product.facade.ts
lderrickable: Dgeni is pretty hellish to try to bend to your will. I think we could make the docs have a link to the interface in the overall documentation for this file, but configuring dgeni to inherit the docs from the interface would be really tricky

griest024: Yeah I think its preferable to have [a link](https://github.com/angular/dgeni-packages#inline-tag-definitions) back to the interface field so we only have to maintain those docs in one place.

libs/product/src/facades/composite-product/composite-product-facade.interface.ts
lderrickable: changed

libs/product/src/facades/composite-product/composite-product.facade.ts
lderrickable: done. Though I did leave the doc for the `dispatch` function, since this is hidden behind an `extends` in the facade interface. I think this could pose a problem with this documentation pattern if there are multiple levels of interface extension.

libs/product/src/facades/product/product-facade.interface.ts
lderrickable: fixed

libs/product/src/facades/product/product.facade.ts
lderrickable: this comment was removed in favor of the interface

libs/product/src/models/composite-product-item.ts
lderrickable: I took the first part added about optional vs required, but left the rest as it was

libs/product/src/models/composite-product.ts
lderrickable: changed

libs/product/src/models/configurable-product.ts
lderrickable: changed

libs/product/src/facades/composite-product/composite-product.facade.ts
griest024: I think you can use `{@link some/uri Some Title}` to link to specific facade interface fields, not sure though. I thought we could link back to the interface fields for each of the facade fields.

libs/product/src/models/composite-product-item.ts
```diff
@@ -10,8 +10,10 @@ export enum DaffCompositeProductItemInputEnum {
 
 /**
  * The composite product item describes one set of product options that the user can add to the composite product.
- * If a composite product is a toolbox bundle, a composite product item might be a screw driver, and the options contained in that
- * item might be a phillips head and a flathead. The customer would choose to add either a phillips head or a flathead to the
+ * A composite product item can be required or optional. If it is required, an option _must_ be chosen in order to add the product to the cart.
+ * If the item is optional, the product can be added to the cart without an option having been chosen.
+ * For example, f a composite product is a toolbox bundle, a composite product item might be a screw driver, and the options contained in that
```
griest024, line 15: `f a composite product` -> `if a composite product`

griest024 (changes requested):

libs/product/src/facades/composite-product/composite-product.facade.ts
lderrickable: Wouldn't we need to create anchors to the places the links go to? And those anchors would need ids. Then we'd need to upkeep the links to go to the correct ids, so we'd end up with the same problem of extra maintenance. Though I guess we wouldn't need to maintain the actual descriptions.

griest024: According to the docs that I linked ^^^, the dgeni jsdoc package can automatically replace link tags with anchors. I think the ID would just be the name of the field so I don't think that would really need any maintenance unless we refactor the name of the field. And in that case a search and replace should pick up the field name in the comments.

lderrickable: I looked into this and it looks doable, since there are examples in the angular material repo. I'll need to configure some things in dgeni, so I'd like to do that in another PR. Added an issue for it here: https://github.com/graycoreio/daffodil/issues/1102

libs/product/src/models/composite-product-item.ts
lderrickable: fixed

libs/product/src/facades/composite-product/composite-product.facade.ts
griest024: sounds good

griest024 (approved):

PR #1100, opened 2020-10-15
docs(design): mark internal properties as private

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Many internal properties are included in the generated api docs, because they are not marked as private.

## What is the new behavior?
The internal properties are all marked as `private`, so they will not end up in the generated api docs.

```
[ ] Yes
[x] No
```

I also moved some properties around so that they would show up in a desired order in the API docs. For example, I made `Input`s come before `Output`s, which come before all non-decorated properties.

### DIFF
--- a/libs/design/src/atoms/button/button.component.ts
+++ b/libs/design/src/atoms/button/button.component.ts
@@ -73,8 +73,9 @@ export class DaffButtonComponent
   extends _daffButtonBase
   implements OnInit, DaffPrefixable, DaffSuffixable, DaffColorable, DaffSizeable<DaffButtonSize> {
     @Input() color: DaffPalette;
-    buttonType: DaffButtonType;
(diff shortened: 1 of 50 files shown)


griest024:
So properties that actually use the TS keyword `private` don't need this doc tag? They are automatically excluded from API docs?

lderrickable:
> So properties that actually use the TS keyword `private` don't need this doc tag? They are automatically excluded from API docs?

yes, though I had to remove a lot of those because it was failing the build. Most of them just use the `@docs-private` tag now

griest024:
Seems like the build is still failing.

damienwebdev (approved):
lgtm

PR #1111, opened 2020-10-24
feat(all): swap error names to AOT compile-safe comparison

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, we swap on function names on errors, turns out this doesn't actually work in AOT builds, as names are mutilated by the compiler. 

Fixes: N/A

## What is the new behavior?
This introduces a static "code" property to Daffodil errors that we can switch against reliably. 

```
[x] Yes
[ ] No
```

This change is breaking, but more of a "fix" as the original code never really functioned correctly.

## Other information

### DIFF
--- a/libs/cart/src/effects/cart-address.effects.ts
+++ b/libs/cart/src/effects/cart-address.effects.ts
@@ -32,7 +32,7 @@ export class DaffCartAddressEffects<T extends DaffCartAddress, V extends DaffCar
       map(() => this.storage.getCartId()),
       switchMap(cartId => this.driver.update(cartId, action.payload)),
       map((resp: V) => new DaffCartAddressUpdateSuccess(resp)),
-      catchError(error => of(error.name === DaffStorageServiceError.name
+      catchError(error => of(error.code === DaffStorageServiceError.code
         ? new DaffCartStorageFailure('Cart Storage Failed')
         : new DaffCartAddressUpdateFailure('Failed to update cart address')
       )),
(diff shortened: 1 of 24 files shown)


libs/cart/src/effects/cart-address.effects.ts
```diff
@@ -32,7 +32,7 @@ export class DaffCartAddressEffects<T extends DaffCartAddress, V extends DaffCar
       map(() => this.storage.getCartId()),
       switchMap(cartId => this.driver.update(cartId, action.payload)),
       map((resp: V) => new DaffCartAddressUpdateSuccess(resp)),
-      catchError(error => of(error.name === DaffStorageServiceError.name
+      catchError(error => of(error.code === DaffStorageServiceError.code
```
griest024, line 35: `code` is a class field so you would need to do something like `error.constructor.code`.

griest024 (changes requested):

libs/cart/src/effects/cart-address.effects.ts
damienwebdev: I'm not 100% certain, but I believe typescript compilation may do this automatically.

I'll test tomorrow and verify.

damienwebdev: I've swapped this for instanceof checking.

libs/core/src/errors/inheritable-error.ts
```diff
@@ -0,0 +1,27 @@
...
+ * 
+ * let myError = new MyError();
+ * 
+ * myError instanceof MyError; // returns false
+ * ```
+ * 
+ * You will see unexpected things. This class fixes that issue as described here
+ * https://github.com/microsoft/TypeScript/issues/13965
+ */
+export class DaffInheritableError extends Error {
```
griest024, line 18: Can you add an integration test demonstrating that inheritance checking works?

griest024 (changes requested):

PR #1114, opened 2020-10-27
feat(order): add magento 2.4.1 driver queries

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
+++ b/libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
@@ -35,21 +35,6 @@ export const orderItemFragment = gql`
 
 export const orderBundleItemFragment = gql`
   fragment orderBundleItem on BundleOrderItem {
-    quantity_ordered
-    quantity_canceled
-    quantity_shipped
-    quantity_invoiced
-    product_url_key
-    product_sku
-    product_name
-    product_sale_price {
-      value
-    }
-    discounts {
-      amount {
-        value
-      }
-    }
     bundle_options {
       ...orderBundleItemSelectedOption
     }
--- a/libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
+++ b/libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
@@ -1,0 +1,58 @@
+import gql from 'graphql-tag';
+
+export const orderBundleItemSelectedOptionFragment = gql`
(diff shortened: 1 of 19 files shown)


lderrickable:
Are there API docs for Magento orders in 2.4.1? I found the release notes, but there isn't too much about orders.

griest024:
@lderrickable https://devdocs.magento.com/guides/v2.4/graphql/queries/customer.html#customerOrders

libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
```diff
@@ -0,0 +1,58 @@
...
+    }
+    selected_options {
+      label
+      value
+    }
+  }
+`;
+
+export const orderBundleItemFragment = gql`
+  fragment orderBundleItem on BundleOrderItem {
```
lderrickable, line 37: I think this could simply be:
```
fragment orderBundleItem on BundleOrderItem {
    bundle_options {
      ...orderBundleItemSelectedOption
    }
  }
  ${orderBundleItemSelectedOptionFragment}
}
```
Then the usage would be like:
```
export const orderShipmentBundleItemFragment = gql`
  fragment orderShipmentBundleItem on BundleShipmentItem {
    quantity_shipped
    order_item {
      ...orderBundleItem,
      ...orderItem
    }
  }
  ${orderBundleItemFragment}
  ${orderItemFragment}
`;
```

griest024: yeah that's better. changed

lderrickable:
While I question Magento's architecture decisions in modeling a lot of the order feature, this PR lgtm.

For posterity, I'll say that our query fragments could be greatly simplified if the CreditMemoItem, InvoiceItem, and ShipmentItem didn't have a `bundle_options` attribute (and therefore would not need a "Bundle" extended version of them; e.g. BundleCreditMemoItem), because that is included in the `cart_item` attribute already.

Also, it would be nice if InvoiceTotal and OrderTotal extended a common model, but this is not included in the API docs.

lderrickable (approved):

griest024:
@lderrickable yeah the actual fields on a lot of the models are exactly the same, its only the name of the type that is different. Very unfortunate. Also the order model does not take into account multiple shipping or billing addresses.

PR #1119, opened 2020-10-28
fix(order): bad bundle item fragments and query field

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

- The guest orders query tries to query the `orders` field; the correct field is `items`.
- Fragments on different types of item cannot be combined.
- Some order "placed order" state operations will fail because `id` was set for the `DaffOrder#id` and that cannot be used for matching to cart place order result order ID.
- The driver does not query for product type.

## What is the new behavior?
- The guest orders query tries to query `items`.
- Fragments on different types of item define all the item fields.
- "placed order" state operations works as expected when operating on a ID that is set to the Magento order number.
- The driver queries for product type.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
+++ b/libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
@@ -43,4 +43,5 @@ export const orderItemFragment = gql`
       }
     }
   }
+  ${orderBundleItemSelectedOptionFragment}
 `;
--- a/libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
+++ b/libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
@@ -2,15 +2,21 @@ import gql from 'graphql-tag';
 
 export const orderBundleItemSelectedOptionFragment = gql`
   fragment orderBundleItemSelectedOption on ItemSelectedBundleOption {
+    __typename
+    id
     label
     values {
+      __typename
+      id
       product_name
     }
   }
 `;
 
 export const orderItemFragment = gql`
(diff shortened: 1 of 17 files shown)


griest024:
@lderrickable FYI combining fragments doesn't actually work like we though because all the concrete schema types are totally distinct.

libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
```diff
@@ -18,6 +18,7 @@ export const orderItemFragment = gql`
     product_url_key
     product_sku
     product_name
+    product_type
```
damienwebdev, line 21: Is this not a __typename?

damienwebdev (changes requested):

libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
griest024: I guess it could be implemented with that but no, `product_type` is an actual field returned on the order item model.

lderrickable (dismissed):

lderrickable:
lgtm

libs/order/driver/magento/2.4.1/src/queries/fragments/order-item.ts
griest024: bump. we need to use `product_type` to differentiate between simple and configurables.

damienwebdev (approved):
lgtm

lderrickable (approved):

PR #1120, opened 2020-10-28
feat(product): change the api for composite product facade and selectors

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
The composite product selectors are now replaced with 
`selectCompositeProductPricesForConfiguration`, 
`selectCompositeProductPrices`,
`selectCompositeProductPricesAsCurrentlyConfigured`

and two functions:
`productPriceRangeHasDiscount`,
`productPriceRangeHasPriceRange`

```
[x] Yes
[ ] No
```

## Other information
- I decided against using a union type (of `DaffPriceRange` and `DaffPrice`) for the return type of the selectors (Damien and I had originally planned on this). I'm actually sure if this is even accurately called a "union" type, because the two types don't share anything in common. It turned out that in order to decide whether to return a `DaffPriceRange` or a `DaffPrice`, I would basically need to pass the final value (a `DaffPriceRange`) into the `productPriceRangeHasPriceRange` helper function, then rebuild the object into a `DaffPrice` if it is not a range of prices. Then the app dev would need to call some function to find out if they're being passed a `DaffPriceRange` or a `DaffPrice` anyway. I figure it's easier for us to just pass the app dev a `DaffPriceRange`, have them call the `productPriceRangeHasPriceRange` function through the facade, then just have them access the minimum values as if it was the only price given (since they now know the min and max prices are equal).
- I made the discounts null for both the amount and percent, because those values really don't make any sense with the composite product. Because each option can have different prices and discounts, the discount amount ceases to mean anything once it is disassociated from a particular item option. For example, one item option might be really cheap ($4), and has a small discount ($0.50). If another option is more expensive ($10) with a large discount ($9), the largest discount amount won't make any sense to the end user. They would see something like 
```
Original Price: $4 - $10; 
DiscountedPrice: $1-$3.5; 
discount of $0.5-9$
``` 
It isn't very obvious how the discount amount is derived from the original and discounted prices.

### DIFF
--- a/libs/product/src/facades/composite-product/composite-product-facade.interface.ts
+++ b/libs/product/src/facades/composite-product/composite-product-facade.interface.ts
@@ -23,7 +23,7 @@ export interface DaffCompositeProductFacadeInterface extends DaffStoreFacade<Act
 	getPricesForConfiguration(id: string, configuration: Dictionary<DaffCompositeConfigurationItem>): Observable<DaffPriceRange>;
 
 	/**
-	 * Get the broadest possible DaffPriceRange for a composite product excluding optional item prices.
+	 * Get the broadest possible DaffPriceRange for a composite product including optional item prices.
 	 * @param id the id of the composite product.
 	 */
 	getPrices(id: string): Observable<DaffPriceRange>;
--- a/libs/product/src/facades/composite-product/composite-product-facade.interface.ts
+++ b/libs/product/src/facades/composite-product/composite-product-facade.interface.ts
@@ -6,6 +6,8 @@ import { DaffStoreFacade } from '@daffodil/core/state';
 
 import { DaffCompositeProductItemOption, DaffCompositeProductItem } from '../../models/composite-product-item';
 import { DaffCompositeProduct } from '../../models/composite-product';
+import { DaffPriceRange } from '../../models/prices';
+import { DaffCompositeConfigurationItem } from '../../models/composite-configuration-item';
 
 /**
  * A facade for interacting with the composite product state.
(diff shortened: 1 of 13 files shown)


griest024:
> `selectCompositeProductPricesIncludingOptionalItems` (a new selector, which would pretty complicated)

Is it complicated because it's difficult to compute the configuration that includes optional items?

lderrickable:
> > `selectCompositeProductPricesIncludingOptionalItems` (a new selector, which would pretty complicated)
> 
> Is it complicated because it's difficult to compute the configuration that includes optional items?

It would be complicated, because I would need to create an additional function that looks almost exactly like the `getMaxPricesForConfiguration` function but that included option item prices. Then I would call those functions directly instead of the `selectCompositeProductPricesForConfiguration`.

lderrickable:
Now that I've written that out and thought through it a little more, I think maybe I should just add the additional function. What do you think?

griest024:
> Now that I've written that out and thought through it a little more, I think maybe I should just add the additional function. What do you think?

It depends on how common you think that use case is. If its likely that an app dev will want a price including optional items, then we should definitely encapsulate that complexity in Daffodil. I'm generally in favor of including more functionality in the lib.

> I could probably just have the selectCompositeProductPrices selector that always includes optional items

Is that the API that is desired? I'm not well versed in the use cases but if you think its best then that sounds good.

lderrickable:
> Is that the API that is desired? I'm not well versed in the use cases but if you think its best then that sounds good.

Ok I think it is correct to always include the optional items for that case. I made the changes

libs/product/src/facades/helpers.ts
```diff
@@ -0,0 +1,19 @@
+import { DaffPriceRange } from '../models/prices';
+
+/**
+ * Returns whether a DaffPriceRange has a discount.
+ * @param priceRange a DaffPriceRange
+ */
+export const productPriceRangeHasDiscount = (priceRange: DaffPriceRange): boolean => {
```
griest024, line 7: These aren't showing up in the dgeni docs, do you know how to get dgeni to pick them up? It might not be in the scope of this PR.

libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -357,3 +128,92 @@ function getMaximumRequiredCompositeItemDiscountedPrice(item: DaffCompositeProdu
...
+		discountedPrice: product.items.reduce((acc, item) => daffAdd(
+			acc, 
+			appliedOptions && appliedOptions[item.id] ? 
+				daffMultiply(
+					appliedOptions[item.id].discount ? daffSubtract(appliedOptions[item.id].price, appliedOptions[item.id].discount.amount) : appliedOptions[item.id].price, 
+					appliedOptions[item.id].quantity
+				) : 
+				getMinimumRequiredCompositeItemDiscountedPrice(item)
+		), product.discount ? daffSubtract(product.price, product.discount.amount) : product.price),
+		discount: { amount: null, percent: null },
```
griest024, line 148: Do these discounts need to have values?

libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -357,3 +128,92 @@ function getMaximumRequiredCompositeItemDiscountedPrice(item: DaffCompositeProdu
...
+/**
+ * Gets the maximum prices of a composite product including optional item prices.
+ * @param product a DaffCompositeProduct
+ */
+function getMaxPricesIncludingOptionalItems(product: DaffCompositeProduct): DaffProductPrices {
+	return {
+		discountedPrice: (<DaffCompositeProduct>product).items.reduce((acc, item) => daffAdd(
+			acc, 
+			Math.max(...item.options.map(option => option.discount ? daffSubtract(option.price, option.discount.amount) : option.price))
+		), product.discount ? daffSubtract(product.price, product.discount.amount) : product.price),
```
griest024, line 197: Can you extract `product.discount ? daffSubtract(product.price, product.discount.amount) : option.price` into its own function? This logic is used multiple times in this file for both options and products.

griest024: Note that you should also use this extracted function in `getMaximumRequiredCompositeItemDiscountedPrice` and `getMinimumRequiredCompositeItemDiscountedPrice`. So for example: 

```ts
function getDiscountablePrice(product: DaffProduct): number {
  return product.discount ? daffSubtract(product.price, product.discount.amount) : product.price
}

function getMaximumRequiredCompositeItemDiscountedPrice(item: DaffCompositeProductItem): number {
  return item.required ? Math.max(...item.options.map(getDiscountablePrice)) : 0;
}
```

griest024 (changes requested):

libs/product/src/facades/helpers.ts
lderrickable: No, I'm not sure. We also need to figure out how to get the selector docs to work. Definitely out of scope though

libs/product/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: done. And thanks for the extra note. Definitely would have missed that

libs/product/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: No, I explain why in the PR description

libs/product/src/facades/composite-product/composite-product-facade.interface.ts
```diff
@@ -14,112 +16,23 @@ import { DaffCompositeProduct } from '../../models/composite-product';
...
-	 * in between the minimum and maximum prices. This is because the resulting UI would be non-sensical to the customer for
-	 * this usecase. For example, if we have a set of item options' prices: (price, discount), (10, 0), (20, 0), (15, 4).
-	 * While it is true that there is technically a discounted price, the user would see the original price range (10-20) followed
-	 * by the discounted price range (10-20). This would happen, because the 15-4=11 price would fall between the two extremes
-	 * of the discounted prices and it wouldn't make sense to the customer what is happening.
+	 * Get the DaffPriceRange for a composite product based on the current configuration of selected item options in redux state.
 	 * @param id the id of the composite product.
 	 */
-	hasRequiredItemDiscount(id: string): Observable<boolean>;
+	getPricesAsCurrentlyConfigured(id: string): Observable<DaffPriceRange>;
```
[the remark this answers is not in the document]
lderrickable: Added to the docs

libs/product/src/selectors/composite-product/composite-product.selectors.ts
griest024: oh right mb.

libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -357,3 +129,96 @@ function getMaximumRequiredCompositeItemDiscountedPrice(item: DaffCompositeProdu
...
+ * @param product a DaffCompositeProduct
+ * @param appliedOptions a Dictionary<DaffCompositeProductItemOption> that determines the current configuration of the composite product.
+ */
+function getMinPricesForConfiguration(product: DaffCompositeProduct, appliedOptions: Dictionary<DaffCompositeProductItemOption>): DaffProductPrices {
+	return {
+		discountedPrice: product.items.reduce((acc, item) => daffAdd(
+			acc, 
+			appliedOptions && appliedOptions[item.id] ? 
+				daffMultiply(
+					appliedOptions[item.id].discount ? daffSubtract(appliedOptions[item.id].price, appliedOptions[item.id].discount.amount) : appliedOptions[item.id].price,
```
griest024, line 144: Use `getDiscountedPrice` here.

libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -357,3 +129,96 @@ function getMaximumRequiredCompositeItemDiscountedPrice(item: DaffCompositeProdu
...
+ * @param product a DaffCompositeProduct
+ * @param appliedOptions a Dictionary<DaffCompositeProductItemOption> that determines the current configuration of the composite product.
+ */
+function getMaxPricesForConfiguration(product: DaffCompositeProduct, appliedOptions: Dictionary<DaffCompositeProductItemOption>): DaffProductPrices {
+	return {
+		discountedPrice: product.items.reduce((acc, item) => daffAdd(
+			acc, 
+			appliedOptions && appliedOptions[item.id] ? 
+				daffMultiply(
+					appliedOptions[item.id].discount ? daffSubtract(appliedOptions[item.id].price, appliedOptions[item.id].discount.amount) : appliedOptions[item.id].price,
```
griest024, line 172: Use `getDiscountedPrice` here.

griest024 (changes requested):

libs/product/src/selectors/composite-product/composite-product.selectors.ts
griest024: You still need to change `getMaximumRequiredCompositeItemDiscountedPrice` and `getMinimumRequiredCompositeItemDiscountedPrice`.

lderrickable: Oh nice. Done

griest024: no line 119 and 129.

lderrickable: sorry about that. I think that's the last ones

griest024 (approved):

PR #1126, opened 2020-11-05
feat(cart): add state to each cart item to indicate if it is new, mutating, etc

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
If a second `DaffCartItemAdd` (or `DaffCartitemUpdate` etc) action is dispatched before the debounce timer is complete, the debounce timer will reset to the `DaffCartItemStateDebounceTimer`. When the debounce timer does hit zero, all cart item states will be reset to "Default". This is to avoid edge cases and complex logic around updating and adding different items in quick succession. I think if we try to manage individually concurrent cart item request states (instead of grouping all concurrent requests together into a single debounce timer), we'll end up with bugs and much more complex code. The only downside to grouping all the concurrent requests into a single debounce timer is that a cart item state can last longer than 4 seconds, since the debounce timer resets whenever another cart item request is made. I don't think this is a big deal though, since people don't usually edit cart items like mad and I don't think it will make much difference in terms of user experience.


libs/cart/src/effects/cart-item.effects.ts
```diff
@@ -65,16 +72,33 @@ export class DaffCartItemEffects<
...
+				}),
         catchError(error => of(new DaffCartItemAddFailure('Failed to add cart item')))
       )
     )
+	)
+
+	@Effect()
+  resetCartItemStateAfterAdd$ = this.actions$.pipe(
+		ofType(DaffCartItemActionTypes.CartItemAddSuccessAction),
+		debounceTime(this.cartItemStateDebounceTime),
```
griest024, line 96: Is it sufficient to have a degrading new state for cart items? Don't we need a static reference to the most recently added cart item?

libs/cart/src/effects/cart-item.effects.ts
```diff
@@ -65,16 +72,33 @@ export class DaffCartItemEffects<
...
+				this.store.pipe(select(
+					getDaffCartItemEntitiesSelectors<V, X, T>().selectAllCartItems)
+				),
+				this.driver.add(this.storage.getCartId(), action.input)
+			]
+			).pipe(
+        map(([cartItems, resp]: [T[], V]) => {
+					//find the cart item that is new or that has a different qty and change the state to "new"
+					//I think we also need to attach the cart item id to the action, so the "new" state can be 
+					//changed back to "default" in a "reset" effect below.
```
griest024, line 85: Seems good, how would you find the cart ID that has changed?

lderrickable: I would check if the length of the cartItems array has changed. If it has, find the cart item id that doesn't belong (it would be nice if we could just take the last cart item, but I'm not sure the new cart item will always be the last one). If the length of the array hasn't changed, compare the quantities of cart items and take the cart item with a different quantity.

libs/cart/src/effects/cart-item.effects.ts
lderrickable: I think it's pretty clean to just have a single source of the cart item state. That way the app dev can have really simple access to any given state of the cart item. And I think if we made the state non-degrading, the app dev would need to find some way to reset the state, because I think the use case is usually for a notification of some sort that disappears after some time. Plus, since we now attach the cart item id to the "DaffCartItemAddSuccess" action, the app dev could always add their own static reference to the most recently added cart item without too much trouble. Is there a use case that I'm not thinking of where we would need to have the static reference? I think that for now, we try to keep the whole thing as simple as possible for the app dev until such a use case comes up and we could add the static reference then as a separate part of state. For the use cases that exist now though, I think the degrading state makes sense and will make life for an app dev really simple.

libs/cart/src/effects/cart-item.effects.ts
damienwebdev: I think:

> new cart item will always be the last

will always hold.

>compare the quantities of cart items and take the cart item with a different quantity.

This is a good solve!

lderrickable: @damienwebdev ok I'll go with taking the last item when possible then

libs/cart/src/effects/cart-item.effects.ts
griest024: Okay sounds good.

libs/cart/src/effects/cart-item.effects.ts
griest024: > I think:
> 
> > new cart item will always be the last
> 
> will always hold.

Isn't that platform specific?

damienwebdev: It is. In retrospect, I think we actually don't care about which one was added, we care about showing the user the updates to the items in their order, so we'd essentially be diffing ANY changes to the quantities and showing notices for those items.

lderrickable: So if it's platform specific, I should not do it, right? I should just be checking if it is there or if the quantity changed

griest024: I think we should be diffing the item quantities; it's the safer and more robust solution.

libs/cart/src/injection-tokens/cart-item-state-debounce-time.ts
```diff
@@ -0,0 +1,9 @@
+import { InjectionToken } from '@angular/core';
+
+/**
+ * An injection token for setting how long (in ms) a cart item remains in a temporary state before
+ * reverting back to a default state. These temporary states are set after mutating or adding a cart item.
+ * Daffodil has a timer that resets the state of all cart items after this debounce time, but the timer will
+ * reset if a new item addition or mutation occurs. The default is 4000.
```
griest024, line 7: Its unclear whether a new addition or mutation will cause all the timers to reset or only the timer that pertains to the affected cart item. I would think the timer would decay separately for each cart item in state but that doesn't seem to be how it works.

libs/cart/src/models/cart-item.ts
```diff
@@ -6,6 +6,7 @@ import { DaffCartItemInputType } from './cart-item-input';
 
 export interface DaffCartItem {
 	item_id: number | string;
+	state: DaffCartItemStateEnum;
```
griest024, line 9: Can you document this field and explain that its just for decaying (client-side) states?

libs/cart/src/selectors/cart-item-entities/cart-item-entities.selectors.spec.ts
```diff
@@ -163,4 +164,22 @@ describe('selectCartItemEntitiesState', () => {
 			expect(selector).toBeObservable(expected);
     });
   });
+
+  describe('selectCartItemState', () => {
+		
+		it('should return whether the given cart item is out of stock', () => {
```
griest024, line 170: Wrong spec message

libs/cart/src/effects/cart-item.effects.ts
```diff
@@ -65,29 +74,45 @@ export class DaffCartItemEffects<
...
-      this.driver.add(
-        this.storage.getCartId(),
-        action.input
-      ).pipe(
-        map((resp: V) => new DaffCartItemAddSuccess(resp)),
+      combineLatest([
+				this.store.pipe(
+					select(getDaffCartItemEntitiesSelectors<V, X, T>().selectAllCartItems),
+					take(1)
+				),
```
griest024, line 81: Do you think its okay to use the facade instead of the store here?

libs/cart/src/effects/cart-item.effects.ts
```diff
@@ -100,4 +125,17 @@ export class DaffCartItemEffects<
       )
     )
   )
+
+	private updateAddedCartItemState(oldCartItems: T[], newCartItems: T[]): T[] {
+		return newCartItems.map(newItem => {
+			const oldItem = oldCartItems.find(originalItem => originalItem.item_id === newItem.item_id);
+			return !oldItem || (oldItem && oldItem.qty !== newItem.qty) ?
```
griest024, line 132: Can you add an optional chaining TODO here?

libs/cart/src/effects/cart-item.effects.ts
```diff
@@ -65,29 +74,45 @@ export class DaffCartItemEffects<
...
+				catchError(error => of(new DaffCartItemUpdateFailure('Failed to update cart item')))
+			)
+		)
+	)
+
+	@Effect()
+  resetCartItemStateAfterChange$ = this.actions$.pipe(
+		ofType(DaffCartItemActionTypes.CartItemAddSuccessAction, DaffCartItemActionTypes.CartItemUpdateSuccessAction),
+		debounceTime(this.cartItemStateDebounceTime),
+		map(() => new DaffCartItemStateReset())
```
griest024, line 115: I would think that the reset would need to know the item ID. Maybe a global reset is the desired behavior here, I'm not sure.

libs/cart/src/effects/cart-item.effects.spec.ts
```diff
@@ -168,15 +194,47 @@ describe('Daffodil | Cart | CartItemEffects', () => {
...
+					items: [{
+						...mockCartItem,
+						qty: mockCartItem.qty + 1
+					}]
+				};
+				daffCartItemDriverSpy.add.and.returnValue(of(newCart));
+				const cartItemAddSuccessAction = new DaffCartItemAddSuccess({
+					...newCart,
+					items: [
+						{ ...newCart.items[0], state: DaffCartItemStateEnum.New }
```
griest024, line 233: I think it should show as updated when you add the same item to your cart. Isn't that effectively the same thing as updating the quantity?

griest024 (changes requested):

libs/cart/src/effects/cart-item.effects.ts
lderrickable: Yea, I explain in the PR description why I decided to go with a global reset instead of tracking the state changes individually.

libs/cart/src/injection-tokens/cart-item-state-debounce-time.ts
lderrickable: Yea, I explain in the PR description why I decided to go with a global reset instead of tracking the state changes individually. I'll take a look at the doc here and try to make it more clear though

griest024: oh sorry I didn't see that. I don't think the implementation would be that complex, but maybe I'm not thinking far enough ahead. Couldn't you just pass the `item_id` to the reset action? I think the reducer part of that would be straightfoward.

The only uncertainty I have is how to fork the stream such that each cart item add/update can be debounced separately.

lderrickable: Yea, I think the tough part is managing individual debounced streams. Imagine the scenario were cartItem1 is added, then cartItem2 is added quickly after (before the debounce timer is complete). I think this should certainly be possible with like some sort of queue or something. But then add in a third action of updating the quantity of cartItem1 directly after adding cartItem2, and we can no longer think of each cart item add/update as a completely separate debounce timer. In this scenario, if we treat each action as a separate debounce timer, the state of cartItem1 would go:
cartItem1 add -> cart item 1 state becomes "New" -> start cartItemAdd1 debounce timer
cartItem2 add -> cart item 2 state becomes "New" -> start cartItemAdd2 debounce timer
cartItem1 update qty -> cart item 1 state becomes "Mutating" then becomes "Updated" -> start cartItemUpdate1 debounce timer
cartItemAdd1 debounce timer ends -> cart item 1 state becomes "Default" prematurely when it should be "Updated" for the remaining cartItemUpdate1 debounce timer.

So instead, we would need to give each cart item a debounce timer that would somehow reset when a new action was performed on that cart item. I'm not saying this is impossible, but I compared this complexity to the downside of combining all cart item interactions into a single debounce timer and thought it was worth the trade off. That trade off is just that cartItem2 in the example above will remain in the "New" state for longer than 4 seconds. It will remain in that state until the cartItem1 update state (from the third action) finishes its 4 second debounce timer. And that doesn't seem that bad too me.

libs/cart/src/effects/cart-item.effects.spec.ts
lderrickable: That's a good question. I think it comes down to what takes precedence: what actually changes between the old state and new state, or which process is used to "update to qty". At first I agreed with you, but then I thought that maybe how the customer begins the process (in this case by adding a product to the cart instead of changing the quantity of a cart item) should determine the state of the cart item. Maybe it should have a "New" state even though all that essentially happened was an "update", because the customer will expect to have added a product to the cart instead of updating a cart item? What do you think?

griest024: I think the state change should be what determines updated vs. new as opposed to the process that generated it. As a customer, I would only expect to see a "new" notification when I added an item that was previously not in my cart.

That brings up some edge cases though. Consider the case where a customer has already added a composite product to their cart with required options A and B and nothing else. The customer then adds the same composite with required options A and B and optional option C. Does the cart item show 2 discrete items, one with A,B and the other with A,B,C or a single item with 2A,2B,C? And in this case, should that be new or updated?

griest024: >  As a customer, I would only expect to see a "new" notification when I added an item that was previously not in my cart.

@damienwebdev I would like your opinion on this UX.

lderrickable: In the composite product example, it would always show as a new cart item, since the "new" status is based on cart item ids and a composite product with different options is always a new cart item on the backend

lderrickable: I was thinking about this and I think you're right. I think it should depend on state change rather than customer intention. On the one hand, I don't think that you would add a whole other product just to update the qty of the cart item. You'd probably just update the cart item qty in the cart. But on the other hand, I was thinking that the only time you would add a product to the cart like that is if you'd forgotten you had already added the item. In that case, it would be more useful to show an "updated" state rather than a "new" state, because the "new" state might cause you to accidentally buy two of the same product unintentionally, while showing an "updated" state might signal to you that something unexpected is going on; i.e. you already had the item in your cart.

lderrickable: I changed it to what you proposed. We can wait to hear from Damien if you'd like though

libs/cart/src/effects/cart-item.effects.ts
lderrickable: added

libs/cart/src/selectors/cart-item-entities/cart-item-entities.selectors.spec.ts
lderrickable: fixed

libs/cart/src/models/cart-item.ts
lderrickable: added

libs/cart/src/effects/cart-item.effects.ts
lderrickable: Though I don't think it's that big of a deal, I think it's probably better practice not to use the facade, because the selectors are more essential than the facade. It just seems like a bad dependency chain, even though it would work.

^I wrote that, then I thought that there is really no dependency relationship between effects and the facade. That dependency is between the selectors and the facade so I think it should be ok to use the facade.

libs/cart/src/effects/cart-item.effects.spec.ts
griest024: Yeah I'd say give Damien a chance to look at it but the code lgtm

libs/cart/src/effects/cart-item.effects.ts
griest024: Its not a big deal either way I just think the facade is cleaner to interact with.

lderrickable: now using the facade. The tests were throwing weird errors when I used the `DaffCartTestingModule`, so I just mocked it out instead and it works. Not sure why

griest024: Its because of the injection token mismatch between `src` and `dist`. I rebased to get my recent spec tsconfig change and changed all of the imports to `@daffodil/cart` and the testing module works now.

I also exported the debounce timer injection token, which wasn't being exported before. I only noticed because I was importing from `@daffodil/cart`, that change is already paying off :D

libs/cart/src/effects/cart-item.effects.ts
```diff
@@ -100,4 +123,24 @@ export class DaffCartItemEffects<
...
+				//todo: add optional chaining when possible
+				case oldItem && oldItem.qty !== newItem.qty:
+					return { ...newItem, state: DaffCartItemStateEnum.Updated };
+				default:
+					return newItem;
+			}
+		})
+	}
+
+	private updateMutatedCartItemState(cartItems: T[], itemId: T['item_id']): T[] {
```
damienwebdev, line 142: This is clearly a reducer method.

libs/cart/src/effects/cart-item.effects.ts
```diff
@@ -100,4 +123,24 @@ export class DaffCartItemEffects<
       )
     )
   )
+
+	private updateAddedCartItemState(oldCartItems: T[], newCartItems: T[]): T[] {
```
damienwebdev, line 127: This is clearly a reducer method.

libs/cart/src/effects/cart-item.effects.ts
```diff
@@ -65,29 +72,45 @@ export class DaffCartItemEffects<
   add$ = this.actions$.pipe(
     ofType(DaffCartItemActionTypes.CartItemAddAction),
     switchMap((action: DaffCartItemAdd<U>) =>
-      this.driver.add(
-        this.storage.getCartId(),
-        action.input
-      ).pipe(
-        map((resp: V) => new DaffCartItemAddSuccess(resp)),
+      combineLatest([
+				this.facade.items$.pipe(
```
damienwebdev, line 76: This facade should likely NOT be used here, I think **generally** facade usage in an effect is an antipattern.

damienwebdev (changes requested):

libs/cart/src/effects/cart-item.effects.spec.ts
damienwebdev: I think the state isn't quite correct as "new". My original intent here is more of a "HEY YOU, THIS THING CHANGED" kind of state indicator.

E.g. if anything changed, we need to let the user know about it.

libs/cart/src/effects/cart-item.effects.ts
lderrickable: The reference to state here isn't needed anymore.

libs/cart/src/effects/cart-item.effects.ts
lderrickable: moved to reducers

libs/cart/src/effects/cart-item.effects.ts
lderrickable: moved to reducers

libs/cart/src/effects/cart-item.effects.spec.ts
lderrickable: I think we talked about this and resolved it. Let me know if that's not the case

lderrickable:
ready for review

damienwebdev (changes requested):

libs/cart/src/cart.module.ts
```diff
@@ -0,0 +1,38 @@
...
+		{ provide: DaffCartShippingMethodGuardRedirectUrl, useValue: '/' },
+		{ provide: DaffCartPaymentMethodGuardRedirectUrl, useValue: '/' },
+		{ provide: DaffCartOrderResultGuardRedirectUrl, useValue: '/' },
+		{ provide: DaffEmptyCartResolverRedirectUrl, useValue: '/' },
+		{ provide: DaffCartResolverRedirectUrl, useValue: '/' },
+		{ provide: DaffCartPaymentMethodIdMap, useValue: {} },
+		{ provide: DaffCartItemStateDebounceTime, useValue: 4000 }
+	]
+})
+export class DaffCartModule { }
```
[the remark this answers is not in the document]
lderrickable: no, removed

libs/cart/state/src/models/stateful-cart-item.ts
```diff
@@ -0,0 +1,18 @@
+import { DaffCartItem } from '@daffodil/cart';
+
+/**
+ * The state of the cart item is intended to enhance the client side UX like indicating when a cart
+ * item was recently added/updated. For states that indicate the completion of some process, the state is given 
+ * a decay time based on the DaffCartItemStateDebounceTime injection token. For example when a cart item is
+ * added to the cart, the state of that item will be "New" for a designated time then will revert to the default state.
+ */
+export interface DaffStatefulCartItem extends DaffCartItem {
+	daffState: DaffCartItemStateEnum
```
griest024, line 10: Should this field not just be named `state`? Whats your reasoning for prefixing it with `daff`?

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
```diff
@@ -115,7 +117,16 @@ const createCartItemEntitiesSelectors = <
 		}
 	);
 
-	return {
+	const selectCartItemState = createSelector(
+		selectCartItemEntities,
+		(cartItems, props) => {
+			const cartItem = selectCartItem.projector(cartItems, { id: props.id });
+
+			return cartItem ? cartItem.daffState : null;
```
griest024, line 125: optional chaining comment

griest024 (changes requested):
You should deprecate the original items selector and facade field.

libs/cart/src/cart.module.ts
```diff
@@ -0,0 +1,38 @@
...
+import { 
+	DaffCartStateModule,
+	DaffCartBillingAddressGuardRedirectUrl,
+	DaffCartShippingAddressGuardRedirectUrl,
+	DaffCartShippingMethodGuardRedirectUrl,
+  DaffCartPaymentMethodGuardRedirectUrl,
+	DaffCartOrderResultGuardRedirectUrl,
+	DaffEmptyCartResolverRedirectUrl, 
+	DaffCartResolverRedirectUrl
+} from '@daffodil/cart/state';
```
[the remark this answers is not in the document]
griest024: mb, already resolved.

libs/cart/state/src/models/stateful-cart-item.ts
```diff
@@ -0,0 +1,18 @@
+import { DaffCartItem } from '@daffodil/cart';
+
+/**
+ * The state of the cart item is intended to enhance the client side UX like indicating when a cart
+ * item was recently added/updated. For states that indicate the completion of some process, the state is given 
+ * a decay time based on the DaffCartItemStateDebounceTime injection token. For example when a cart item is
+ * added to the cart, the state of that item will be "New" for a designated time then will revert to the default state.
+ */
+export interface DaffStatefulCartItem extends DaffCartItem {
```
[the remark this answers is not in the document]
lderrickable: In my mind, `DaffCartItemEntity` just says that the model is held in ngrx entity state. But anything that can be stored as an entity should be stored as an entity, so I don't think `DaffCartItemEntity` really tells you any extra information, because cart items should be expected to be an entity. On the other hand, `DaffStatefulCartItem` says that it is a cart item that has a state. Is there another reason you prefer `DaffCartItemEntity`?

libs/cart/state/src/actions/cart-item.actions.ts
```diff
@@ -108,6 +109,10 @@ export class DaffCartItemDeleteFailure implements Action {
   constructor(public payload: string) {}
 }
 
+export class DaffCartItemStateReset implements Action {
+	readonly type = DaffCartItemActionTypes.CartItemStateResetAction;
+}
+
 export type DaffCartItemActions<
   T extends DaffCartItem = DaffCartItem,
```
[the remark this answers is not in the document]
lderrickable: I didn't make those `DaffStatefulCartItem`s because the stateful stuff only gets added to the model in the reducers/selectors layer. The actions and effects only see `DaffCartItem`s. For that reason, I think the actions and effects should only deal with `DaffCartItem`s

lderrickable:
> You should deprecate the original items selector and facade field.

I deprecated the selector, but the facade field will be used in the future. It just needs to be pointed at the cart item entity selector instead

libs/cart/state/src/models/stateful-cart-item.ts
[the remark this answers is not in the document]
griest024: No actually I was misunderstanding what you meant by "stateful". I agree that your name is more informative and therefore better. However, the terminology gets confusing with the entity's state and ngrx state. What do you think about "status" instead of "state"?

libs/cart/state/src/actions/cart-item.actions.ts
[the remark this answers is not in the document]
griest024: What if some effects reference some selectors? They would pass in a generic `T extends DaffCartItem` that can't satisfy the selectors generic constraint of `T extends DaffStatefulCartItem`.

griest024:
> I deprecated the selector, but the facade field will be used in the future. It just needs to be pointed at the cart item entity selector instead

Sounds good, can you make that change now or add a TODO for it?

libs/cart/state/src/models/stateful-cart-item.ts
griest024: ah that makes sense

libs/cart/state/src/models/stateful-cart-item.ts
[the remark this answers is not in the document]
lderrickable: oh, I see where the misunderstanding was. 

"status" is fine with me. I'll make that change

libs/cart/state/src/actions/cart-item.actions.ts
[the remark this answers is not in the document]
lderrickable: In that case, I guess we'd need to have `T extends DaffStatefulCartItem` and `V extends DaffCartItem`

lderrickable:
> > I deprecated the selector, but the facade field will be used in the future. It just needs to be pointed at the cart item entity selector instead
> 
> Sounds good, can you make that change now or add a TODO for it?

I made a TODO. I want to do it when we make the decision about whether or not the cart.items part of the state will be trimmed out or not.

libs/cart/state/src/models/stateful-cart-item.ts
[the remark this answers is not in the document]
lderrickable: changed

libs/cart/state/src/actions/cart-item.actions.ts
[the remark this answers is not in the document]
griest024: I think we should make the decision now whether or not the effects layer in general can be dependent on the selector layer. I think so, and therefore the effects layer should, in general, be dealing with the same models that the selector layer is even if it doesn't actually touch any of the fields of that model.

I'm in favor of replaciing `DaffCartItem` across the entire state package with `DaffStatefulCartItem`. I think we open ourselves up to covariance issues later on if we don't.

libs/cart/state/src/models/stateful-cart-item.ts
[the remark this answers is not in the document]
griest024: I just realized that we have a `DaffStatusEnum` in the design package. We should think about if this conflict is a problem. Should we change the name of `DaffStatefulCartItem`? How about `DaffStatusableCartItem`?

libs/cart/state/src/actions/cart-item.actions.ts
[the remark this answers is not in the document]
lderrickable: done

libs/cart/state/src/models/stateful-cart-item.ts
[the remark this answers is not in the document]
lderrickable: reverted this back to stateful and daffState

libs/cart/state/testing/src/factories/stateful-composite-cart-item.factory.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@angular/core';
+
+import { DaffModelFactory } from '@daffodil/core/testing';
+import { DaffMockCompositeCartItem } from '@daffodil/cart/testing';
+import { DaffStatefulCartItem, DaffCartItemStateEnum, DaffStatefulCompositeCartItem } from '@daffodil/cart/state';
+
+export class DaffMockStatefulCompositeCartItem extends DaffMockCompositeCartItem implements DaffStatefulCartItem {
```
griest024, line 7: Should implement `DaffStatefulCompositeCartItem`.

libs/cart/state/testing/src/factories/stateful-configurable-cart-item.factory.ts
```diff
@@ -0,0 +1,19 @@
+import { Injectable } from '@angular/core';
+
+import { DaffModelFactory } from '@daffodil/core/testing';
+import { DaffMockConfigurableCartItem } from '@daffodil/cart/testing';
+import { DaffStatefulCartItem, DaffCartItemStateEnum, DaffStatefulConfigurableCartItem } from '@daffodil/cart/state';
+
+export class DaffMockStatefulConfigurableCartItem extends DaffMockConfigurableCartItem implements DaffStatefulCartItem {
```
griest024, line 7: Should implement `DaffStatefulConfigurableCartItem`

griest024 (changes requested):

libs/cart/state/testing/src/factories/stateful-composite-cart-item.factory.ts
lderrickable: fixed

libs/cart/state/testing/src/factories/stateful-configurable-cart-item.factory.ts
lderrickable: fixed

griest024 (approved):

libs/cart/src/injection-tokens/public_api.ts
```diff
@@ -1,2 +1,3 @@
 export { DaffCartPaymentMethodIdMap } from './payment-method-map';
 export { DAFF_CART_ERROR_MATCHER } from './error-matcher.token';
+export { DaffCartItemStateDebounceTime } from './cart-item-state-debounce-time';
```
damienwebdev, line 3: Shouldn't this token be in `state`?

damienwebdev (changes requested):

libs/cart/src/injection-tokens/public_api.ts
lderrickable: yes, moved

damienwebdev (approved):

PR #1135, opened 2020-11-10
feat(cart): cart item feature loading state remains the same when a c…

lderrickable:
…art item is updated or deleted

## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The cart item feature area loads whenever a cart item is mutated or deleted.

## What is the new behavior?
The cart item feature area doesn't load when a cart item is mutated or deleted, so that individual cart item states may be shown instead.

```
[ ] Yes
[x] No
```


lderrickable:
@griest024 @damienwebdev See if you guys agree with how I'm handling the two edge cases I describe in the docs I added.

libs/cart/guides/loading-states.md
```diff
@@ -0,0 +1,11 @@
+# Cart Loading States
+## Cart Feature Areas
+This doc explains how the Daffodil cart loading states work, and how they can be used to create a better user experience. The cart loading states are split into different feature areas, and there are three loading states possible for each of these features: loading, mutating, and complete. The purpose of splitting loading states into feature areas is to allow the app developer to show a loading state only for the parts of the cart that are actually changing rather than showing the entire cart as loading every time a user updates any part of the cart. Additionally, for the case that the app developer does want to show a loading state every time anything related to the cart changes, there is also a loading state for that too. For more details on each particular feature area, see the DaffCartFacadeInterface.
```
griest024, line 3:
```
Additionally, for the case that the app developer does want to show a loading state every time anything related to the cart changes, there is also a loading state for that too
```
->
```
Additionally, for the case that the app developer does want to show a loading state when anything related to the cart changes, the cart state library provides `featureLoading`, `featureMutating`, and `featureResolving`
```

libs/cart/guides/loading-states.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Cart Item States
+There are times when the app developer not only wants to show that the cart item feature area is loading, but that a particular cart item is loading or has a particular state. For example, one might want to indicate that a recently added cart item was recently added, or that a cart item whose quantity is being changed is mutating. For these use cases daffodil provides states for each cart item. The possible states for a cart item are new, mutating, updated, and default.
```
griest024, line 6:
```
For example, one might want to indicate that a recently added cart item was recently added, or that a cart item whose quantity is being changed is mutating.
```
->
```
For example, one might want to indicate that a cart item was recently added or that a cart item whose quantity is being changed is mutating.
```

libs/cart/guides/loading-states.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Edge Cases
+While we try to leave as much creative freedom as possible in the hands of the app developer, there are some edge cases on which we needed to make a decision in order for this state system to work. Here are a list of edge cases where it isn't very obvious how daffodil state would work:
```
griest024, line 9:
```
While we try to leave as much creative freedom as possible in the hands of the app developer, there are some edge cases on which we needed to make a decision in order for this state system to work. Here are a list of edge cases where it isn't very obvious how daffodil state would work
```
-> 
```
While we try to leave as much creative freedom as possible in the hands of the app developer, there are some edge cases for which the cart item loading state system must be strongly opinionated. The following list attempts to clarify some ambiguous edge cases:
```

libs/cart/guides/loading-states.md
```diff
@@ -0,0 +1,11 @@
...
...developer to show loading states for particular cart items that are being mutated or deleted, while avoiding the complexity that comes with trying to show individual loading states for cart items being added to the cart. Some parts of this complexity are questions like: How do I indicate to the customer that a new cart item is being added to the cart if that cart item does not exist in state until after the item has been added? Possible to solve, but ugly. How do I discern if a composite product being added already exists on the cart or if it will be a new cart item? The difficulty here is that the variantId is needed to add a composite product to the cart to distinguish it from other configurations of the same composite product, but this variantId isn't returned on the DaffCompositeCartItem. The only solves to this problem are either to require an additional field (an array of the selected options) in the DaffCartItemAdd action, or to change the backend to return the variantId associated with a composite cart item. Neither of these are particularly attractive solutions, so we elected to simply cause the entire Cart Item Feature area to load whenever a product is added to the cart.
```
griest024, line 10: `the cart items feature state` -> `the cart item's feature state`

griest024, line 10:
```
Some parts of this complexity are questions like: How do I indicate to the customer that a new cart item is being added to the cart if that cart item does not exist in state until after the item has been added? Possible to solve, but ugly. How do I discern if a composite product being added already exists on the cart or if it will be a new cart item?
```

Wrap these questions with quotations: `"`.

griest024, line 10: `The only solves to this problem are either to require an additional field` -> `The only solutions to this problem are either to require an additional field`

libs/cart/src/reducers/cart/cart.reducer.ts
```diff
@@ -30,10 +30,21 @@ export function cartReducer<T extends DaffCart>(
     case DaffCartActionTypes.CartClearAction:
     case DaffCartActionTypes.AddToCartAction:
     case DaffCartActionTypes.CartCreateAction:
+		case DaffCartItemActionTypes.CartItemUpdateAction:
+		case DaffCartItemActionTypes.CartItemDeleteAction:
```
griest024, line 34: So item update or delete will cause the cart feature slice to be mutating? I don't understand that behavior.

griest024 (changes requested):

libs/cart/src/reducers/cart/cart.reducer.ts
lderrickable: All cart related actions cause the cart feature area to be mutating, but they usually do it through their respective feature areas. The cart feature loading derives its state from the other loading states. But because the cart item update and delete actions no longer trigger the cart item feature area to be loading, neither will they trigger the cart feature area to load. So I needed to trigger the cart feature loading manually with these two action streams.

libs/cart/guides/loading-states.md
lderrickable: changed

lderrickable: done

lderrickable: I actually did that on purpose to differentiate between the loading state of the entire cart items feature area and the particular state of individual cart items. But I switched it around to `the feature state of the cart items` to make it more clear.

libs/cart/guides/loading-states.md
lderrickable: changed, but edited it a bit

libs/cart/guides/loading-states.md
lderrickable: changed

libs/cart/guides/loading-states.md
lderrickable: changed

lderrickable:
@griest024 All of this will be replaced with a new PR. Damien and I talked about it and came up with a better solution

libs/cart/src/reducers/cart/cart.reducer.ts
lderrickable: Also, I was wrong here. I was under the impression that I was adding to the cart feature reducer, when that reducer doesn't actually exist.

lderrickable:
Here is the new PR: https://github.com/graycoreio/daffodil/pull/1140

PR #1142, opened 2020-11-13
Category page server side resolver

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
This is a resolver that allows the client-side code to render a loading state for the category page, while keeping the server-side code from rendering until after the category data is available. This way, web crawlers will receive a fully loaded page (needed for seo audits), while client-side users will see a loading state in the meantime.


libs/category/src/resolvers/category-page-server-side/category-page-server-side.resolver.ts
```diff
@@ -0,0 +1,43 @@
...
+    private store: Store<DaffCategoryReducersState>,
+    private dispatcher: ActionsSubject,
+	) {}
+	
+	resolve(route: ActivatedRouteSnapshot): Observable<boolean> {
+		this.store.dispatch(new DaffCategoryLoad({
+      id: route.paramMap.get('id'), page_size: this.defaultCategoryPageSize
+		}));
+
+		return isPlatformBrowser(this.platformId) ?
```
griest024, line 32: Should the resolver care about whether its on the server or not? Should it be up to the app dev to selectively invoke it on the server vs the client? I feel like a resolver should be environment-agnostic.

griest024: Or, at the very least, there should be an environment-agnostic resolver that this server-specific resolver can wrap and invoke.

lderrickable: changed

libs/category/src/category-state.module.ts
```diff
@@ -4,11 +4,15 @@ import { EffectsModule } from '@ngrx/effects';
...
+import { DaffDefaultCategoryPageSize } from './resolvers/public_api';
 
 @NgModule({
   imports: [
     StoreModule.forFeature('category', daffCategoryReducers),
     EffectsModule.forFeature([DaffCategoryEffects]),
-  ]
+	],
+	providers: [
+		{ provide: DaffDefaultCategoryPageSize, useValue: 12 }
```
griest024, line 15: Not saying that you should do this, but default values can be passed to the injection token constructor. Just FYI in case you find that a better way to provide default token values.

libs/category/src/resolvers/category-page/category-page.resolver.ts
```diff
@@ -0,0 +1,39 @@
...
+	resolve(route: ActivatedRouteSnapshot): Observable<boolean> {
+		this.store.dispatch(new DaffCategoryLoad({
+      id: route.paramMap.get('id'), page_size: this.defaultCategoryPageSize
+		}));
+
+		return this.dispatcher.pipe(
+			filter((action: DaffCategoryLoadSuccess | DaffCategoryLoadFailure) => 
+				action.type === DaffCategoryActionTypes.CategoryLoadSuccessAction
+				|| action.type === DaffCategoryActionTypes.CategoryLoadFailureAction
+			),
```
griest024, line 34: Same here. If `ofType` works, you should use it here.

griest024 (changes requested):

libs/category/src/resolvers/category-page/category-page.resolver.ts
lderrickable: changed

libs/category/src/category-state.module.ts
lderrickable: I'm good with as is

griest024 (approved):

PR #1150, opened 2020-11-17
feat(navigation): injectable category tree query depth in magento driver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The navigation tree query depth is hardcoded and unconfigurable.

## What is the new behavior?
The navigation tree query depth is configurable by an injection token.

```
[ ] Yes
[x] No
```

## Other information
The implementation is non trivial. In order to handle a configurable number of nested fragments, they had to be generated dynamically with different names. The fragment name must be returned separately from the fragment so it can be referenced dynamically in the enclosing gql doc. This resulted in an interface and a couple functions for generating the fragments.

### DIFF
--- a/libs/navigation/src/drivers/injection-tokens/category-tree-query-depth.token.ts
+++ b/libs/navigation/src/drivers/injection-tokens/category-tree-query-depth.token.ts
@@ -1,0 +1,8 @@
+import { InjectionToken } from '@angular/core';
+
+/**
+ * The maximum depth of category children that the navigation driver will query.
+ * Defaults to 3.
+ */
+export const DaffNavigationCategoryTreeQueryDepth =
+  new InjectionToken<number>('DaffNavigationCategoryTreeQueryDepth', {factory: () => 3});
--- a/libs/navigation/src/drivers/injection-tokens/category-tree-query-depth.token.ts
+++ b/libs/navigation/src/drivers/injection-tokens/category-tree-query-depth.token.ts
@@ -1,8 +1,0 @@
-import { InjectionToken } from '@angular/core';
-
-/**
- * The maximum depth of category children that the navigation driver will query.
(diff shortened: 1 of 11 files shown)


damienwebdev:
@griest024 code looks ok to me, but there's a test that needs fixing.

griest024:
@damienwebdev I just need to rebase to get the spec tsconfig

libs/navigation/src/drivers/injection-tokens/category-tree-query-depth.token.ts
```diff
@@ -0,0 +1,8 @@
+import { InjectionToken } from '@angular/core';
+
+/**
+ * The maximum depth of category children that the navigation driver will query.
+ * Defaults to 3.
+ */
+export const DaffNavigationCategoryTreeQueryDepth =
```
damienwebdev, line 7: This is platform specific (not all platforms are graphql or even configurable). We should add this as a config property of the driver module's `.forRoot()` and use the injectable config object.

damienwebdev (changes requested):

libs/navigation/src/drivers/injection-tokens/category-tree-query-depth.token.ts
griest024: good point, although I left the token the same and pulled the value off the config instead of injecting the entire object

griest024: changed

damienwebdev: LGTM

damienwebdev (approved):

PR #1151, opened 2020-11-17
feat(category): add page load action

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

## What is the new behavior?
- Renames the "category load" actions to "category page load".
- Renames `DaffCategoryEffects` to `DaffCategoryPageEffects`.
- Adds `DaffCategoryLoad` actions that do not save category config state.
- Adds `DaffCategoryEffects` that handle just category load without filters or sorting.

```
[x] Yes
[ ] No
```

Change references of `DaffCategoryLoad*` to `DaffCategoryPageLoad*` unless you *specifically* want the behavior that does not touch config or handle filters and sorting.

## Other information

### DIFF
--- a/libs/category/src/effects/category-page.effects.spec.ts
+++ b/libs/category/src/effects/category-page.effects.spec.ts
@@ -147,44 +147,16 @@ describe('DaffCategoryPageEffects', () => {
     });
 	});
 
-	describe('validateFilters', () => {
-
-		it('should throw an error when an invalid range type format is given', () => {
-			expect(() => {
-				effects.validateFilters([{
-					name: 'price',
-					value: ['invalidFormat'],
-					type: DaffCategoryFilterType.Range
-				}])
-			}).toThrowError('Category range filter is in an invalid format. Should be **-**');
-		});
-
-		it('should throw an error when two filters share the same name', () => {
-			expect(() => {
-				effects.validateFilters([{
-					name: 'name',
-					value: ['value'],
-					type: DaffCategoryFilterType.Equal
-				},
-				{
-					name: 'name',
-					value: ['value2'],
-					type: DaffCategoryFilterType.Equal
-				}
-			])
-			}).toThrowError('More than one category filter of the same name exists. These should' +
-						' be combined into a single filter action with multiple values.');
-		});
-	});
-
   describe('when CategoryPageLoadAction is triggered', () => {
     let categoryPageLoadSuccessAction: DaffCategoryPageLoadSuccess;
 		let expected;
 		let categoryPageLoadAction;
+    let categoryRequest: DaffCategoryRequest;
 
 		beforeEach(() => {
-       categoryPageLoadAction = new DaffCategoryPageLoad({id: stubCategory.id});
-       categoryPageLoadSuccessAction = new DaffCategoryPageLoadSuccess({
+      categoryRequest = {id: stubCategory.id};
+      categoryPageLoadAction = new DaffCategoryPageLoad(categoryRequest);
+      categoryPageLoadSuccessAction = new DaffCategoryPageLoadSuccess({
         category: stubCategory,
         categoryPageConfigurationState: stubCategoryPageConfigurationState,
         products: stubProducts
@@ -221,7 +193,7 @@ describe('DaffCategoryPageEffects', () => {
       });
     });
 
-    it('should call get category with only an id', () => {
+    it('should call get category with the category request from the action payload', () => {
 			spyOn(daffCategoryDriver, 'get').and.returnValue(of({
 				category: stubCategory,
 				categoryPageConfigurationState: stubCategoryPageConfigurationState,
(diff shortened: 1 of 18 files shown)


lderrickable:
What if instead, the `DaffCategoryPageLoad` action just triggers the `DaffCategoryLoad`, and we add a substream to watch for `DaffCategoryPageLoad` and `DaffCategoryLoadSuccess` to fire the `DaffCategoryPageLoadSuccess` action. That way we don't need any of the reducer changes.

griest024:
@lderrickable I thought about that, but that would be bad action hygeine. If we were to do that though, that could just be implemented in the app.

lderrickable:
That's true. Maybe it should just be in the app? The only good thing about having it in daffodil is that it might save the app dev some time. But it seems like it should really go in the app. @damienwebdev , do you have an opinion?

damienwebdev:
I strongly believe in the two distinct and separate actions. Action hygiene in this case is worth minor changes to reducer behavior.

I think, and a subsequent PR here is likely necessary, that the original actions behavior in certain effects/certain reducers, should be removed. 

Substream is a currently necessary evil of an architectural weakness, that I would like to move away from as time goes on. 

Any chance we can get to avoid a substream use should be preferred.

griest024:
> that the original actions behavior in certain effects/certain reducers, should be removed

@damienwebdev the original actions being `DaffCategoryLoad` and the corresponding success and failure? How would the effect or reducer behavior change? I was under the impression that the behavior associated with these actions would be identical apart from the action names.

lderrickable:
> "I strongly believe in the two distinct and separate actions."

@damienwebdev Could you explain why? I think it should go in the app, because if we ever do figure out how to fix the architectural problem, we will remove the DaffCategoryPageLoad actions and presumably the app dev will be able somehow to inject their own actions that take the place of a DaffCategoryLoad. I think the upgrade process for an app dev will be more straightforward if all of the places that they can now inject their new actions are currently covered by substream. If we provide some actions like DaffCategoryPageLoad that are separate patches for the same architectural problem, then the upgrade notes will be more complicated and seemingly arbitrary. 

Upgrade notes if we just use substream to patch this problem:
`Actions can now be injected into the daffodil action stream. To upgrade, simply replace all instances where substream is used with {insert solution here}`

Upgrade notes with this PR:
`Action can now be injected into the daffodil action stream. To upgrade, replace all instances where substream is used with {insert solution here}. Also, replace the deprecated `DaffCategoryPageLoad` and DaffProductPageLoad` actions with your own injected actions.`

griest024:
Changed and ready for review.

libs/category/src/effects/category.effects.spec.ts
```diff
@@ -103,229 +96,61 @@ describe('DaffCategoryEffects', () => {
...
-			}).toThrowError('More than one category filter of the same name exists. These should' +
-						' be combined into a single filter action with multiple values.');
-		});
-	});
 
-  describe('when CategoryLoadAction is triggered', () => {
-
-    let expected;
-    
     it('should call get category with only an id', () => {
```
lderrickable, line 143: This test isn't quite correct. It can get a category with all of the other inputs, but they need to be passed through the action instead of through state.

libs/category/src/effects/category-page.effects.spec.ts
```diff
@@ -0,0 +1,369 @@
...
+        const categoryPageLoadFailureAction = new DaffCategoryPageLoadFailure(error);
+        expected = cold('--b', { b: categoryPageLoadFailureAction });
+      });
+
+      it('should dispatch a CategoryPageLoadFailure action', () => {
+        expect(effects.loadCategoryPage$).toBeObservable(expected);
+      });
+    });
+
+    it('should call get category with only an id', () => {
```
lderrickable, line 224: I wrote this test originally, but I think it isn't quite right. It should be that it calls get category with only the request that is passed and not with any of the configurations in state.

lderrickable (changes requested):

libs/category/src/effects/category.effects.ts
```diff
@@ -38,125 +29,17 @@ export class DaffCategoryEffects<
...
-						' be combined into a single filter action with multiple values.')
-				}
-			}
-		})
-	}
-
-  private processCategoryGetRequest(payload: T) {
-    return this.driver.get(payload).pipe(
-      switchMap((resp: DaffGetCategoryResponse<T, V, U, W>) => [
+    switchMap((action: DaffCategoryLoad<T>) => this.driver.get(action.request).pipe(
```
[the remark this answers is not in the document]
griest024: done

libs/category/src/effects/category.effects.spec.ts
griest024: I think I fixed this

libs/category/src/effects/category-page.effects.spec.ts
griest024: I think I fixed this

lderrickable (approved):

PR #1158, opened 2020-11-23
fix(product): out of stock item options result in a price bug for non…

lderrickable:
…-one, unchangeable item option quantity products

## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

For a composite product with item options that have non-one, unchangeable quantities, if the default option is out of stock, the prices shown will not be multiplied by the option quantity as they should be.

## What is the new behavior?
For this scenario, the prices returned will be a price range based on the quantity of the default option.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/product/src/selectors/composite-product/composite-product.selectors.ts
+++ b/libs/product/src/selectors/composite-product/composite-product.selectors.ts
@@ -47,6 +47,7 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
 			}
 
 			const appliedOptions = props.configuration ? getAppliedOptionsForConfiguration(<DaffCompositeProduct>product, props.configuration) : {};
+			console.log(appliedOptions);
 			return {
 				minPrice: getMinPricesForConfiguration(<DaffCompositeProduct>product, appliedOptions),
 				maxPrice: getMaxPricesForConfiguration(<DaffCompositeProduct>product, appliedOptions)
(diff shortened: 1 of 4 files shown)


libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -47,6 +47,7 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
 			}
 
 			const appliedOptions = props.configuration ? getAppliedOptionsForConfiguration(<DaffCompositeProduct>product, props.configuration) : {};
+			console.log(appliedOptions);
```
griest024, line 50: debug log

libs/product/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -209,6 +210,7 @@ function getMaxPricesForConfigurationIncludingOptionalItems(product: DaffComposi
  * @param configuration a Dictionary<DaffCompositeConfigurationItem> used to build the appliedOptions object.
  */
 function getAppliedOptionsForConfiguration(product: DaffCompositeProduct, configuration: Dictionary<DaffCompositeConfigurationItem>): Dictionary<DaffCompositeProductItemOption> {
+	console.log(configuration);
```
griest024, line 213: debug log

griest024 (changes requested):

libs/product/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: removed

libs/product/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: removed

griest024 (approved):

PR #1164, opened 2020-11-25
feat(cart): add non empty cart guard

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[x] No
```

## Other information


damienwebdev:
The naming here is a bit odd, what does a NonEmptyCartGuard do? Does it guard against any empty cart, does it guard to ensure that the cart is empty?

griest024:
@damienwebdev  It guards against an empty cart, i.e., it will emit `true` if the cart has items in it, `false` if its empty. I agree the name could be better; I'm open to suggestions.

lderrickable:
What about just `EmptyCartGuard`? When I think of "guard", I think of something guarding one thing against something else. The "one thing" is just the page that is being routed, so I think the naming of any guard should be the "something else" being guarded against, which would be an empty cart. So `EmptyCartGuard`

griest024:
@lderrickable hm well I don't disagree but our other guards really aren't named like that.

lderrickable:
lol true

lderrickable:
I guess we could either call this one `DaffCartItemsGuard` or change the other guards to `DaffEmptyBillingAddressGuard`

griest024:
I vote for `DaffCartItemsGuard`

damienwebdev:
DaffVerifyCartHasItemsGuard?

griest024:
> DaffVerifyCartHasItemsGuard?

While I'm not totally against that, I prefer `DaffCartItemsGuard`. Its the same pattern as our other guards. Whatever we choose should be consistent across all of the guards.

damienwebdev:
`DaffCartItemsGuard` is fine along with a nice doc-block.

griest024:
changed

libs/cart/state/src/guards/non-empty-cart/non-empty-cart.guard.ts
```diff
@@ -0,0 +1,38 @@
+import { CanActivate, Router } from '@angular/router';
```
lderrickable, line 1: The file name and path needs to be changed to `cart-item` from `non-empty-cart`

griest024: good catch, changed

lderrickable (approved):

damienwebdev (approved):
Lgtm

PR #1165, opened 2020-11-27
fix(cart): cart item state is set incorrectly in cart item reducers

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The cart item state reducers aren't correct, and sometimes even set "state" on the cart item entity instead of "daffState".

```
[ ] Yes
[x] No
```

## Notes
The reason the tests were passing before is because the `DaffCartItemStateFactory` was setting the `daffState` incorrectly. I fixed the factory here.

### DIFF
--- a/libs/cart/state/src/reducers/cart-item-entities/cart-item-entities.reducer.ts
+++ b/libs/cart/state/src/reducers/cart-item-entities/cart-item-entities.reducer.ts
@@ -18,14 +18,20 @@ export function daffCartItemEntitiesReducer<
 	U extends DaffCartItemInput = DaffCartItemInput,
 	V extends DaffCart = DaffCart,
 >(
-  state = daffCartItemEntitiesAdapter<T>().getInitialState({ daffState: DaffCartItemStateEnum.Default }),
+  state = daffCartItemEntitiesAdapter<T>().getInitialState(),
   action: DaffCartItemActions<T, U, V> | DaffCartActions<V>): EntityState<T> {
 	const adapter = daffCartItemEntitiesAdapter<T>();
   switch (action.type) {
     case DaffCartItemActionTypes.CartItemListSuccessAction:
-			return adapter.addAll(action.payload, state);
+			return adapter.addAll(action.payload.map(item => ({
+				...item,
+				daffState: DaffCartItemStateEnum.Default
+			})), state);
 		case DaffCartItemActionTypes.CartItemLoadSuccessAction:
-			return adapter.upsertOne(action.payload, state);
+			return adapter.upsertOne({
+				...action.payload,
+				daffState: DaffCartItemStateEnum.Default
+			}, state);
 		case DaffCartItemActionTypes.CartItemAddSuccessAction:
 			return adapter.addAll(
 				updateAddedCartItemState<T>(state.entities, <T[]>action.payload.items), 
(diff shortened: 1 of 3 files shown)


libs/cart/state/src/reducers/cart-item-entities/cart-item-entities.reducer.ts
```diff
@@ -18,14 +18,20 @@ export function daffCartItemEntitiesReducer<
 	U extends DaffCartItemInput = DaffCartItemInput,
 	V extends DaffCart = DaffCart,
 >(
-  state = daffCartItemEntitiesAdapter<T>().getInitialState({ daffState: DaffCartItemStateEnum.Default }),
+  state = daffCartItemEntitiesAdapter<T>().getInitialState(),
```
griest024, line 21: Why did you remove this arg? Did it not set a default value for `daffState`?

libs/cart/state/src/reducers/cart-item-entities/cart-item-entities.reducer.ts
```diff
@@ -18,14 +18,20 @@ export function daffCartItemEntitiesReducer<
...
+			return adapter.addAll(action.payload.map(item => ({
+				...item,
+				daffState: DaffCartItemStateEnum.Default
+			})), state);
 		case DaffCartItemActionTypes.CartItemLoadSuccessAction:
-			return adapter.upsertOne(action.payload, state);
+			return adapter.upsertOne({
+				...action.payload,
+				daffState: DaffCartItemStateEnum.Default
+			}, state);
```
griest024, line 34: Are you sure we want to reset the state on item loads?

griest024 (changes requested):

libs/cart/state/src/reducers/cart-item-entities/cart-item-entities.reducer.ts
lderrickable: Turns out it just did this:
```
{
    entities: {},
    ids: [],
    daffState: "default"
}
```
I replaced what this tried to do with setting `daffState: "default"` on each cart item on a `DaffCartLoadSuccess` and a `DaffCartItemListSuccess`

libs/cart/state/src/reducers/cart-item-entities/cart-item-entities.reducer.ts
lderrickable: I think it should be ok, because the `DaffCartLoad`, `DaffCartItemLoad`, and `DaffCartItemList` actions should only be called on initial state loading. I maybe could go without it, but the initial cart items daffState would be undefined until some kind of mutating action happened to them, which could be a problem if an app dev has a `daffState === "default"` check somewhere

griest024: > I think it should be ok, because the DaffCartLoad, DaffCartItemLoad, and DaffCartItemList actions should only be called on initial state loading

I think that largely depends on the app dev, we don't decide when this happens.

I know its more verbose, but I think `daffState: action.payload.daffState || DaffCartItemStateEnum.Default` is safer.

lderrickable: daffState will never exist on a DaffCartLoadSuccess.payload.items[], because it doesn't exist in the platform. 

`daffState: action.payload.daffState || DaffCartItemStateEnum.Default` will always result in `daffState: DaffCartItemStateEnum.Default`

griest024: oh yeah...so you would have to do something like:
```ts
adapter.upsertOne({
	...action.payload,
	daffState: state.entities[action.payload.id].daffState || DaffCartItemStateEnum.Default
}, state);
```

not great...a slightly more robust pattern that works for all default values would be something like:
```ts
const defaultStatefulCartItem: Partial<DaffStatefulCartItem> = {
  daffState: DaffCartItemStateEnum.Default,
}
...
adapter.upsertOne({
  ...defaultStatefulCartItem,
  ...state.entities[action.payload].id,
  ...action.payload,
}, state);
```

I'm not sure if its worth it, but I still don't really like the idea of resetting the state for immutable loads.

lderrickable: `state.entities[action.payload.id].daffState || DaffCartItemStateEnum.Default` makes sense to me. Definitely better. Changed

griest024 (approved):

PR #1168, opened 2020-11-27
fix(navigation): the getCategoryNodeFragment magento query does not work

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The magento navigation query has a small problem that prevents it from working. Apparently you can't pass a fragment string into another function like this:
```
export function getCategoryNodeFragment(depth: number = 3): DocumentNode {
  const fragmentBody = new Array(depth).fill(null).reduce(acc => `
    ...categoryNode
    children_count
    children {
      ${acc}
    }
  `, '...categoryNode')

  return gql`
    fragment recursiveCategoryNode on CategoryTree {
      ${fragmentBody}
    }
    ${categoryNodeFragment}
  `
}
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
+++ b/libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
@@ -199,7 +199,7 @@ describe('Navigation | Driver | Magento | getCategoryNodeFragment', () => {
 	
 	//todo: remove this test when this bug is fixed: https://github.com/magento/magento2/issues/31086
 	//This test only exists to test the workaround.
-	it('should return the expected fragment', () => {
+	it('should not use nested fragments', () => {
 		const expectedFields = [
 			'id',
 			'level',
--- a/libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
+++ b/libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
@@ -171,7 +171,7 @@ describe('Navigation | Driver | Magento | getCategoryNodeFragment', () => {
     let response: Observable<ApolloQueryResult<CategoryNode>>;
 
     beforeEach(() => {
-      const fragment = getCategoryNodeFragment(3);
+			const fragment = getCategoryNodeFragment(3);
 
       query = gql`
         query TestQuery {
@@ -195,7 +195,23 @@ describe('Navigation | Driver | Magento | getCategoryNodeFragment', () => {
         data: depth3NavigationTree
       })
     });
-  });
+	});
+	
+	it('should return the expected fragment', () => {
+		const expectedFields = [
+			'id',
+			'level',
+			'name',
+			'include_in_menu',
+			'products',
+			'children_count',
+			'children'
+		];
+
+		(<any>getCategoryNodeFragment(3).definitions[0]).selectionSet.selections.forEach((selection, index) => {
+			expect(selection.name.value).toEqual(expectedFields[index]);
+		});
+	});
 
   afterEach(() => {
     controller.verify();
--- a/libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
+++ b/libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
@@ -197,6 +197,8 @@ describe('Navigation | Driver | Magento | getCategoryNodeFragment', () => {
     });
 	});
 	
+	//todo: remove this test when this bug is fixed: https://github.com/magento/magento2/issues/31086
+	//This test only exists to test the workaround.
 	it('should return the expected fragment', () => {
 		const expectedFields = [
 			'id',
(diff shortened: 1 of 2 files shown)


damienwebdev:
@lderrickable can you add a test that covers the bug here?

lderrickable:
> @lderrickable can you add a test that covers the bug here?

Added a test. It isn't pretty, but let me know if you have a better idea

libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
```diff
@@ -195,7 +195,23 @@ describe('Navigation | Driver | Magento | getCategoryNodeFragment', () => {
...
+			'name',
+			'include_in_menu',
+			'products',
+			'children_count',
+			'children'
+		];
+
+		(<any>getCategoryNodeFragment(3).definitions[0]).selectionSet.selections.forEach((selection, index) => {
+			expect(selection.name.value).toEqual(expectedFields[index]);
+		});
```
griest024, line 213: This spec doesn't test the behavior of the function, only its particular implementation. Even if the function was working as it should it wouldn't pass this test if it used fragments.

griest024 (changes requested):
After doing some more testing, I have discovered the source of the error. Spreading `...categoryNode` into the root of the `recursiveCategoryNode` causes Magento to throw. This might be something specific to the Magento GraphQL implementation; AFAIK this is allowed by the spec. I think this implementation is good but that test case should be removed and a healthy comment explaining the bug.

griest024:
Opened an issue with Magento: https://github.com/magento/magento2/issues/31086

libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
lderrickable: Do you mean "Even if the function had been working as it should..."? Because it does work now. While I agree that this test wouldn't have worked if the fragment spreading had been working as expected, it does work as the function stands now. It seems to me that this test does test the behavior of the function, because the way the function was implemented changed. Just because it wouldn't have passed before doesn't mean it doesn't accomplish its intended purpose. That being said, I also would agree that ideally this test would pass/fail for both implementations, and I'm really not sure how to write that test.

lderrickable:
> After doing some more testing, I have discovered the source of the error. Spreading `...categoryNode` into the root of the `recursiveCategoryNode` causes Magento to throw. This might be something specific to the Magento GraphQL implementation; AFAIK this is allowed by the spec. I think this implementation is good but that test case should be removed and a healthy comment explaining the bug.

I'd be ok with this. I'd also be ok with adding a comment to the test explaining that it is only necessary because of the workaround.

libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
griest024: Sorry I meant my previous implementation. Even if my implementation wasn't causing Magento to throw errors it wouldn't pass this test. So this test doesn't correctly describe the behavior of the function, only this particular implementation.

Hypothetically, let's say that Magento fixes that bug (if indeed it is a bug) and we can move back to my fragment implementation. We would have to remove your test case. Ideally, tests should only change when the feature requirements change. Since the "black box" behavior is not changing, swapping equivalent implementations should not fail tests. This indicates to me that this test case does not belong.

>I'm really not sure how to write that test

I don't think its possible. I think this is Magento-specific behavior and is a bug.

griest024:
@lderrickable 

> I'd be ok with this. I'd also be ok with adding a comment to the test explaining that it is only necessary because of the workaround.

After discussing this with Damien we decided that this test should stay. So just add a comment to the test explaining why its there, link to the Magento issue, and add a TODO saying to remove the test case once the bug is fixed.

Can you also add a TODO to the implementation saying to use fragments once the bug is fixed? The query complexity limit might start becoming an issue with larger query depths so we should return to the fragment implementation ASAP.

libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
```diff
@@ -195,7 +195,23 @@ describe('Navigation | Driver | Magento | getCategoryNodeFragment', () => {
         data: depth3NavigationTree
       })
     });
-  });
+	});
+	
+	it('should return the expected fragment', () => {
```
griest024, line 200: The spec message should be `should not use nested fragments` since that is specifically the behavior that we are targeting to workaround the bug.

griest024 (changes requested):

libs/navigation/src/drivers/magento/queries/fragments/category-node.spec.ts
lderrickable: changed

griest024 (approved):

PR #1176, opened 2020-12-03
feat(cart): improve CartResolution action hygiene, change resolved state to enum

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
- Changes the resolved state to use an enum
- Adds selectors for checking if resolution succeeded or failed and corresponding facade fields
- Changes the resolver effects to no longer use `substream` and to only dispatch resolver related actions
- Changes the cart store ID effect to listen for resolve success as well
- Changes the guards to not wait for cart resolution before emitting
- Changes the resolved cart guard to conditionally redirect based on if the URL has been specified

```
[x] Yes
[ ] No
```

The type of the `selectCartResolved` has changed to `DaffCartResolvedState` enum. The type of the `resolved$` facade field has changed in the same way.

## Other information

### DIFF
--- a/libs/cart/state/src/effects/cart-resolver.effects.spec.ts
+++ b/libs/cart/state/src/effects/cart-resolver.effects.spec.ts
@@ -136,26 +136,26 @@ describe('DaffCartResolverEffects', () => {
       });
 
       describe('and the cart is not found when attempting to load it', () => {
+        const newCartId = 'newCartId';
+
         beforeEach(() => {
           const response = cold('#', {}, new DaffCartNotFoundError('error'));
           driver.get.withArgs(cartId).and.returnValue(response);
+          driver.create.and.returnValue(of({id: newCartId}));
+        });
+
+        it('should create a new cart', () => {
+          const expected = cold('--b', {
+            b: jasmine.anything(),
+          });
+
+          expect(effects.onResolveCart$).toBeObservable(expected);
+          expect(driver.create).toHaveBeenCalled();
         });
 
         describe('and the driver calls succeed', () => {
           beforeEach(() => {
-            const newCartId = 'newCartId';
-            driver.create.and.returnValue(of({id: newCartId}));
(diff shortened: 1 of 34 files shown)


libs/cart/state/src/effects/cart-resolver.effects.spec.ts
```diff
@@ -62,278 +55,178 @@ describe('DaffCartResolverEffects', () => {
...
-          a: resolveCartAction,
-          b: createCartSuccessAction
+        describe('and the driver calls succeed', () => {
+          beforeEach(() => {
+            const newCartId = 'newCartId';
+            driver.create.and.returnValue(of({id: newCartId}));
+            driver.get.withArgs(newCartId).and.returnValue(of(stubCart));
+          });
+
+          it('should create a new cart', () => {
```
lderrickable, line 151: I got a little confused when reading through these last set of `and the cart is not found when attempting to load it` tests. I think it is because this test should be moved up one level, since it happens whether the driver calls are successful or not.

lderrickable:
Changes need to be made to the `cart-item-entities.reducer.ts` too, since the resolve cart flow doesn't rely on `DaffCartLoad` or `DaffCartCreate` anymore.

griest024:
> Changes need to be made to the `cart-item-entities.reducer.ts` too, since the resolve cart flow doesn't rely on `DaffCartLoad` or `DaffCartCreate` anymore.

@lderrickable good catch, changed

libs/cart/state/src/effects/cart-resolver.effects.spec.ts
griest024: moved up a level

damienwebdev (approved):
lgtm

lderrickable (approved):

PR #1184, opened 2020-12-08
fix(cart): some cart fields disappear during mutations that do not re…

lderrickable:
…trieve the entire cart

## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

A few magento cart mutations don't return the entire cart, but are transformed through the `DaffMagentoCartTransformer` (instead of individual transformers like the `DaffMagentoCartPaymentTransformer`). Because magento can return null values in place of objects or empty arrays, the `DaffMagentoCartTransformer` transformer can't just omit `DaffCart` fields if they are not present; instead, these `DaffCart` fields must be set to null. Therefore, when the magento apply coupon mutation only retrieves part of the cart and uses the `DaffMagentoCartTransformer`, the fields that were not retrieved are transformed into null fields instead of being omitted completely. Then, when the `DaffCart` gets to the reducer step, the cart with incorrect null fields are spread over the cart state.

## What is the new behavior?
Because magento returns null values in place of objects or empty arrays, we can't change the `DaffMagnetoCartTransformer` to omit fields that are null. This leaves us with two options: 1. add special transformers for these particular mutations or 2. retrieve the entire cart for every mutation of the cart. I went with number 2 since the mutations already retrieved a lot of the cart and it was a simpler change. 

```
[ ] Yes
[x] No
```


lderrickable:
I don't think there is a way for this to be tested with the `ApolloTestingModule`

libs/cart/driver/magento/src/queries/set-guest-email.ts
```diff
@@ -3,17 +3,20 @@ import gql from 'graphql-tag';
...
+
 export const setGuestEmail = (extraCartFragments: DocumentNode[] = []) => gql`
   mutation SetGuestEmail($cartId: String!, $email: String!) {
     setGuestEmailOnCart(input: {
       cart_id: $cartId,
       email: $email
     }) {
       cart {
-        email
+        ...cart
```
griest024, line 15: I don't think this change is needed. I don't even think this query is actually used anywhere anyway.

griest024 (changes requested):
I'm almost positive that @damienwebdev will prefer the distinct transformers option. Its definitely the more performant option...

lderrickable:
> I'm almost positive that @damienwebdev will prefer the distinct transformers option. Its definitely the more performant option...

Ok sure, I'll change it

lderrickable:
Ok good thing you said that, because that was actually how it already worked. I just needed to retrieve the shipping total through the apply coupon and remove all coupons queries. While that is extra data that isn't needed technically, the only other option is to update the reducer to have different behavior for the coupon actions, which I could do. I prefer less code in this case, since it would be adding code to `@daffodil/cart/state` instead of just the magento driver layer, but what do you think?

libs/cart/driver/magento/src/queries/set-guest-email.ts
lderrickable: I just removed this since it wasn't used

lderrickable:
Just realized I need to change the response objects too, so I'll do that later. Also, I think I might be able to test this. Will also do that later today

lderrickable:
@griest024 I updated the response models, but I couldn't get any useful test to work. I was going to add a test to the coupon driver service, but the problem is that the Apollo testing module doesn't care what query/mutation is given it. It just returns whatever object is flushed in the test.

griest024:
@lderrickable So you just updated the coupon queries to pull in the shipping address? What about the other fields like billing address or items?

lderrickable:
The shipping address is needed because it has the shipping totals. The billing address is not needed, because it has no value that can change. The cart items were retrieved previously and still are.

griest024 (approved):

PR #1185, opened 2020-12-08
feat(cart): resolve success on server side storage error

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When cart resolution throws storage errors in SSR, the resolve effect dispatches a resolve failure. This causes the resolve guard to emit `false`. This will cause  an app to never render any pages that are guarded by the cart resolve guard in SSR.

## What is the new behavior?
The resolve effect will return resolve success in response to the new server side storage error. This will cause the resolve cart guard to emit `true` in SSR contexts. Also adds a server side state module.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/state/src/effects/cart-resolver.effects.ts
+++ b/libs/cart/state/src/effects/cart-resolver.effects.ts
@@ -1,11 +1,11 @@
 import { Injectable, Inject } from '@angular/core';
 import { Actions, Effect, ofType } from '@ngrx/effects';
-import { Action } from '@ngrx/store';
+import { Action, select, Store } from '@ngrx/store';
 import { Observable, of } from 'rxjs';
-import { switchMap, catchError, map, mapTo } from 'rxjs/operators';
+import { switchMap, catchError, map } from 'rxjs/operators';
 
-import { DaffStorageServiceError } from '@daffodil/core';
-import { DaffCart, DaffCartStorageService } from '@daffodil/cart';
+import { DaffStorageServiceError, DaffServerSideStorageError } from '@daffodil/core';
+import { DaffCart, DaffCartOrderResult, DaffCartStorageService } from '@daffodil/cart';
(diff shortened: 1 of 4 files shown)


libs/cart/state/src/effects/cart-resolver.effects.ts
```diff
@@ -36,6 +43,11 @@ export class DaffCartResolverEffects<T extends DaffCart = DaffCart> {
       map(resp => new DaffResolveCartSuccess(resp)),
       catchError(error => {
         switch(true) {
+          case error instanceof DaffServerSideStorageError:
+            return this.store.pipe(
+              select(getDaffCartSelectors<T, V, U>().selectCartValue),
+              map(cart => new DaffResolveCartSuccess(cart))
+            );
```
lderrickable, line 50: So this switch case would only ever be triggered server side, right? 

How do we know that it should always resolve with success if no cart was retrieved? Wouldn't the cart in state always be null at this point, because it couldn't even get the id from local storage? Doesn't this mean that we should never use the resolve cart guard server side, because the server can never verify that there is a valid cart id in local storage? Should the ResolveCartGuard just always resolve immediately server-side?

damienwebdev:
I think this should be solved a bit differently, so I think the current version here is too far off from where we want to go. 

I'm going to open a PR to address this a little differently.

griest024:
@damienwebdev 

> I think the current version here is too far off from where we want to go.

Is this not what we had discussed or are you just reconsidering our idea?

damienwebdev:
@griest024 reconsidering. @lderrickable and I discussed offline that there's probably a more precise way to guard and render the SSR state resolution than "faking" a successful state.

PR #1196, opened 2020-12-23
fix(core): queued apollo mutation errors prevent subsequent requests

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

A queued mutation that threw an error would not cause the mutation queued to be processed and therefore subsequent requests would never be sent.

## What is the new behavior?
The mutation queued will be processed for all observable events (emit, error, and complete) and subsequent requests will be processed when all prior requests complete in any way.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/core/graphql/src/apollo/queued-apollo.service.spec.ts
+++ b/libs/core/graphql/src/apollo/queued-apollo.service.spec.ts
@@ -2,7 +2,7 @@ import { TestBed } from '@angular/core/testing';
 import { Apollo } from 'apollo-angular';
 import gql from 'graphql-tag';
 import { interval } from 'rxjs';
-import { delay, switchMap, switchMapTo, take } from 'rxjs/operators';
+import { delay, switchMap, take } from 'rxjs/operators';
 import { TestScheduler } from 'rxjs/testing';
 
 import { DaffQueuedApollo } from './queued-apollo.service'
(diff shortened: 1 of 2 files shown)


libs/core/graphql/src/apollo/queued-apollo.service.spec.ts
```diff
@@ -100,25 +100,50 @@ describe('Core | GraphQL | DaffQueuedApollo', () => {
...
+            })).and.returnValue(cold('--a', {a: data1}))
+
+            ob0 = service.mutate({
+              mutation: req0
+            });
+            ob1 = service.mutate({
+              mutation: req1
+            });
+
+            expectObservable(ob0).toBe('--#', {}, 'error');
```
griest024, line 143: @damienwebdev @lderrickable  Should the outside observable, the one returned by `QueuedApollo#mutate`, complete on error (`'--(#|)'` or is it fine like this?

damienwebdev: I suspect that all observables from Apollo complete upon an error, so I think it should be safe to complete.

griest024: I'm not sure that rxjs even supports completing after error like this but I called `complete` anyway.

damienwebdev: http://reactivex.io/documentation/contract.html

```
OnError
indicates that the Observable has terminated with a specified error condition and that it will be emitting no further items
```

damienwebdev: It should actually be sufficient as you previously had it.

griest024: I just reverted it to how I previously had it. The test cases are the same anyway.

lderrickable (approved):

PR #1197, opened 2020-12-24
test(cart,order): allow discount to be equal to price in item factories

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The factories allow for the discount to be the same as the price. The tests assert that the discount is strictly less than the price and will therefor occasionally fail when they are equal.

## What is the new behavior?
The tests allow for discounts less than or equal to the price.

```
[ ] Yes
[x] No
```

## Other information
I think allowing for discounts equal to the price is the behavior we want since items can be free by some discounts.


damienwebdev (approved):
This would only happen in the scenario when the item is free, correct?

griest024:
> This would only happen in the scenario when the item is free, correct?

Correct, the tests only previously failed when the item is free.

Issue #1202, opened 2020-12-29
[BUG] Cart item state is incorrect when more than one cart item mutation occurs concurrently


lderrickable:
# :bug: Bug report

## Current Behavior
When two cart item mutations occur concurrently, only the state of the lastly mutated cart item will be correct when all requests complete, because the `DaffCartItemUpdateSuccess` action only has the id of the last cart item. The previous `DaffCartItemUpdateSuccess` action gets canceled by the `switchMap` in the cart item effects.

## Expected Behavior
Both cart items should have the "updated" state when the requests finish, or each cart item should be "updated" as each request completes.

## Minimal reproduction of the problem with instructions
Update the quantity of cartItem1. Update the quantity of cartItem2 before the previous request completes. The resulting state will have the state of the cartItem2 correct, while the state of cartItem2 is incorrect.

## Suggested fix
Change the `switchMap` to a `concatMap`. This will cause the loading state of the cartItemReducer state to break, because that loading state is just a simple enum value and the first `DaffCartItemUpdateSuccess` will toggle the cartItemReducer loading state to "Complete" when it should still be "Mutating". I think this is an inherent flaw with the loading states of feature areas that process mutations, because while concurrent get requests can always be covered accurately by a `switchMap`, concurrent mutate requests need to be queued by a `concat/mergeMap`. So this suggested fix also requires a rework of the all cart reducer state that handles mutations; i.e. that the `DaffLoadingState` enum should be a queue of `DaffLoadingState`s. Then upon each `DaffCartItemUpdateSuccess` the queue is popped, and the selectors determine the loading state based on this queue.

PR #1203, opened 2021-01-05
fix(cart): fix bug where concurrent cart item updates would result in…

lderrickable:
… incorrect item states

BREAKING CHANGE: The selectItemMutating selector is now a derived selector of all cart item states and is called selectCartItemMutating

## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Cart item update success and cart item delete success actions are only dispatched when the latest request is returned. This makes concurrent cart item mutations result in incorrect cart item states.

Fixes: https://github.com/graycoreio/daffodil/issues/1202

## What is the new behavior?
Cart item update success and cart item delete success actions are dispatched when every relevant request is returned. 

```
[x] Yes
[ ] No
```

### DIFF
--- a/libs/cart/state/src/effects/cart-item.effects.ts
+++ b/libs/cart/state/src/effects/cart-item.effects.ts
@@ -1,5 +1,5 @@
 import { Injectable, Inject } from '@angular/core';
-import { switchMap, map, catchError, debounceTime, concatMap } from 'rxjs/operators';
+import { switchMap, map, catchError, debounceTime, concatMap, mergeMap } from 'rxjs/operators';
 import { of } from 'rxjs';
 import { Actions, Effect, ofType } from '@ngrx/effects';
 
@@ -81,7 +81,7 @@ export class DaffCartItemEffects<
   @Effect()
   update$ = this.actions$.pipe(
     ofType(DaffCartItemActionTypes.CartItemUpdateAction),
-    concatMap((action: DaffCartItemUpdate<T>) =>
+		mergeMap((action: DaffCartItemUpdate<T>) => 
 			this.driver.update(
 				this.storage.getCartId(),
 				action.itemId,
@@ -103,7 +103,7 @@ export class DaffCartItemEffects<
   @Effect()
   delete$ = this.actions$.pipe(
     ofType(DaffCartItemActionTypes.CartItemDeleteAction),
-    concatMap((action: DaffCartItemDelete<T>) =>
+    mergeMap((action: DaffCartItemDelete<T>) =>
       this.driver.delete(this.storage.getCartId(), action.itemId).pipe(
         map((resp: V) => new DaffCartItemDeleteSuccess(resp)),
         catchError(error => of(new DaffCartItemDeleteFailure(this.errorMatcher(error))))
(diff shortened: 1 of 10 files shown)


libs/cart/state/src/selectors/cart/cart.selector.ts
```diff
@@ -384,18 +377,18 @@ const createCartSelectors = <
...
       selectShippingInformationMutating,
       selectPaymentMutating,
       selectCouponMutating,
-			selectItemMutating,
-			selectItemAdding,
+			selectItemAdding
     ].map(selector =>
       selector.projector(loadingObject)
-    ).reduce((acc, mutating) => acc || mutating, false)
+    ).reduce((acc, mutating) => acc || mutating || cartItemMutating, false)
```
griest024, line 391: `cartItemMutating` shouldn't be inside the `reduce` callback. It can either be the initial value for `reduce` or outside the `reduce` call entirely.

libs/cart/state/src/effects/cart-item.effects.ts
```diff
@@ -103,7 +103,7 @@ export class DaffCartItemEffects<
   @Effect()
   delete$ = this.actions$.pipe(
     ofType(DaffCartItemActionTypes.CartItemDeleteAction),
-    switchMap((action: DaffCartItemDelete<T>) =>
+    concatMap((action: DaffCartItemDelete<T>) =>
```
griest024, line 106: Can you write a test that fails for the `switchMap` but succeeds for this implementation?

griest024 (changes requested):
You can remove the item entry from `DaffCartLoading`.

lderrickable:
> You can remove the item entry from `DaffCartLoading`.

I'm pretty sure it needs to stay because it still tracks the resolve/loading status for all cart items. The only part that is now tracked by the cart item states is the mutation state

lderrickable:
> > I'm pretty sure it needs to stay because it still tracks the resolve/loading status for all cart items. The only part that is now tracked by the cart item states is the mutation state
> 
> good point. you can probably modify `DaffCartItemLoadingState` to reflect that though, right?

yea true. I'll make that change

libs/cart/state/src/selectors/cart/cart.selector.ts
lderrickable: Moved to outside the `reduce`

libs/cart/state/src/effects/cart-item.effects.ts
lderrickable: added

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
```diff
@@ -126,6 +127,12 @@ const createCartItemEntitiesSelectors = <
 		}
 	);
 
+	const selectCartItemMutating = createSelector(
+		selectAllCartItems,
+		(cartItems: U[]) => cartItems && cartItems.reduce((acc, item) =>
```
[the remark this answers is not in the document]
lderrickable: added

libs/cart/state/src/effects/cart-item.effects.ts
lderrickable: also, I made it a mergeMap instead. concatMap holds the request until the previous one is successful, which I don't think we want, especially because the queued apollo service already queues the requests

griest024 (approved):

PR #1227, opened 2021-01-12
feat(newsletter): sharded @daffodil/newsletter library

Nolan-Arnold:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Touches #1340 

Fixes: N/A

```
[x ] Yes
[ ] No
```

## Other information


libs/newsletter/state/testing/ng-package.prod.json
```diff
@@ -0,0 +1,7 @@
+{
```
lderrickable, line 1: This used to be a `package.json` instead of `ng-package.prod.json` what's the reason for the change?

lderrickable (changes requested):

libs/newsletter/state/testing/ng-package.prod.json
Nolan-Arnold: I was just keeping consistent with the changes made in `@daffodil/contact`. If package.json is the standard in the rest of the packages ill change it.

lderrickable: I'm not sure if it matters either way. This really isn't my forte. @griest024 @damienwebdev either of you know which is correct, or if it matters?

griest024: AFAIK the subpackages need only a `ng-package.json` and a `package.json`. The `package.json` might be unnecessary since the folder structure takes care of the pathing; @damienwebdev will know for sure.

I do see that contact does not use a `package.json` for the subpackages but the other libs do.

damienwebdev: Under the hood this uses https://github.com/ng-packagr/ng-packagr. Their docs may have changed, but it warrants consideration to re-read their docs on secondary entrypoints.

lderrickable: From what I could gather, it seems like the common practice is for your ng-package file to be named `package.json`. So that means all of the other packages aren't correct either. In most other packages, we have a file named `package.json`, but I don't think it contains anything of value. So that can be deleted, and then all of the ng-package files should be `package.json`. 

So I think for this PR, the `ng-package.json` and `ng-package.prod.json` files should be renamed to `package.json` and `package.prod.json`. Then we should do another PR to convert all the other ng-package.json files in the other packages.

lderrickable: I looked into this a little more, and while the docs say to put this in a package.json, if you try to put it in the root package.json (@daffodil/newsletter), you get errors. Also, in some other stackoverflow questions it seems normal to name it ng-package.json. So I think we just leave it as is in this PR, and we could remove the sub-package `package.json`s in the other packages because I'm pretty sure they don't do anything

lderrickable:
build is failing

lderrickable:
The was completed in another PR: https://github.com/graycoreio/daffodil/pull/1385

PR #1264, opened 2021-02-03
feat(core, driver): add infrastructure for making cacheable apollo queries to different platforms

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
This token is for providing to apollo which requests should be sent as GETs. More info on usage in the doc comment.


libs/core/graphql/src/apollo/get-registration/get-registration.ts
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * A multi-provider injection token of the operation names of apollo queries that should 
+ * be GETs instead of POSTs. This is needed because apollo sends POST requests by default and daffodil
+ * defines all of the operation names. Each daffodil package will automatically register the requests
+ * that should be GETs when its driver package is loaded into the app. Simply filter on these provided
+ * request names for each outgoing request.
+ */
+export const DAFF_APOLLO_GET_REQUESTS = new InjectionToken<string>('DAFF_APOLLO_GET_REQUESTS', {
+	factory: () => ''
```
damienwebdev, line 11: Should this return an array?

libs/core/graphql/src/apollo/get-registration/get-registration.ts
```diff
@@ -0,0 +1,12 @@
+import { InjectionToken } from '@angular/core';
+
+/**
+ * A multi-provider injection token of the operation names of apollo queries that should 
+ * be GETs instead of POSTs. This is needed because apollo sends POST requests by default and daffodil
+ * defines all of the operation names. Each daffodil package will automatically register the requests
+ * that should be GETs when its driver package is loaded into the app. Simply filter on these provided
+ * request names for each outgoing request.
+ */
+export const DAFF_APOLLO_GET_REQUESTS = new InjectionToken<string>('DAFF_APOLLO_GET_REQUESTS', {
```
damienwebdev, line 10: Should this be scoped per driver?

damienwebdev (changes requested):

libs/core/graphql/src/apollo/get-registration/get-registration.ts
lderrickable: I'm not really sure what it should technically return. I thought that because it is a multiprovider, it will automatically be an array of whatever it is. It does work in the demo app as expected and comes through as an array of strings.

libs/core/graphql/src/apollo/get-registration/get-registration.ts
lderrickable: I don't think so. Because apollo will expect the same format no matter the driver, and I was planning on this being provided in each platform specific driver module.

libs/core/graphql/src/apollo/get-registration/get-registration.ts
lderrickable: upon further consideration, I decided it is better as an array. Changed

lderrickable:
@griest024 I put the apollo cache tokens and service in `@daffodil/core/graphql`, because that is where the other apollo stuff is. I think if we want it in driver, we should just move the entire apollo directory to the driver module. And maybe we should actually do that, because I think apollo things will only be used by devs who are also using the drivers. Probably doesn't make sense for any of it to be in core.

libs/driver/magento/src/graphql/apollo-cached-operation-converter.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+
+describe('Driver | Magento | GraphQl | DaffMagentoApolloCachedOperationConverter', () => {
+	
+	it('should set the operation to a GET method', () => {
+		const mockOperation = {
+			query: null,
+			operationName: null,
+			extensions: null,
+			variables: null,
+			setContext: jasmine.createSpy(),
```
griest024, line 11: How difficult would it be to create a real operation object and assert the HTTP method is `'GET'`?

griest024 (changes requested):

griest024:
> @griest024 I put the apollo cache tokens and service in `@daffodil/core/graphql`, because that is where the other apollo stuff is. I think if we want it in driver, we should just move the entire apollo directory to the driver module. And maybe we should actually do that, because I think apollo things will only be used by devs who are also using the drivers. Probably doesn't make sense for any of it to be in core.

Yeah I could see it all being moved to driver. We can worry about that later though.

libs/core/graphql/src/apollo/cacheable-operations/cacheable-operations.ts
```diff
@@ -0,0 +1,50 @@
...
+ * This token should be used by daffodil driver platform modules to register operation names that are cacheable, and is
+ * intended for internal daffodil use only.
+ */
+export const DAFF_APOLLO_CACHEABLE_OPERATIONS = new InjectionToken<string[]>('DAFF_APOLLO_CACHEABLE_OPERATIONS');
+/**
+ * An injection token for a converter function that makes apollo operations cacheable. The kind of conversion necessary is 
+ * determined by the platform used. These converter functions are defined by daffodil in the respective `@daffodil/driver/{platform}`
+ * packages, but need to be provided by the app.
+ */
+export const DAFF_APOLLO_CACHEABLE_OPERATIONS_CONVERTER = new InjectionToken<DaffApolloCachedOperationConverter>('DAFF_APOLLO_CACHEABLE_OPERATIONS_CONVERTER');
```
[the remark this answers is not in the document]
lderrickable: I had to separate the tokens from the service, because it was causing the build to fail when I added the default for the token. Done

libs/driver/magento/src/graphql/apollo-cached-operation-converter.spec.ts
lderrickable: It's a little janky, but better than the spy. Done

libs/core/graphql/src/apollo/cacheable-operations/cacheable-operations-converter.ts
```diff
@@ -0,0 +1,14 @@
...
+ * The type of function to be provided via the DAFF_APOLLO_CACHEABLE_OPERATIONS_CONVERTER.
+ */
+export type DaffApolloCachedOperationConverter = (o: Operation) => Operation;
+
+/**
+ * An injection token for a converter function that makes apollo operations cacheable. The kind of conversion necessary is 
+ * determined by the platform used. These converter functions are defined by daffodil in the respective `@daffodil/driver/{platform}`
+ * packages, but need to be provided by the app.
+ */
+export const DAFF_APOLLO_CACHEABLE_OPERATIONS_CONVERTER = new InjectionToken<DaffApolloCachedOperationConverter>('DAFF_APOLLO_CACHEABLE_OPERATIONS_CONVERTER');
```
griest024, line 14: I think this one should have the identity function as a default.

libs/driver/magento/src/graphql/apollo-cached-operation-converter.spec.ts
griest024: did you forget to push it? I don't see it

libs/core/graphql/src/apollo/cacheable-operations/cacheable-operations-converter.ts
lderrickable: makes sense. Added

libs/core/graphql/src/apollo/get-registration/get-registration.ts
griest024: IME, you should type multi tokens as strings

libs/driver/magento/src/graphql/apollo-cached-operation-converter.spec.ts
lderrickable: I don't think so

griest024: LUL

I like this test much better though even though its more integration than unit.

griest024 (dismissed):

libs/driver/magento/src/graphql/apollo-cached-operation-converter.ts
```diff
@@ -0,0 +1,12 @@
+import { Operation } from '@apollo/client/core';
+
+/**
+ * A function to be provided via the DAFF_APOLLO_CACHEABLE_OPERATIONS_CONVERTER injection token when Magento
+ * is the backend platform. This function will convert cacheable operations into an equivalent operation
+ * that Magento will recognize as cacheable.
+ */
+export const DaffMagentoApolloCachedOperationConverter = (o: Operation): Operation => {
```
damienwebdev, line 8: Why wouldn't we just expose the ApolloLink? It seems more suited, and intended, for the purpose than a direct mutation of the underlying operation.

damienwebdev (changes requested):

libs/driver/magento/src/graphql/apollo-cached-operation-converter.ts
lderrickable: We could do that, but we would need two ApolloLinks per platform instead of one per platform and one in core. The reason we would need two per platform is to cover the edgecase of using two platforms for one app. If the app-dev uses two platforms, then the ApolloLinks needs to filter on the platform name as well as `DAFF_APOLLO_CACHEABLE_OPERATIONS`. We could try to compress these "two ApolloLinks per platform" into one, but then we would need to always filter on the platform name, which we don't want when only one platform is used.

So I went with one ApolloLink in each platform (not done in this PR) to cover the edge case, plus one ApolloLink in core instead of two ApolloLinks in each `@daffodil/driver/${platform}`. I could see it either way though, honestly. If you prefer it the other way, I can change it. The naming might get a little tricky, but it can be cleared up with a good doc

libs/driver/magento/src/graphql/cacheable-operations/apollo-cacheable-operations-link-generator.ts
```diff
@@ -0,0 +1,26 @@
...
+	providedIn: 'root'
+})
+export class DaffMagentoApolloCacheableOperationsLinkGenerator implements DaffApolloCacheableOperationLinkGenerator {
+	constructor(
+		@Inject(DAFF_MAGENTO_CACHEABLE_OPERATIONS) private apolloGetRequests: string[],
+	) {}
+
+	getLink(): ApolloLink {
+		return new ApolloLink((operation, forward) => {
+			if(this.apolloGetRequests.includes(operation.operationName)) {
```
damienwebdev, line 20: `includes` isn't supported by IE, can you check whether or not typescript will compile this appropriately to `indexOf`?

damienwebdev (changes requested):
minor comment, but LGTM

libs/driver/magento/src/graphql/cacheable-operations/apollo-cacheable-operations-link-generator.ts
lderrickable: I don't think it does. changed to `indexOf` usage

damienwebdev (approved):

PR #1267, opened 2021-02-05
feat(cart): add cart item price selectors

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Adds selectors for cart item total discount, price, quantity, and row total.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/src/models/cart-item.ts
+++ b/libs/cart/src/models/cart-item.ts
@@ -20,7 +20,13 @@ export interface DaffCartItem {
   sku: string;
   name: string;
   qty: number;
+  /**
+   * The price per item, not including cart discounts.
+   */
   price: number;
+  /**
+   * The total for the entire row, not including cart discounts.
+   */
   row_total: number;
 	in_stock: boolean;
   /**
--- a/libs/cart/state/src/facades/cart/cart-facade.interface.ts
+++ b/libs/cart/state/src/facades/cart/cart-facade.interface.ts
@@ -3,7 +3,7 @@ import { Action } from '@ngrx/store';
 import { Dictionary } from '@ngrx/entity';
 
 import { DaffStateError, DaffStoreFacade } from '@daffodil/core/state';
-import { DaffCart, DaffCartOrderResult, DaffCartTotal, DaffConfigurableCartItemAttribute, DaffCompositeCartItemOption } from '@daffodil/cart';
+import { DaffCart, DaffCartOrderResult, DaffCartTotal, DaffConfigurableCartItemAttribute, DaffCompositeCartItemOption, DaffCartItemDiscount } from '@daffodil/cart';
 
 import { DaffCartErrors } from '../../reducers/errors/cart-errors.type';
 import { DaffCartOperationType } from '../../reducers/cart-operation-type.enum';
@@ -232,7 +232,31 @@ export interface DaffCartFacadeInterface<
 	getCompositeCartItemOptions(itemId: U['item_id']): Observable<DaffCompositeCartItemOption[]>;
 	isCartItemOutOfStock(itemId: U['item_id']): Observable<boolean>;
 	/**
-	 * The state of a cart item.
+   * The state of a cart item.
 	 */
   getCartItemState(itemId: U['item_id']): Observable<DaffCartItemStateEnum>;
+  /**
+   * Selects the specified item's price.
+   * Zero by default.
+   */
+	getCartItemPrice(itemId: U['item_id']): Observable<number>;
+  /**
+   * Selects the specified item's quantity.
+   * Zero by default.
+   */
+	getCartItemQuantity(itemId: U['item_id']): Observable<number>;
+  /**
+   * Selects the specified item's row total.
+   * Zero by default.
+   */
+	getCartItemRowTotal(itemId: U['item_id']): Observable<number>;
+  /**
+   * Selects the specified item's array of discounts.
+   */
+	getCartItemDiscounts(itemId: U['item_id']): Observable<DaffCartItemDiscount[]>;
+  /**
+   * Selects the specified item's sum of all discounts for the entire row.
+   * Zero by default.
+   */
+	getCartItemTotalDiscount(itemId: U['item_id']): Observable<number>;
 }
--- a/libs/cart/state/src/facades/cart/cart.facade.spec.ts
+++ b/libs/cart/state/src/facades/cart/cart.facade.spec.ts
@@ -1606,4 +1606,64 @@ describe('DaffCartFacade', () => {
       expect(facade.canPlaceOrder$).toBeObservable(expected);
     });
   });
+
+  describe('getCartItemPrice', () => {
+
+    it('should be the cart item\'s price', () => {
+      const cart = cartFactory.create({
+        items: statefulCartItemFactory.createMany(2)
+      });
+      const expected = cold('a', { a: cart.items[0].price });
+      facade.dispatch(new DaffCartLoadSuccess(cart));
+      expect(facade.getCartItemPrice(cart.items[0].item_id)).toBeObservable(expected);
+    });
+  });
+
+  describe('getCartItemQuantity', () => {
+
+    it('should be the cart item\'s quantity', () => {
+      const cart = cartFactory.create({
+        items: statefulCartItemFactory.createMany(2)
+      });
+      const expected = cold('a', { a: cart.items[0].qty });
+      facade.dispatch(new DaffCartLoadSuccess(cart));
+      expect(facade.getCartItemQuantity(cart.items[0].item_id)).toBeObservable(expected);
+    });
+  });
+
+  describe('getCartItemRowTotal', () => {
+
+    it('should be the cart item\'s row total', () => {
+      const cart = cartFactory.create({
+        items: statefulCartItemFactory.createMany(2)
+      });
+      const expected = cold('a', { a: cart.items[0].row_total });
+      facade.dispatch(new DaffCartLoadSuccess(cart));
+      expect(facade.getCartItemRowTotal(cart.items[0].item_id)).toBeObservable(expected);
+    });
+  });
+
+  describe('getCartItemDiscounts', () => {
+
+    it('should be the cart item\'s array of discounts', () => {
+      const cart = cartFactory.create({
+        items: statefulCartItemFactory.createMany(2)
+      });
+      const expected = cold('a', { a: cart.items[0].discounts });
+      facade.dispatch(new DaffCartLoadSuccess(cart));
+      expect(facade.getCartItemDiscounts(cart.items[0].item_id)).toBeObservable(expected);
+    });
+  });
+
+  describe('getCartItemTotalDiscount', () => {
+
+    it('should be the cart item\'s sum of all discounts', () => {
+      const cart = cartFactory.create({
+        items: statefulCartItemFactory.createMany(2)
+      });
+      const expected = cold('a', { a: cart.items[0].discounts.reduce((acc, {amount}) => acc + amount, 0) });
+      facade.dispatch(new DaffCartLoadSuccess(cart));
+      expect(facade.getCartItemTotalDiscount(cart.items[0].item_id)).toBeObservable(expected);
+    });
+  });
 });
--- a/libs/cart/state/src/facades/cart/cart.facade.ts
+++ b/libs/cart/state/src/facades/cart/cart.facade.ts
@@ -5,7 +5,7 @@ import { Action, Store, select } from '@ngrx/store';
 import { Dictionary } from '@ngrx/entity';
 
 import { DaffStateError } from '@daffodil/core/state';
-import { DaffCart, DaffCartOrderResult, DaffCartTotal, DaffCartPaymentMethodIdMap, DaffConfigurableCartItemAttribute, DaffCompositeCartItemOption } from '@daffodil/cart';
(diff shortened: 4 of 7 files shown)


libs/cart/state/src/facades/cart/cart-facade.interface.ts
```diff
@@ -228,12 +228,51 @@ export interface DaffCartFacadeInterface<
...
 	getConfiguredCartItemAttributes(itemId: U['item_id']): Observable<DaffConfigurableCartItemAttribute[]>;
 	getCompositeCartItemOptions(itemId: U['item_id']): Observable<DaffCompositeCartItemOption[]>;
 	isCartItemOutOfStock(itemId: U['item_id']): Observable<boolean>;
 	/**
-	 * The state of a cart item.
+   * The state of a cart item.
 	 */
   getCartItemState(itemId: U['item_id']): Observable<DaffCartItemStateEnum>;
+  /**
+   * Selects the specified item's total discount for the entire line.
```
lderrickable, line 239: What does "for the entire line" mean?

lderrickable, line 239: What does "for the entire line" mean?

libs/cart/state/src/facades/cart/cart.facade.spec.ts
```diff
@@ -1618,4 +1607,100 @@ describe('DaffCartFacade', () => {
...
+    });
+  });
+
+  describe('getCartItemDiscount', () => {
+
+    it('should be the cart item\'s discounted price', () => {
+      const cart = cartFactory.create({
+        items: statefulCartItemFactory.createMany(2)
+      });
+      const expected = cold('a', { a: daffDivide(cart.items[0].total_discount, cart.items[0].qty) });
```
lderrickable, line 1629: I think we should have an additional field on the cart item model for `discount` and do this computation once in the magento transformer. Hopefully other platforms will just provide this value too

libs/cart/state/src/facades/cart/cart.facade.spec.ts
```diff
@@ -1618,4 +1607,100 @@ describe('DaffCartFacade', () => {
...
+    });
+  });
+
+  describe('getCartItemDiscountedPrice', () => {
+
+    it('should be the cart item\'s discounted price', () => {
+      const cart = cartFactory.create({
+        items: statefulCartItemFactory.createMany(2)
+      });
+      const expected = cold('a', { a: daffSubtract(cart.items[0].price, daffDivide(cart.items[0].total_discount, cart.items[0].qty)) });
```
lderrickable, line 1665: I think this should be an additional field on cart-item, so the computation can be done only once in the magento transformer

lderrickable (changes requested):

libs/cart/state/src/facades/cart/cart-facade.interface.ts
griest024: For the line item. I should have said 'row' since that's the terminology that we use. Changed

libs/cart/state/src/facades/cart/cart.facade.spec.ts
griest024: added

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
```diff
@@ -21,6 +22,46 @@ export interface DaffCartItemEntitiesMemoizedSelectors<T extends DaffStatefulCar
 	selectIsCartItemOutOfStock: MemoizedSelectorWithProps<object, object, boolean>;
 	selectCartItemMutating: MemoizedSelector<object, boolean>;
 	selectCartItemState: MemoizedSelectorWithProps<object, object, DaffCartItemStateEnum>;
+  /**
+   * Selects the specified item's total discount for the entire row.
+   * Zero by default.
+   */
+	selectCartItemTotalDiscount: MemoizedSelectorWithProps<object, {id: DaffCartItem['item_id']}, number>;
```
damienwebdev, line 29: What is the difference between `selectCartItemTotalDiscount` and `selectCartItemDiscountedRowTotal`

damienwebdev (changes requested):

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
griest024: `selectCartItemDiscountedRowTotal` is a price, the row total minus the total discount

griest024: maybe `selectCartItemRowTotalWithDiscount` is a clearer name?

lderrickable: Another option is `selectCartItemTotalDiscountAmount` and `selectCartItemTotalDiscountedPrice`.

lderrickable: oh, or just `selectCartItemTotalDiscount` and `selectCartItemTotalDiscountedPrice`

libs/cart/state/src/facades/cart/cart.facade.ts
```diff
@@ -317,18 +339,46 @@ export class DaffCartFacade<
...
+
+	getCartItemQuantity(itemId: U['item_id']): Observable<number> {
+    return this.store.pipe(select(this._selectCartItemQuantity, { id: itemId }))
+  }
+
+	getCartItemRowTotal(itemId: U['item_id']): Observable<number> {
+    return this.store.pipe(select(this._selectCartItemRowTotal, { id: itemId }))
+  }
+
+  getCartItemDiscountedTotal(itemId: U['item_id']): Observable<number> {
```
lderrickable, line 378: I think the facade name should match the selector name: `getCartItemDiscountedRowTotal`

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.spec.ts
```diff
@@ -214,4 +223,148 @@ describe('selectCartItemEntitiesState', () => {
...
+			const selector = store.pipe(select(selectCartItemTotalDiscount, { id: mockCartItems[0].item_id + 'notId' }));
+			const expected = cold('a', { a: 0 });
+
+			expect(selector).toBeObservable(expected);
+    });
+  });
+
+  describe('selectCartItemDiscount', () => {
+
+		it('should return the total discount divided by the quantity of the cart item', () => {
```
lderrickable, line 247: This it clause can be updated

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.spec.ts
```diff
@@ -214,4 +223,148 @@ describe('selectCartItemEntitiesState', () => {
...
+			const selector = store.pipe(select(selectCartItemPrice, { id: mockCartItems[0].item_id + 'notId' }));
+			const expected = cold('a', { a: 0 });
+
+			expect(selector).toBeObservable(expected);
+    });
+  });
+
+  describe('selectCartItemDiscountedPrice', () => {
+
+		it('should return the price of the cart item', () => {
```
lderrickable, line 301: "it" clause is incorrect

lderrickable: a lot of these aren't right actually

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
```diff
@@ -142,6 +183,80 @@ const createCartItemEntitiesSelectors = <
 		}
 	)
 
+  const selectCartItemQuantity = createSelector(
+    selectCartItemEntities,
+    (cartItems, props) => {
+      const cartItem: DaffCartItem = selectCartItem.projector(cartItems, { id: props.id });
```
lderrickable, line 189: I think you can do: `selectCartItem.projector(cartItems, { id: props.id })?.qty || 0` instead. So you should be able to remove a lot of the new `const`s in these selectors

griest024: changed

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.spec.ts
griest024: fixed

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.spec.ts
griest024: fixed

libs/cart/state/src/facades/cart/cart.facade.ts
griest024: changed

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
```diff
@@ -134,13 +175,59 @@ const createCartItemEntitiesSelectors = <
...
+  const selectCartItemDiscount = createSelector(
+		selectCartItemEntities,
+		(cartItems, props) => selectCartItem.projector(cartItems, { id: props.id })?.discount || 0
+	);
+
+  const selectCartItemHasDiscount = createSelector(
 		selectCartItemEntities,
 		(cartItems, props) => {
-			const cartItem = selectCartItem.projector(cartItems, { id: props.id });
+			const discount = selectCartItemDiscount.projector(cartItems, { id: props.id });
```
lderrickable, line 200: don't need the const here

griest024: changed

lderrickable (dismissed):

libs/cart/state/src/selectors/cart-item-entities/cart-item-entities.selectors.ts
griest024: bump

libs/cart/driver/magento/src/transforms/outputs/cart-item/simple-cart-item-transformer.ts
```diff
@@ -18,6 +19,7 @@ export function transformMagentoSimpleCartItem(cartItem: MagentoCartItem): DaffC
 		name: cartItem.product.name,
 		qty: cartItem.quantity,
 		price: cartItem.prices.price.value,
+    discount: daffDivide(cartItem.prices.total_item_discount.value, cartItem.quantity),
```
damienwebdev, line 22: I'm not sure that the math here is sound. What happens when there's two qty and an odd decimal. It appears we'd be off by a penny.

damienwebdev (changes requested):

libs/cart/driver/magento/src/transforms/outputs/cart-item/simple-cart-item-transformer.ts
lderrickable: I don't think that can happen, unless there's a problem with the backend. Presumably, the total discount on the backend is derived from the (discount of one item) * quantity. So the total discount will always be divisible by the quantity, unless the backend is giving us the wrong number of significant figures.

If the backend instead multiplies the price by the quantity, then multiplies by some percentage to get the total discount, then that seems like a magento/shopify bug. Though I doubt this is how it works, because it doesn't really make sense.

libs/cart/src/models/cart-item.ts
```diff
@@ -15,8 +15,21 @@ export interface DaffCartItem {
...
+   * The price per item, not including cart discounts.
+   */
   price: number;
+  /**
+   * The total for the entire row, not including cart discounts.
+   */
   row_total: number;
+  /**
+   * The discount per item.
+   */
```
damienwebdev, line 28: After a lengthy discussion, this is definitely not possible in the manner examined in this PR. There are some nasty edge-cases with rounding and the actually expected behavior for rounding isn't consistent for every platform,  making this field either nullable or not per platform which is not a good enough interface for us to describe in Daffodil.

For example, consider the scenario:

```txt
round(A * B * C) ?= round(A*B) * C
```


One can formally prove that these are not necessarily the same, an exercise left for the reader.

Magento falls into the `round(A * B * C)` which means that if we attempt to compute a discount per item, we'll be left with half-pennies in various scenarios which is both numerically unsafe and potentially confusing.

damienwebdev (changes requested):

libs/cart/src/models/cart-item.ts
griest024: It seems like the only perfectly accurate option is to expose every single price-related field on the backend and directly map to Daffodil fields.

lderrickable: I think we can just never do division with discount fields

damienwebdev: Generally speaking, I'd argue doing any division on the client is going to cause us headaches unless we're extremely careful.

PR #1268, opened 2021-02-07
perf(demo): don't negate async pipes

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe: Perf
```

Some async pipes are negated. Since async pipes initially emit `null`, this will cause the negation to always be true even when it really shouldn't be. When the component is initialized and the pipes emit their correct values, many of these negations will switch from being `true` to `false`. This can cause many elements to be added or removed, thrashing the layout and causing perf issues.

See http://codelyzer.com/rules/template-no-negated-async/ for more info.

## What is the new behavior?
Async pipes are now explicitly compared to `false` when the negation is desired.

In the case of the newsletter component, a new `hasError$` field had to be added to ensure that the emitted value was a boolean.

```
[ ] Yes
[x] No
```

## Other information
This is preparation for the eslint migration, where this style will be enforced by a linter rule (its a recommended angular eslint rule).


damienwebdev (approved):
lgtm

PR #1270, opened 2021-02-09
feat(all): add cacheable queries to the magento cacheable queries list

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Cacheable queries are now cacheable by Magento through the apollo link generators.

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/demo/src/app/drivers/magento/magento.module.ts
+++ b/apps/demo/src/app/drivers/magento/magento.module.ts
@@ -14,6 +14,7 @@ import { DaffAuthMagentoDriverModule } from '@daffodil/auth';
 import { environment } from '../../../environments/environment';
 import { MagentoEnvironmentDriverConfiguration } from '../../../environments/environment.interface';
 import { possibleTypes } from './fragmentTypes.json';
+import { DaffMagentoApolloCacheableOperationsLinkGenerator } from '@daffodil/driver/magento';
 
 const cache = new InMemoryCache({ possibleTypes });
 
@@ -26,11 +27,16 @@ const cache = new InMemoryCache({ possibleTypes });
     DaffCheckoutInMemoryDriverModule.forRoot(),
     DaffNavigationMagentoDriverModule.forRoot(),
     DaffNewsletterInMemoryDriverModule.forRoot()
-  ]
+	]
 })
 export class DemoMagentoDriverModule {
   // Magento
-  constructor(apollo: Apollo, httpLink: HttpLink) {
+  constructor(
+		apollo: Apollo, 
+		httpLink: HttpLink,
+		private apolloCacheableOperationsLinkGenerator: DaffMagentoApolloCacheableOperationsLinkGenerator
+	) {
+
     apollo.create({
       link: ApolloLink.from([
         onError(({graphQLErrors, networkError}) => {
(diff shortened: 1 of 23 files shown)


apps/demo/src/app/drivers/magento/magento.module.ts
```diff
@@ -39,12 +45,13 @@ export class DemoMagentoDriverModule {
               console.log(
                 `[GraphQL error]: Message: ${message}, Location: ${locations}, Path: ${path}`,
               ),
-            );
+						);
           if (networkError) console.log(`[Network error]: ${networkError}`);
-        }),
+				}),
+				this.apolloCacheableOperationsLinkGenerator.getLink(),
```
damienwebdev, line 51: variable name should probably be -- magtentoLinkGen.getLink()

damienwebdev (changes requested):
Take a gander at https://devdocs.magento.com/guides/v2.4/graphql/caching.html -- I don't think countries are cachable.

apps/demo/src/app/drivers/magento/magento.module.ts
lderrickable: changed

lderrickable:
> Take a gander at https://devdocs.magento.com/guides/v2.4/graphql/caching.html -- I don't think countries are cachable.

fixed. Also, I removed the provider functions. It fails the build because of the "no function calls in decorators" aot rule.

damienwebdev:
@lderrickable the provider functions can be kept, but they need to be written like:

```
export function myFunction(): {
 return "test";
}

not 

export const myFunction = () => {
return "test";
};
```

lderrickable:
@damienwebdev I don't think that works. Seems like the point of restricting function calls in decorators is to ensure that you can't pass in data at runtime. I think the only reason you can do the factory function thing is because it forces you not to be able to pass anything in. Their example is:

```
export function serverFactory() {
  return new Server();
}

@Component({
  ...
  providers: [{provide: server, useFactory: serverFactory}]
})
```

I think the only reason you're allowed to do this is because you can't pass anything into the function, so even though it is a function, it can be completely understood before runtime. I think the only way we could make this work is something like:

```
export function navigationCacheableQueriesProvider() {
  return ({
     provide: MAGENTO_OPERATIONS_TOKEN,
     useValue: NAVIGATION_QUERY_NAME,
     multi: true
  })
}

@Component({
  ...
  providers: [{provide: server, useFactory: navigationCacheableQueriesProvider}]
})
```

I don't even know if that would work, but it certainly defeats the purpose of having the functions anyway. Maybe I'm wrong about this, but that's what I gathered from the angular docs. I tried using the new function you wrote too. It doesn't work

lderrickable:
@damienwebdev Added the provider functions for angular 9+ apps

libs/driver/magento/src/graphql/cacheable-operations/cacheable-operations-token.ts
```diff
@@ -3,6 +3,29 @@ import { InjectionToken } from '@angular/core';
...
+		provide: DAFF_MAGENTO_CACHEABLE_OPERATIONS,
+		useValue: operationName,
+		multi: true
+	}
+}
+
+/**
+ * Adds many operation names to the list of cacheable Magento operations. Use only with Angular 9+.
+ */
+export function provideManyDaffMagentoCacheableOperations(operationNames: string[]) {
```
griest024, line 25: I think its a little cleaner to have this be a rest param.

griest024 (changes requested):

libs/driver/magento/src/graphql/cacheable-operations/cacheable-operations-token.ts
lderrickable: I'm not sure what you mean

griest024: Like `provideManyDaffMagentoCacheableOperations(...operationNames: string[])` so you can just pass args into the function instead of an array.

lderrickable: oh, I didn't know that's what that was called. Changed

griest024 (approved):

damienwebdev (approved):

PR #1302, opened 2021-02-23
feat(core,cart): extend and clarify DaffStates and DaffStateable

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, DaffState has details shared across several packages.

Fixes: N/A

## What is the new behavior?
This moves most of that burden to core for organizational purposes. I've also outlined a new interface called DaffStateable which would be implemented to enforce consistency across the codebase.

```
[x] Yes
[ ] No
```

Users who depend on the old `DaffLoadingState` enum should use the new `DaffStates` enum.

## Other information

### DIFF
--- a/libs/cart/state/src/reducers/cart-billing-address/cart-billing-address.reducer.ts
+++ b/libs/cart/state/src/reducers/cart-billing-address/cart-billing-address.reducer.ts
@@ -1,5 +1,5 @@
 import { DaffCart } from '@daffodil/cart';
-import { DaffLoadingState } from '@daffodil/core/state';
+import { DaffState } from '@daffodil/core/state';
 
 import {
   DaffCartBillingAddressActionTypes,
@@ -27,14 +27,14 @@ export function cartBillingAddressReducer<T extends DaffCart>(
     case DaffCartBillingAddressActionTypes.CartBillingAddressLoadAction:
       return {
         ...state,
-        ...setLoading(state.loading, DaffLoadingState.Resolving),
+        ...setLoading(state.loading, DaffState.Resolving),
       };
 
     case DaffCartBillingAddressActionTypes.CartBillingAddressUpdateAction:
     case DaffCartAddressActionTypes.CartAddressUpdateAction:
       return {
         ...state,
-        ...setLoading(state.loading, DaffLoadingState.Mutating),
+        ...setLoading(state.loading, DaffState.Mutating),
       };
 
     case DaffCartBillingAddressActionTypes.CartBillingAddressLoadSuccessAction:
@@ -45,7 +45,7 @@ export function cartBillingAddressReducer<T extends DaffCart>(
           ...state.cart,
           billing_address: action.payload,
         },
-        ...setLoading(state.loading, DaffLoadingState.Complete),
+        ...setLoading(state.loading, DaffState.Complete),
       };
 
     case DaffCartBillingAddressActionTypes.CartBillingAddressUpdateSuccessAction:
@@ -57,7 +57,7 @@ export function cartBillingAddressReducer<T extends DaffCart>(
           ...state.cart,
           ...action.payload,
         },
-        ...setLoading(state.loading, DaffLoadingState.Complete),
+        ...setLoading(state.loading, DaffState.Complete),
       };
 
     case DaffCartBillingAddressActionTypes.CartBillingAddressLoadFailureAction:
@@ -66,7 +66,7 @@ export function cartBillingAddressReducer<T extends DaffCart>(
       return {
         ...state,
         ...addError(state.errors, action.payload),
-        ...setLoading(state.loading, DaffLoadingState.Complete),
+        ...setLoading(state.loading, DaffState.Complete),
       };
 
     default:
--- a/libs/cart/state/src/reducers/cart-order/cart-order-state.interface.ts
+++ b/libs/cart/state/src/reducers/cart-order/cart-order-state.interface.ts
@@ -1,11 +1,12 @@
 import { DaffCartOrderResult } from '@daffodil/cart';
 import {
-  DaffLoadingState,
   DaffStateError,
+  DaffMutableLoadingState,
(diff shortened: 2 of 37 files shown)


libs/cart/state/src/reducers/cart-billing-address/cart-billing-address.reducer.ts
```diff
@@ -42,7 +42,7 @@ export function cartBillingAddressReducer<T extends DaffCart>(
           ...state.cart,
           billing_address: action.payload
         },
-        ...setLoading(state.loading, DaffLoadingState.Complete),
+        ...setLoading(state.loading, DaffStates.Complete),
```
griest024, line 45: All of these completes should be stable instead, right? Are you going to do that in another PR?

libs/cart/state/src/reducers/cart-order/cart-order-state.interface.ts
```diff
@@ -1,8 +1,8 @@
-import { DaffLoadingState, DaffStateError } from '@daffodil/core/state';
+import { DaffStates, DaffStateError } from '@daffodil/core/state';
 import { DaffCartOrderResult } from '@daffodil/cart';
 
 export interface DaffCartOrderReducerState<T extends DaffCartOrderResult = DaffCartOrderResult> {
   cartOrderResult: T;
-  loading: DaffLoadingState;
+  loading: DaffStates;
```
griest024, line 6: This can still be typed as a loading mutable state, right?

libs/core/state/src/states/loading.ts
```diff
@@ -0,0 +1,7 @@
+import { DaffStates } from "./state.enum";
```
griest024, line 1: Use single quotes

libs/core/state/src/states/mutable-loading.ts
```diff
@@ -0,0 +1,7 @@
+import { DaffLoadingState } from "./loading";
+import { DaffStates } from "./state.enum";
+
+/**
+ * A description for an object that is only loadable, but changeable.
```
griest024, line 5: I don't think only loadable makes sense here

libs/core/state/src/states/state.enum.ts
```diff
@@ -0,0 +1,58 @@
+/**
+ * The DaffState enumerates the vast majority of state modifications that can exist 
+ * in UI applications. 
+ */
+export enum DaffStates {
```
griest024, line 5: Is it DaffState singular or DaffStates plural?

libs/core/state/src/states/state.enum.ts
```diff
@@ -0,0 +1,58 @@
...
+  /**
+   * The "Stable" state represents when no action is being taken upon an object.
+   */
+  Stable = 'Stable',
+
+  /**
+   * The "Resolving" state represents when there's uncertainty in the current status
```
griest024, line 12: Does it represent uncertainty? I would say it's explicitly just for the immutable loading of some data, analogous to a classic GET request.

libs/core/state/src/states/stateable.interface.ts
```diff
@@ -0,0 +1,9 @@
+import { DaffStates } from './state.enum';
+
+/**
+ * The interface an object must implement to consistently describe its current 
+ * state.
+ */
+export interface DaffStateable<T extends DaffStates> {
```
griest024, line 7: Would DaffLoadingState satisfy this generic constraint? I don't think it would.

griest024 (changes requested):

lderrickable (changes requested):

libs/cart/state/src/reducers/cart-billing-address/cart-billing-address.reducer.ts
damienwebdev: Yes. I'd like to leave this diff simply the transfer of DaffLoadingState to the equivalent DaffStates.

libs/core/state/src/states/state.enum.ts
damienwebdev: DaffState singular is better. Changed.

libs/core/state/src/states/mutable-loading.ts
damienwebdev: Fixed.

libs/core/state/src/states/loading.ts
damienwebdev: Resolved.

libs/cart/state/src/reducers/cart-order/cart-order-state.interface.ts
damienwebdev: Yes.

libs/core/state/src/states/state.enum.ts
damienwebdev: I think we need to highlight that we're typically doing this when we don't already know the information. E.g. We're uncertain whether or not the content we're attempting to load may or may not exist.

libs/core/state/src/states/stateable.interface.ts
damienwebdev: Yes, it does.

libs/core/state/src/states/state.enum.ts
```diff
@@ -0,0 +1,58 @@
...
+   * of a state, e.g. when we're loading something that hasn't been loaded into 
+   * state before.
+   */
+  Resolving = 'Resolving',
+
+  /**
+   * The "Adding" state represents a temporary period while an item is currently
+   * in the process of being added to state.
```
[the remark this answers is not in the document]
damienwebdev: a20c9be6

damienwebdev:
@lderrickable @griest024 I had to rebase the PR, please take a second look.

lderrickable (approved):

PR #1306, opened 2021-02-28
feat(cart): use cart resolution specific error in resolve effect

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Cart resolve effect passes original error to failure action with specialized text.

## What is the new behavior?
Cart resolve effect passes specialized error to failure action with original error text.

```
[x] Yes
[ ] No
```

If you construct `DaffResolveCartServerSide`, you must now pass an argument.

## Other information


libs/cart/state/src/effects/cart-resolver.effects.ts
```diff
@@ -60,29 +64,32 @@ export class DaffCartResolverEffects<T extends DaffCart = DaffCart>
...
+										new DaffResolveCartFailure(this.errorMatcher(
+                      new DaffCartNotFoundOrCreatedResolutionError(innerError.message)
+                    )),
 									);
 								}),
 							);
 						default:
-							error.message = 'Cart resolution has failed.';
-							return of(new DaffResolveCartFailure(this.errorMatcher(error)));
+							return of(new DaffResolveCartFailure(this.errorMatcher(
```
lderrickable, line 90: Just to make sure I understand the errorMatcher function...its purpose is to give the app dev a way to transform errors into state differently if they want to? So if the state that they define has some extra field in addition to `code` and `message` that they want to map to as well

lderrickable (changes requested):

libs/cart/state/src/effects/cart-resolver.effects.ts
griest024: Actually I don't think typescript will allow additional fields to be defined on the `DaffStateError` object. The main use case is either changing the error message or splitting a general error into more specific errors based on the message.

libs/cart/src/errors/resolution/cart-server-side-resolution.ts
```diff
@@ -0,0 +1,16 @@
+import { DaffError, DaffInheritableError } from '@daffodil/core';
+
+import { DaffCartErrorCodes } from '../codes.enum';
+
+/**
+ * An error thrown when the cart fails to resolve.
+ * This happens when the app in running in a server environment.
```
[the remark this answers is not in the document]
griest024: fixed

libs/cart/state/src/effects/cart-resolver.effects.spec.ts
```diff
@@ -91,7 +97,8 @@ describe('DaffCartResolverEffects', () => {
 			});
 
 			it('should emit a an action indicating that server side resolution occurred', () => {
```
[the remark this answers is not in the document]
griest024: fixed

libs/cart/state/src/effects/cart-resolver.effects.ts
lderrickable: > The main use case is either changing the error message or splitting a general error into more specific errors based on the message.

Ok, so it's just for allowing the app dev to do this? Because right now, we just map `code => code` and `message => message` by default, right?

griest024: > Ok, so it's just for allowing the app dev to do this? Because right now, we just map code => code and message => message by default, right?

Correct

lderrickable (approved):

lderrickable:
@griest024 do you want me to merge this or wait and see if the eslint pr gets merged?

griest024:
@lderrickable I say go ahead and merge, I can rebase the eslint PR

PR #1323, opened 2021-03-08
feat(core, product): rename DaffSortable to DaffOrderable

damienwebdev:
BREAKING CHANGE: the interface name changed, the above swap is sufficient to breakfix.

## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, we have an existing interface called "DaffSortable". This name will collide with the refactor on-going in DaffCategory. 

Fixes: N/A

## What is the new behavior?
In order to prevent the collision, I'm proposing "DaffOrderable" instead.

```
[x] Yes
[ ] No
```

### DIFF
--- a/libs/core/src/orderable/orderable.ts
+++ b/libs/core/src/orderable/orderable.ts
@@ -1,0 +1,9 @@
+/**
+ * An interface for models where the order of elements is important
+ * and definable. 
+ * 
+ * Disambiguation: This is does not relate to ecommerce "Orders", e.g.a purchase.
+ */
+export interface DaffOrderable {
+	order: number;
+}
--- a/libs/core/src/orderable/orderable.ts
+++ b/libs/core/src/orderable/orderable.ts
@@ -2,7 +2,7 @@
  * An interface for models where the order of elements is important
  * and definable.
  *
- * Disambiguation: This is does not relate to ecommerce "Orders", e.g.a purchase.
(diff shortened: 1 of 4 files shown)


libs/core/src/orderable/orderable.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * An interface for models where the order of elements is important
+ * and definable.
+ *
+ * Disambiguation: This is does not relate to ecommerce "Orders", e.g.a purchase.
```
griest024, line 5: missing space after `e.g.`

griest024 (changes requested):

lderrickable:
What does `DaffSortable` collide with in the category refactor? 

`Orderable` seems a little ambiguous in an e-commerce library, because it could mean that you can buy it or that you can sort it. So I think "sortable" is preferable, but I don't know what the particular collision is.

lderrickable:
The only alternative I can think of is `DaffPositioned` with the field being `position` rather than `order`. I think I actually like `DaffOrderable` better with the disambiguity comment though

libs/core/src/orderable/orderable.ts
damienwebdev: 13ac9d76

griest024 (approved):

PR #1326, opened 2021-03-08
chore(all): allow angular analytics

damienwebdev:


griest024 (approved):
The only docs I can find metnion the `cli.analyticsSharing` option but I trust that this is the right config option.

PR #1329, opened 2021-03-09
build(all): add recommended extensions for VSCode and update VSCode e…

damienwebdev:


.vscode/settings.json
```diff
@@ -23,5 +23,20 @@
     },
     "cSpell.words": [
         "Daffio"
+    ],
+    "editor.codeActionsOnSave": {
```
griest024, line 27: This is invalid JSON. Should be:
```json
"editor.codeActionsOnSave": {
    "source.fixAll.eslint": true
},
```

griest024 (changes requested):

.vscode/settings.json
damienwebdev: Yep -- I messed up a rebase here. My intent was to leave that particular setting up to devs who want it, but the other's are QoL improvements.

griest024 (approved):

PR #1353, opened 2021-03-12
feat(category): expose a configuration object for `@daffodil/category/state`

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, we force modification of tokens like `DaffDefaultCategoryPageSize`. We should instead be exposing configuration objects of modules.

Fixes: N/A

## What is the new behavior?
We expose a configuration object of the module.

```
[x] Yes
[ ] No
```

The DaffCategoryRoutingModule has been removed and the `DaffDefaultCategoryPageSize` has moved to `@daffodil/category/state`.

## Other information


libs/category/state/src/category-state.module.ts
```diff
@@ -16,4 +25,14 @@ import { DAFF_CATEGORY_STORE_FEATURE_KEY } from './reducers/public_api';
...
   ],
 })
-export class DaffCategoryStateModule { }
+export class DaffCategoryStateModule {
+  static forRoot(config: DaffCategoryStateConfiguration = defaultDaffCategoryStateConfiguration): ModuleWithProviders<DaffCategoryStateModule> {
+    return {
+      ngModule: DaffCategoryStateModule,
+      providers: [
+        { provide: DaffCategoryStateConfigurationToken, useValue: config },
+        { provide: DaffDefaultCategoryPageSize, useValue: config.defaultPageSize },
```
lderrickable, line 34: If this is provided here, I don't think `@daffodil/category/routing` should be its own subpackage

libs/category/state/src/config/config.ts
```diff
@@ -0,0 +1,25 @@
...
+   * The default size of the number of items to retrieve for a category
+   */
+  defaultPageSize: number;
+}
+
+/**
+ * The default configuration object used to configure the `@daffodil/category/state`
+ * package.
+ */
+export const defaultDaffCategoryStateConfiguration: DaffCategoryStateConfiguration = {
```
griest024, line 23: Should this be exported as well?

libs/category/state/src/category-state.module.ts
damienwebdev: @griest024 the routing module's were your idea, what do you think?

libs/category/state/src/config/config.ts
damienwebdev: Yes.

damienwebdev: 619b7146

libs/category/state/src/category-state.module.ts
griest024: I don't think it makes sense for guards and resolvers to be in a `state` package. I am still in favor of the `routing` subpackage, despite what the dependency graph looks like.

griest024 (approved):

libs/category/state/src/category-state.module.ts
lderrickable: so, what I'm saying is that it doesn't make sense to provide the default page size here if we're going to have a routing subpackage. If it doesn't make sense for guards and resolvers to be in a state package, then it also shouldn't make sense to provide the default configuration for the routing subpackage in the state package.

damienwebdev: I don't think this token is limited solely to the resolvers/guards.

lderrickable: oh, I hadn't noticed. That makes sense

lderrickable (approved):

PR #1364, opened 2021-03-17
feat(cart): set coupon code on invalid coupon error

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The invalid coupon error doesn't store the coupon code that caused the error.

## What is the new behavior?
The error stores the coupon code so that info can be used in the error transformer.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/driver/magento/src/errors/transform.ts
+++ b/libs/cart/driver/magento/src/errors/transform.ts
@@ -13,7 +13,7 @@ import {
 } from './map';
 
 
-function transformMagentoCartGraphQlError(error: ApolloError, requestPayload?: any): Error {
+function transformMagentoCartGraphQlError(error: ApolloError, requestPayload?): Error {
   // TODO: readdress this when we move to eslint
   // eslint-disable-next-line
   for (const code in DaffCartMagentoErrorMessageRegexMap) {
@@ -33,7 +33,7 @@ function transformMagentoCartGraphQlError(error: ApolloError, requestPayload?: a
   return daffTransformMagentoError(error, DaffCartMagentoErrorMap);
 };
 
-export function transformCartMagentoError(error, requestPayload?: any) {
+export function transformCartMagentoError(error, requestPayload?) {
   if (error.graphQLErrors?.length) {
     return transformMagentoCartGraphQlError(error, requestPayload);
   } else {
--- a/libs/cart/driver/magento/src/errors/transform.ts
+++ b/libs/cart/driver/magento/src/errors/transform.ts
@@ -1,6 +1,10 @@
 import { ApolloError } from '@apollo/client/core';
 
-import { DaffCartDriverErrorMap } from '@daffodil/cart/driver';
+import { DaffCartCoupon } from '@daffodil/cart';
+import {
+  DaffCartDriverErrorMap,
+  DaffInvalidCouponCodeError,
+} from '@daffodil/cart/driver';
 import { daffTransformMagentoError } from '@daffodil/driver/magento';
 
 import {
@@ -9,23 +13,29 @@ import {
 } from './map';
 
 
-function transformMagentoCartGraphQlError(error: ApolloError): Error {
+function transformMagentoCartGraphQlError(error: ApolloError, requestPayload?: any): Error {
   // TODO: readdress this when we move to eslint
   // eslint-disable-next-line
   for (const code in DaffCartMagentoErrorMessageRegexMap) {
     const matchIndex = error.graphQLErrors[0].message.search(DaffCartMagentoErrorMessageRegexMap[code]);
 
     if (matchIndex > -1 && DaffCartDriverErrorMap[code]) {
-      return new DaffCartDriverErrorMap[code](error.message);
+      const errObj = new DaffCartDriverErrorMap[code](error.message);
+
+      if (errObj instanceof DaffInvalidCouponCodeError) {
+        (<DaffInvalidCouponCodeError>errObj).coupon = (<DaffCartCoupon>requestPayload)?.code;
+      }
+
+      return errObj;
     }
   }
 
   return daffTransformMagentoError(error, DaffCartMagentoErrorMap);
 };
 
-export function transformCartMagentoError(error) {
+export function transformCartMagentoError(error, requestPayload?: any) {
   if (error.graphQLErrors?.length) {
-    return transformMagentoCartGraphQlError(error);
+    return transformMagentoCartGraphQlError(error, requestPayload);
   } else {
     return daffTransformMagentoError(error, DaffCartMagentoErrorMap);
   }
--- a/libs/cart/driver/magento/src/errors/transform.ts
+++ b/libs/cart/driver/magento/src/errors/transform.ts
@@ -13,7 +13,7 @@ import {
 } from './map';
 
 
-function transformMagentoCartGraphQlError(error: ApolloError, requestPayload?): Error {
+function transformMagentoCartGraphQlError(error: ApolloError, requestPayload?: unknown): Error {
   // TODO: readdress this when we move to eslint
   // eslint-disable-next-line
   for (const code in DaffCartMagentoErrorMessageRegexMap) {
@@ -33,7 +33,7 @@ function transformMagentoCartGraphQlError(error: ApolloError, requestPayload?): 
   return daffTransformMagentoError(error, DaffCartMagentoErrorMap);
 };
 
-export function transformCartMagentoError(error, requestPayload?) {
+export function transformCartMagentoError(error, requestPayload?: unknown) {
   if (error.graphQLErrors?.length) {
     return transformMagentoCartGraphQlError(error, requestPayload);
   } else {
--- a/libs/cart/driver/src/errors/invalid-coupon-code.ts
+++ b/libs/cart/driver/src/errors/invalid-coupon-code.ts
@@ -1,3 +1,4 @@
+import { DaffCartCoupon } from '@daffodil/cart';
 import {
   DaffError,
   DaffInheritableError,
@@ -12,7 +13,7 @@ import { DaffCartDriverErrorCodes } from './codes.enum';
 export class DaffInvalidCouponCodeError extends DaffInheritableError implements DaffError {
 	public readonly code: string = DaffCartDriverErrorCodes.INVALID_COUPON_CODE;
 
-	constructor(message?: string) {
+	constructor(message?: string, public coupon?: DaffCartCoupon['code']) {
 	  super(message);
 	}
 }
--- a/libs/cart/driver/magento/src/cart-coupon.service.spec.ts
+++ b/libs/cart/driver/magento/src/cart-coupon.service.spec.ts
@@ -10,7 +10,10 @@ import {
   DaffCart,
   DaffCartCoupon,
 } from '@daffodil/cart';
-import { DaffCartNotFoundError } from '@daffodil/cart/driver';
+import {
+  DaffCartNotFoundError,
+  DaffInvalidCouponCodeError,
+} from '@daffodil/cart/driver';
 import {
   MagentoCart,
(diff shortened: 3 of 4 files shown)


libs/cart/state/src/effects/cart-coupon.effects.spec.ts
```diff
@@ -132,6 +137,21 @@ describe('Daffodil | Cart | CartCouponEffects', () => {
       it('should dispatch a CartCouponApplyFailure action', () => {
         expect(effects.apply$).toBeObservable(expected);
       });
+
+      describe('because of an invalid coupon', () => {
+        beforeEach(() => {
+          const error = new DaffInvalidCouponCodeError('Invalid coupon code', mockCoupon.code);
```
lderrickable, line 143: I may be wrong, but I think you need to remove the `mockCoupon.code` from this error in order to test the code you added. Because the new code is what adds the coupon code to the error, right? When the error comes in, it doesn't have the coupon code; otherwise you wouldn't need to set it in the effect.

libs/cart/state/src/effects/cart-coupon.effects.ts
```diff
@@ -63,10 +64,19 @@ export class DaffCartCouponEffects<
...
-        ? new DaffCartStorageFailure(this.errorMatcher(error))
-        : new DaffCartCouponApplyFailure(this.errorMatcher(error)),
-      )),
+      catchError(error => {
+        switch (true) {
+          case error instanceof DaffStorageServiceError:
+            return of(new DaffCartStorageFailure(this.errorMatcher(error)));
+
+          case error instanceof DaffInvalidCouponCodeError:
+            (<DaffInvalidCouponCodeError>error).coupon = action.payload.code;
```
damienwebdev, line 73: Shouldn't this be handled where the error is thrown?

damienwebdev (changes requested):

libs/cart/state/src/effects/cart-coupon.effects.ts
griest024: The implementation is non-trivial. The code isn't in the error and the error transformer would have to be rewritten to accomodate that special case.

damienwebdev: What precisely does the relevant Magento error response look like?

griest024:
```
extensions: {category: "graphql-no-such-entity"}
locations: Array(1)
0: {line: 1, column: 65}
length: 1
__proto__: Array(0)
message: "The coupon code isn't valid. Verify the code and try again."
path: ["applyCouponToCart"]
```

lderrickable: The more I look at this, the more I think this shouldn't be in `@daffodil/cart/state`. Because the core code of the daffodil packages (agnostic to platform) shouldn't have platform specific code in it, and if the error message we're given from Magento doesn't have the coupon code, then the frontend shouldn't be responsible for knowing which coupon code caused the error. But I think the only other option is to write something custom on the backend, right?

griest024: > and if the error message we're given from Magento doesn't have the coupon code, then the frontend shouldn't be responsible for knowing which coupon code caused the error

But that is connecting the platform-agnostic frontend to the specific platform.

> Because the core code of the daffodil packages (agnostic to platform) shouldn't have platform specific code in it

What part of the code is platform-specific?

griest024: > But I think the only other option is to write something custom on the backend, right?

The error code in the Magento driver can be reworked to set the coupon code.

lderrickable: > What part of the code is platform-specific?

There is nothing explicitly platform-specific, but it is implicitly platform specific (meaning there is nothing named "magento", but there is code here that only exists because of the specific way that magento works). For example, if we implement Shopify and they do give the coupon code in the error message. That would make this code unnecessary for the shopify implementation, and it would only exist because magento doesn't give the coupon code in the error message. In that case the code that adds the coupon code to the daffodil error would be platform specific. If we could show that no platforms return the coupon code in the error messages, then it could be said that this is not platform specific. Unfortunately, there isn't a unifying set of rules that all platforms follow, so we can't know this and must assume that it is specific to magento.

> But that is connecting the platform-agnostic frontend to the specific platform's behavior.

I'm not sure what you mean by "that" here. As I've argued above, I think this code itself makes the frontend code, not platform-agnostic.

> The error transformer in the Magento driver can be reworked to set the coupon code.

If this works, that's cool. So would you need to send the request payload into the error transformer along with the magento error?

lderrickable: Oh I think I see what you meant by this:

>>and if the error message we're given from Magento doesn't have the coupon code, then the frontend shouldn't be responsible for knowing which coupon code caused the error.
>But that is connecting the platform-agnostic frontend to the specific platform's behavior.

I worded that kind of strangely. What I should have said is "the frontend shouldn't be responsible for knowing which coupon code caused the error, even if the error message we're given from Magento doesn't have the coupon code. It is magento's responsibility to pass @daffodil/state an error message with the coupon code."

griest024: > So would you need to send the request payload into the error transformer along with the magento error?

Yes

> there is code here that only exists because of the specific way that magento works

That makes sense.

> I'm not sure what you mean by "that" here. As I've argued above, I think this code itself makes the frontend code, not platform-agnostic.

Well I think this is tangential to the main point about where to set the coupon code on the error but:

> if the error message we're given from Magento doesn't have the coupon code

so here you reference a platform-specific behavior

> then the frontend shouldn't be responsible for knowing which coupon code caused the error

and then make a conclusion about platform-agnostic frontend behavior. So in that sense I say you have connected them.

I think its okay to draw inspiration for our agnostic frontend models from specific platforms but I think its wrong to create a hard link between their behaviors in this way. I think its reasonable for a frontend to expect a coupon error to contain information about which code caused the error; I see this on many ecommerce sites.

Back to the main point, I also think its reasonable to demand that the driver layer set the coupon code on the error despite how ugly that implementation might be. That's the purpose of our drivers: to encapsulate all the quirks and shortcomings of each individual platform behind a single well-defined interface. And although the interface doesn't specify the errors that can be thrown (and maybe it should), I agree that a fully-formed invalid coupon code error is implicitly part of the cart coupon driver's behavior.

I'll go ahead and make this change

griest024: > I worded that kind of strangely. What I should have said is "the frontend shouldn't be responsible for knowing which coupon code caused the error, even if the error message we're given from Magento doesn't have the coupon code. It is the platform's responsibility to pass @daffodil/state an error message with the coupon code."

okay I understand what you mean. yeah I briefly considered leaving it to the app to figure out which code caused the error but I came to the same conclusion: the app, and to a certain extend the state layer, isn't responsible for those kinds of things. It really is the responsibility of the backend, and therefore the drivers (if the backend doesn't do it), to provide that info.

griest024: @lderrickable @damienwebdev changed

libs/cart/driver/src/errors/invalid-coupon-code.ts
```diff
@@ -12,7 +13,7 @@ import { DaffCartDriverErrorCodes } from './codes.enum';
 export class DaffInvalidCouponCodeError extends DaffInheritableError implements DaffError {
 	public readonly code: string = DaffCartDriverErrorCodes.INVALID_COUPON_CODE;
 
-	constructor(message?: string) {
+	constructor(message?: string, public coupon?: DaffCartCoupon['code']) {
```
lderrickable, line 16: I think we should actually modify the error message instead. If we add this extra field, we'll end up needing to do the same thing in every platform coupon driver, so the app dev won't need to deal with the differences between the `DaffInvalidCouponCodeError`s when they change platforms. So maybe just append `coupon.code + " coupon code is invalid"` to the actual message.

griest024: > So maybe just append coupon.code + " coupon code is invalid" to the actual message.

That makes it significantly harder to use the coupon code in a custom error message, which is the use case being covered by this PR. I think it should stay as a field on the error.

> we'll end up needing to do the same thing in every platform coupon driver

This is preferable to forcing the app dev to regex the coupon code out of the error message.

libs/cart/driver/magento/src/errors/transform.ts
```diff
@@ -9,23 +13,29 @@ import {
 } from './map';
 
 
-function transformMagentoCartGraphQlError(error: ApolloError): Error {
+function transformMagentoCartGraphQlError(error: ApolloError, requestPayload?: any): Error {
```
damienwebdev, line 16: No explicit any.

damienwebdev (changes requested):

libs/cart/state/src/effects/cart-coupon.effects.ts
damienwebdev: Very much agree with this reasoning.

libs/cart/driver/magento/src/errors/transform.ts
griest024: removed

libs/cart/driver/src/errors/invalid-coupon-code.ts
damienwebdev: @lderrickable why would there be a difference in DaffInvalidCouponError between platforms?

lderrickable: @damienwebdev Because if Shopify returns the coupon code in the error message itself, then the Daffodil error wouldn't need the `public coupon?: DaffCartCoupon['code']` to pass on the coupon code information. So the app dev would need to stop using the `coupon` field, because that would now be `null`.

griest024: The use case in question is that the app dev wants to customize the error message in a way that requires having the coupon code. Regardless of whether or not the code is in the message, I think it should be set as a field on the error. The app dev should not be forced to regex the code out of the error message.

lderrickable: @griest024 oh ok. If we want to allow the app dev to use the coupon code in a custom error message, then I actually think it makes more sense for this to go in @daffodil/state, since we'll need the coupon field to be separate for every platform. Sorry for confusion. I guess I didn't understand the purpose of this PR.

libs/cart/driver/magento/src/errors/transform.ts
damienwebdev: Oh, not that, my bad. What I meant was, this needs to be strictly typed. We can't use `any` we should either set it to `unknown` or narrow the type to something more usable than any.

libs/cart/driver/src/errors/invalid-coupon-code.ts
lderrickable: Ok, I thought about this again and convinced myself that it makes more sense for this to be in the driver

libs/cart/driver/magento/src/errors/transform.ts
griest024: I think the best we can do is `unknown` unless we set it to a union of all the driver request types.

griest024: I just typed it as `unknown`

lderrickable (approved):

damienwebdev (approved):

PR #1379, opened 2021-03-22
feat(external-router): add id to resolvable route

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There is no way to get the id of an externally resolved entity through the external router driver.

## What is the new behavior?
The type now includes the entity ID, which is useful for downstream drivers.

```
[x] Yes
[ ] No
```

`id` is now a required field of `DaffExternallyResolvableUrl`

## Other information


damienwebdev (dismissed):
lgtm

libs/external-router/driver/magento/src/transforms/resolution-to-resolvable-url.ts
```diff
@@ -5,6 +5,7 @@ import { MagentoUrlResolver } from '../model/url-resolver';
 export const transformResolutionToResolvableUrl = (
   resolution: MagentoUrlResolver,
 ): DaffExternallyResolvableUrl => ({
+  id: resolution.id?.toString() || resolution.entity_uid,
```
lderrickable, line 8: I think this means we should now have two different transfomers

griest024: @damienwebdev do you agree?

damienwebdev: Eh, I don't mind splitting the transformer one way or the other, the type is so similar as it's a single key. If this were to expand in later versions, I can foresee a split transformer, but I can't say with any confidence that splitting right now is valuable.

griest024: well its done now anyway haha

lderrickable (approved):

Issue #1383, opened 2021-03-23
[FEAT] External Router function for injecting routes


griest024:
# :bulb: Feature request

## Feature Name

External Router function for injecting routes

## The Desired Behavior

External router should support mechanisms for configuring into what part of the routes config externally resolved routes get injected.

## Your Use Case

As an app dev, I want to be able to have externally resolved routes injected into route configs other than the root, also based on the route type (product vs. category)

PR #1390, opened 2021-03-25
feat(external-router): add route inserter injection token

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #1383 

```
[ ] Yes
[x] No
```

## Other information
I chose to make it a multi token because I don't think the app dev can always know where to insert each route at root. Each child routing module may need to decide that behavior for themselves.


libs/external-router/src/router/helper/insert-route-before-wildcard.ts
```diff
@@ -4,12 +4,13 @@ import {
...
 
 import { DaffExternalRouterNoWildcardError } from '../../errors/no-wildcard';
+import { DaffExternalRouterRouteInserter } from '../../model/route-inserter.type';
 
 /**
  * Inserts a Route into Routes right before the wildcard route.
  * If no wildcard is found, it fails fast with a `DaffExternalRouterNoWildcardError`.
  */
-export const insertRouteBeforeWildCard = (
+export const daffExternalRouterInsertRouteBeforeWildCard: DaffExternalRouterRouteInserter = (
```
damienwebdev, line 13: This function name is wild. Since this type of function won't exist elsewhere, I think we can simply remove the package prefix.

libs/external-router/src/external-router.module.ts
```diff
@@ -22,6 +23,7 @@ import { provideRoutesResolvableByType } from './token/type-resolvable-routes.to
 export class DaffExternalRouterModule {
   static forRoot(
     routes: TypeRoutePair[] = [],
+    inserters: DaffExternalRouterTypeRouteInserterPair[] = [],
```
damienwebdev, line 26: Can we instead expand the TypeRoutePair with an optional

```ts
{
 ...
 insertionStrategy?: DaffExternalRouterRouteInserter
 }
```

damienwebdev (changes requested):

libs/external-router/src/router/helper/insert-route-before-wildcard.ts
griest024: its possible that the app dev might want to delegate to this function so I think its best to export it.

libs/external-router/src/external-router.module.ts
griest024: That would force the app dev to provide the insertion strategy when they provide the route type. Is that what we want?

damienwebdev: Note the optional, in the event the prop isn't defined, we use the afterWildcard strategy.

libs/external-router/src/router/helper/insert-route-before-wildcard.ts
damienwebdev: I agree, I'm simply arguing to reduce the length of the function name.

griest024: `daffExternalRouterRootInsertionStrategy`? I think the type names should also be changed to "insertion strategy", as that is more descriptive.

libs/external-router/src/external-router.module.ts
griest024: yes but if they do want to provide a strategy, it must be when they provide the route type.

griest024: I suppose its likely that they would be able to provide the strategy at that point. As soon as you know the route type you most likely know how to insert it into your route config.

damienwebdev: I suspect so, and it keeps the API much more simple.

libs/external-router/src/router/helper/insert-route-before-wildcard.ts
damienwebdev: daffInsertRoutrBeforeWildCardStrategy

griest024: changed

libs/external-router/src/external-router.module.ts
griest024: changed

libs/external-router/src/model/type-route-pair.ts
```diff
@@ -1,17 +1,25 @@
...
 /**
  * A model used to describe the relationship between a {@link DaffExternalRouterType}
  * and its corresponding route.
  *
- * The TypeRoutePair is the pre-provided "glue" that a developer will provide
+ * The DaffTypeRoutePair is the pre-provided "glue" that a developer will provide
  * ahead-of-time so that that the External Router knows what the list of available
  * "types" are (along with that type's corresponding route).
  *
+ * An insertion strategy can be specified if the route needs to be inserted somewhere
```
lderrickable, line 13: I think it should be mentioned that the default insertion strategy will be the `daffInsertRouteBeforeWildCardStrategy` if none is provided. That's accurate, right?

griest024: Yes that is correct, I'll add it

griest024: added

lderrickable:
lgtm

libs/external-router/src/transform/resolved-route-to-insertion-strategy.ts
```diff
@@ -0,0 +1,17 @@
...
+import { DaffTypeRoutePair } from '../model/type-route-pair';
+
+/**
+ * Finds the inserter associated with the resolved route, if such an inserter exists.
+ */
+export const daffTransformResolvedRouteToInsertionStrategy = (
+  resolvedRoute: DaffExternallyResolvableUrl,
+  routeTypes: DaffTypeRoutePair[],
+): DaffExternalRouterInsertionStrategy => {
+  const routeType = routeTypes
```
damienwebdev, line 12: We already do a lookup very similar to this in another function, is there a way to merge these two lookup together so that we don't have to do this lookup twice?

damienwebdev: https://github.com/graycoreio/daffodil/blob/2d7ad2766bd0cafe9c15a3e20c7cb08c427b91bb/libs/external-router/src/transform/resolved-route-to-route.ts

damienwebdev (changes requested):

libs/external-router/src/transform/resolved-route-to-insertion-strategy.ts
griest024: changed

damienwebdev (approved):
lgtm

PR #1398, opened 2021-03-31
feat(product): add selector and facade field for composite product di…

lderrickable:
…scount amount

## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Added a selector (and corresponding facade field) for getting the discount amount for a composite product. Returns undefined if all required items are not chosen.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
+++ b/libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
@@ -119,7 +119,7 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
       const appliedOptions = getAppliedOptionsForConfiguration(<DaffCompositeProduct>product, appliedConfigurationItem);
 
       return (<DaffCompositeProduct>product).items.reduce((acc, item) =>
-        daffAdd(acc, daffMultiply(appliedOptions[item.id].discount.amount, appliedOptions[item.id].quantity)),
+        daffAdd(acc, daffMultiply(appliedOptions[item.id]?.discount?.amount || null, appliedOptions[item.id].quantity)),
       product.discount.amount);
     },
   );
--- a/libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
+++ b/libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
@@ -39,6 +39,10 @@ export interface DaffCompositeProductMemoizedSelectors {
 	 */
 	selectCompositeProductPricesAsCurrentlyConfigured: MemoizedSelectorWithProps<Record<string, any>, { id: DaffCompositeProduct['id'] }, DaffPriceRange>;
 	/**
+	 * Get the discount amount for a composite product. This value will be undefined if all required options are not chosen.
+	 */
+	selectCompositeProductDiscountAmount: MemoizedSelectorWithProps<Record<string, any>, { id: DaffCompositeProduct['id'] }, number>;
+	/**
 	 * Get the discount percent for a composite product. This value will be undefined if all required options are not chosen.
 	 * Note: this percent is computed client-side and should be treated as an estimate rather than an exact value.
 	 */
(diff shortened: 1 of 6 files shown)


libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -99,6 +103,27 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
...
+    selectProductEntities,
+    selectCompositeProductAppliedOptionsEntities,
+    (products, appliedOptionsEntities, props) => {
+      const product = selectProduct.projector(products, { id: props.id });
+      if(product.type !== DaffProductTypeEnum.Composite) {
+        return undefined;
+      }
+
+      const appliedConfigurationItem = appliedOptionsEntities[product.id].items;
+      if((<DaffCompositeProduct>product).items.filter(item => item.required && appliedConfigurationItem[item.id].value === null).length > 0) {
```
griest024, line 116: It would be more performant to implement this with `indexOf` since you're only looking for a single element that matches this criteria. As soon as you find it, you know that you can exit and return, you don't need to keep checking the rest of the list.

libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
```diff
@@ -99,6 +103,27 @@ const createCompositeProductSelectors = (): DaffCompositeProductMemoizedSelector
...
+        return undefined;
+      }
+      const appliedOptions = getAppliedOptionsForConfiguration(<DaffCompositeProduct>product, appliedConfigurationItem);
+
+      return (<DaffCompositeProduct>product).items.reduce((acc, item) =>
+        daffAdd(acc, daffMultiply(appliedOptions[item.id].discount.amount, appliedOptions[item.id].quantity)),
```
griest024, line 122: Is `appliedOptions[item.id]` guaranteed to exist?

griest024 (changes requested):

libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: nope. fixed and added a test for it.

libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: I think `indexOf` won't work, because I'd need to pass it an equivalent object rather than being able to define a custom condition. Am I crazy?

griest024: you are not, I meant to say `findIndex`

libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
griest024: you need to guard the quantity expression as well, right?

I'm not sure optional chaining is the way to go here, I might just use a ternary like:

```ts
appliedOptions[item.id] ? daffAdd(acc, daffMultiply(appliedOptions[item.id].discount.amount, appliedOptions[item.id].quantity)) : acc
```

so that it avoids calling arithmetic function unless its actually necessary.

lderrickable: Turned out that `appliedOptions[item.id]` is guaranteed to exist, but your way makes more sense. I just needed to guard up to `appliedOptions[item.id].discount.amount`

libs/product/state/src/selectors/composite-product/composite-product.selectors.ts
lderrickable: changed

griest024: ah sorry...your original implementation here is actually the only one that works on IE11. I though `findIndex` had IE support...

lderrickable: np, reverted

griest024 (approved):

PR #1400, opened 2021-03-31
feat(product): add getByUrl actions and effects

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No way to resolve products by URLs through state.

## What is the new behavior?
Adds product and product page actions for resolving a product by URL.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/product/state/src/reducers/product/product.reducer.spec.ts
+++ b/libs/product/state/src/reducers/product/product.reducer.spec.ts
@@ -8,6 +8,7 @@ import {
 } from '@daffodil/product/state';
 import { DaffProductFactory } from '@daffodil/product/testing';
 
+import { DaffProductPageLoadByUrl } from '../../actions/public_api';
 import {
   initialState,
   daffProductReducer,
@@ -34,6 +35,24 @@ describe('Product | Product Reducer', () => {
       const result = daffProductReducer(initialState, action);
 
       expect(result).toBe(initialState);
+    });
+  });
+
+  describe('when ProductLoadByUrlAction is triggered', () => {
+    let result;
+
+    beforeEach(() => {
+      const productLoadAction = new DaffProductPageLoadByUrl(product.url);
+
+      result = daffProductReducer(initialState, productLoadAction);
+    });
+
+    it('sets loading state to true', () => {
+      expect(result.loading).toEqual(true);
+    });
+
+    it('sets selectedProductId to the product url', () => {
+      expect(result.selectedProductId).toEqual(product.url);
     });
   });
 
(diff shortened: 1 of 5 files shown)


libs/product/state/src/reducers/product/product-reducer-state.interface.ts
```diff
@@ -5,7 +5,7 @@ import { DaffProduct } from '@daffodil/product';
  * Interface for product state.
  */
 export interface DaffProductReducerState {
-  selectedProductId: DaffProduct['id'];
+  selectedProductId: DaffProduct['id'] | DaffProduct['url'];
```
lderrickable, line 8: This seems like it might become problematic. But I think this field is deprecated, so maybe just revert this change and we'll just remove this field entirely in a future PR.

libs/product/state/src/actions/product.actions.ts
```diff
@@ -7,11 +7,23 @@ import { DaffProduct } from '@daffodil/product';
...
     ProductLoadSuccessAction = '[Product] Load Success Action',
     ProductLoadFailureAction = '[Product] Load Failure Action',
 }
 
+/**
+ * Triggers a request for a Product.
+ *
+ * @param payload - URL of requested Product
+ */
+export class DaffProductLoadByUrl<T extends DaffProduct = DaffProduct> implements Action {
```
lderrickable, line 21: Can you think of an example when this would be used? I can't

griest024: Honestly I'm not even sure what the purpose of the non-page load actions are so I'll defer to you on this.

griest024: removed

libs/product/state/src/reducers/product/product-reducer-state.interface.ts
griest024: removed

libs/product/state/src/actions/product.actions.ts
lderrickable: It would be a similar usage as loading a category for a home page or something. Probably less useful since I would think that loading categories that aren't on category pages is a lot more common than loading a single product on a page that isn't for that particular product.

libs/product/state/src/reducers/product/product.reducer.spec.ts
```diff
@@ -37,6 +38,24 @@ describe('Product | Product Reducer', () => {
...
+      const productLoadAction = new DaffProductPageLoadByUrl(product.url);
+
+      result = daffProductReducer(initialState, productLoadAction);
+    });
+
+    it('sets loading state to true', () => {
+      expect(result.loading).toEqual(true);
+    });
+
+    it('sets selectedProductId to the product url', () => {
```
lderrickable, line 54: remove this test, since it is an accidental byproduct of deprecated code, and not a feature we are intending.

griest024: removed

lderrickable (approved):

PR #1401, opened 2021-04-03
Category applied filters

lderrickable:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information


libs/category/src/filters/behaviors/apply/apply-filter.spec.ts
```diff
@@ -0,0 +1,17 @@
+describe('daffApplyFilter', () => {
+  it('should apply a toggle request', () => {
```
lderrickable, line 2: I don't understand what this function has to do with a toggle request. Should this be Equal vs Range instead of toggle vs idempotent?

libs/category/src/filters/behaviors/apply/apply-request-to-filters.spec.ts
```diff
@@ -0,0 +1,17 @@
+describe('daffApplyRequestToFilters', () => {
+  it('should apply a request', () => {
+
+  });
+
+  it('should apply a toggle request', () => {
```
lderrickable, line 6: I don't understand what this function has to do with a toggle request. Should this be Equal vs Range instead of toggle vs idempotent?

libs/category/src/filters/type/equal/behaviors/apply/apply.ts
```diff
@@ -0,0 +1,27 @@
...
+export const daffApplyFilterEqual = (
+  request: DaffCategoryFilterEqualRequest,
+  filter: DaffCategoryEqualFilter,
+): DaffCategoryEqualFilter => ({
+  ...filter,
+  options: request.value.reduce((acc, value) => {
+    const option = filter.options[value];
+
+    if (option) {
+      acc[value] = {
```
lderrickable, line 19: This doesn't really make sense to me. The key for the dictionary entry of `options` is the `value` field of the entry itself? Adding docs to the `DaffCategoryFilterEqualOption` model and the fields inside it would help clarify this.

libs/category/src/filters/behaviors/toggle/toggle-request-on-filters.spec.ts
```diff
@@ -0,0 +1,17 @@
+describe('daffApplyRequestToFilters', () => {
+  it('should apply a request', () => {
```
lderrickable, line 2: doesn't this only apply toggle requests?

libs/category/src/filters/behaviors/apply/apply-filter.spec.ts
griest024: I don't think this spec name is correct. Damien and I did not do anything at all to the specs after a significant amount of work on the source files so most of them are likely wrong.

libs/category/src/filters/type/equal/behaviors/apply/apply.ts
griest024: The key must be the value to ensure that it is unique

lderrickable: Oh, I missed the part where `DaffCategoryEqualFilter extends DaffCategoryFilterBase`. I thought you were losing the filter name information, but its just on the `DaffCategoryFilterBase`

griest024: The models are slightly convoluted because we had to use base types and union types to workaround TS's inability to infer some types based on the usage

libs/category/src/filters/type/range/behaviors/apply/apply.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * @idempotent
+ */
+export const daffApplyFilterRange = <T>(
+  request: DaffCategoryFilterRangeRequestBase<T>,
+  filter: DaffCategoryFilterRangeBase<T>,
+): DaffCategoryFilterRangeBase<T> => ({
+    ...filter,
+    options: {
+      [daffCategoryComputeFilterRangePairLabel(request.value.min, request.value.max)]: {
```
lderrickable, line 18: Won't this lose idempotence? If there is a filter applied for `price: 30-40` and you add another filter for `price: 40-50`, shouldn't there be either two filters for `30-40` and `40-50` or one filter for `30-50`? This looks like it would override the original `price: 30-40`.

libs/category/src/models/filters/type/range/request-base.ts
```diff
@@ -0,0 +1,8 @@
+import { DaffCategoryFilterRangeRequestOption } from '../range/public_api';
+import { DaffCategoryFilterRangeType } from './filter-type';
+
+export interface DaffCategoryFilterRangeRequestBase<T> {
+	type: DaffCategoryFilterRangeType;
+	name: string;
+	value: DaffCategoryFilterRangeRequestOption<T>;
```
lderrickable, line 7: Is it intentional that you can only add one ranged filter to a request at a time?

libs/category/src/filters/type/range/behaviors/apply/apply.ts
griest024: We discussed this at length. We might support multiple range filters at some later date, but for now, applying a range filter option always removes the previous one.

damienwebdev: @lderrickable, I'm not sure this violates idempotence. 

Simply stated, idempotence implies that f(a) = b that f(b) = b. 

So, upon reapplying the function to a result, you will get the same result.

Since the range filters "apply" always gets you to the same result assuming you use the same request on the filters returned from the prior apply, apply is idempotent.

PR #1424, opened 2021-04-09
add validator tests

griest024:
also fixes an issue with the filter factory


libs/category/testing/src/factories/filters/filter.factory.ts
```diff
@@ -20,10 +20,10 @@ export class DaffCategoryFilterFactory extends DaffModelFactory<DaffCategoryFilt
     super(<any>MockDaffCategoryFilter);
   }
 
-  create(partial?: Partial<DaffCategoryFilter>): DaffCategoryFilter {
+  create(partial: Partial<DaffCategoryFilter> = {}): DaffCategoryFilter {
     return {
       ...new this.type(),
-      ...faker.random.number({ min: 1, max: 2 }) === 2 ? this.equalFactory.create() : this.rangeFactory.create(),
```
damienwebdev, line 26: I had intentionally left this off as it doesn't make a lot of sense to provide a partial for a union type.

libs/category/src/filters/validators/request-type-match.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    equalFilterRequest = equalFilterRequestFactory.create();
+  });
+
+  describe('when the types do not match', () => {
+    beforeEach(() => {
+      filter = rangeFilterFactory.create();
+    });
+
+    it('should throw a DaffCategoryFilterRequestTypeMismatch indicating a name mismatch', () => {
+      const error = new DaffCategoryFilterRequestTypeMismatch('Filter types don\'t match');
```
damienwebdev, line 38: I've preferred toThrowMatching to prevent test changes when changing error message text.

damienwebdev (changes requested):

libs/category/testing/src/factories/filters/filter.factory.ts
griest024: There are still guaranteed overlaps in fields like `name` and `type`.

libs/category/src/filters/validators/request-type-match.spec.ts
griest024: changed

libs/category/testing/src/factories/filters/filter.factory.ts
damienwebdev: Then we should modify the partial's type (and the internal factory calls) to match that more accurately.

griest024: like this?

griest024: like this?

damienwebdev: No, I think that's worse. I think if the user should be force to provide a "type" argument at least and we should use that type to infer the appropriate child factory to call.

griest024: changed. lmk if you think I should test via a factory spy or if just asserting the type like I did is okay

damienwebdev (approved):
lgtm

PR #1433, opened 2021-04-13
fix test imports

griest024:
also exports the filter errors

### DIFF
--- a/libs/category/src/filters/public_api.ts
+++ b/libs/category/src/filters/public_api.ts
@@ -8,7 +8,6 @@ export * from './to-request/public_api';
 export * from './type/range/public_api';
 export * from './type/equal/public_api';
 export * from './validators/public_api';
-export * from './errors/public_api';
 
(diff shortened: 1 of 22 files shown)


libs/category/src/filters/public_api.ts
```diff
@@ -8,6 +8,7 @@ export * from './to-request/public_api';
 export * from './type/range/public_api';
 export * from './type/equal/public_api';
 export * from './validators/public_api';
```
damienwebdev, line 10: I don't think the errors should be exported here, they should only be from the root errors API.

damienwebdev (changes requested):

libs/category/src/filters/public_api.ts
griest024: moved export

damienwebdev: I misunderstood what was in the API of filter errors, my mistake. They can be exported from the filters public API.

griest024: reverted

PR #1434, opened 2021-04-13
add some missing tests

griest024:


libs/category/src/filters/type/range-numeric/to-request/filter-to-request.spec.ts
```diff
@@ -0,0 +1,56 @@
+import { TestBed } from '@angular/core/testing';
+import { DaffCategoryFilterRangePair } from 'libs/category/src/models/public_api';
```
damienwebdev, line 2: Import path

griest024: fixed

PR #1438, opened 2021-04-14
feat(core): add a new jasmine subpackage with idempotence matchers

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, we don't have a way to test function idempotence.

Fixes: N/A

## What is the new behavior?
This adds a way to test function idempotence quickly and effectively. This will be used in #1428 

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/tools/jasmine/README.md
+++ b/tools/jasmine/README.md
@@ -2,19 +2,13 @@
 
 This packages provides some useful matchers to make testing applications with jasmine a little nicer.
 
-## Installation
-
-```bash
-npm install @daffodil/jasmine --save
-```
-
 ## With Tests
 In your `test.ts` file:
 
 
 ```ts
 //Import Daffodil Jasmine Features
-import { setup } from '@daffodil/core/testing/jasmine';
+import { setup } from '@daffodil/jasmine';
 setup();
 
 declare const require: any;
--- a/tools/jasmine/README.md
+++ b/tools/jasmine/README.md
@@ -1,0 +1,21 @@
+# @daffodil/jasmine
+
+This packages provides some useful matchers to make testing applications with jasmine a little nicer.
+
+## Installation
+
+```bash
+npm install @daffodil/jasmine --save
+```
+
+## With Tests
+In your `test.ts` file:
+
+
+```ts
+//Import Daffodil Jasmine Features
+import { setup } from '@daffodil/core/testing/jasmine';
+setup();
+
+declare const require: any;
+```
--- a/tools/jasmine/jasmine.json
+++ b/tools/jasmine/jasmine.json
@@ -1,0 +1,7 @@
+{
+  "spec_files": [
+      "**/*[sS]pec.ts"
+  ],
+  "stopSpecOnExpectationFailure": false,
+  "random": true
+}
--- a/tools/jasmine/package.json
+++ b/tools/jasmine/package.json
@@ -1,0 +1,24 @@
+{
+  "name": "@daffodil/jasmine",
+  "version": "0.0.0-PLACEHOLDER",
+  "description": "A package that provides some useful matchers to make testing applications with jasmine a little nicer",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/graycoreio/daffodil"
+  },
+  "author": "Graycore LLC",
+  "license": "MIT",
+  "bugs": {
+    "url": "https://github.com/graycoreio/daffodil/issues"
+  },
+  "scripts": {
+    "build": "tsc",
+    "lint": "eslint src/**/*.ts",
+    "lint:fix": "eslint src/**/*.ts --fix",
+    "test": "ts-node --project tsconfig.spec.json -r tsconfig-paths/register ../../node_modules/jasmine/bin/jasmine.js --config=jasmine.json"
+  },
+  "homepage": "https://github.com/graycoreio/daffodil",
+  "peerDependencies": {
+    "jasmine": "0.0.0-PLACEHOLDER"
+  }
+}
--- a/tools/jasmine/package.json
+++ b/tools/jasmine/package.json
@@ -19,6 +19,7 @@
   },
   "homepage": "https://github.com/graycoreio/daffodil",
   "peerDependencies": {
-    "jasmine": "0.0.0-PLACEHOLDER"
+    "jasmine": "0.0.0-PLACEHOLDER",
+    "@types/jasmine": "0.0.0-PLACEHOLDER"
   }
 }
--- a/tools/jasmine/src/idempotence.spec.ts
+++ b/tools/jasmine/src/idempotence.spec.ts
@@ -1,0 +1,29 @@
+import { setup } from './index';
+
+describe('@daffodil/core/testing/jasmine | idempotenceMatcher', () => {
+  const anArityOneIdempotentFunction = (a: number) => a;
+  const anArityOneNonIdempotentFunction =  (a: number) => a * 2;
+  const anArityTwoIdempotentFunctionOverTheSecondArgument = (a: number, b: number) => a * 2;
+
+  setup();
+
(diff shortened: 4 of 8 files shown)


griest024 (approved):

libs/core/testing/jasmine/src/idempotence.ts
```diff
@@ -0,0 +1,26 @@
+import MatchersUtil = jasmine.MatchersUtil;
+
+/**
+ * A matcher for jasmine that would allow you to test whether or not a function
+ * was idempotent.
+ *
+ * A simple example, where the co-domain and the domain of the function under test
+ * are the same would be:
```
lderrickable, line 8: The code makes sense to me, though I don't understand this doc. What is the co-domain and what is the domain of a function? Could you post some literature on them and the differences between the two here?

lderrickable: Nevermind, I found a helpful youtube video on it. I think this doc isn't quite right though. Or at the very least, it just makes everything more confusing. The function under test is the function being tested for idempotence, right? And you're not testing that function for whether or not the co-domain and the domain are the same even if they are in the second instance of calling the function under test. The function that always should have the same domain and co-domain is the function that wraps the function under test. Do you see how that might be confusing?

damienwebdev: So, I clarified this for a better example to explain what I meant.

damienwebdev: The first example describes how one would test a function of one argument that is idempotent "over" that argument.

The second examples describe idempotence when multiple arguments are involved.

Unfortunately, I don't know of, or couldn't find any literature on the topic of multivariate idempotence, so @griest024 and I jokingly came up with "spiritually" idempotent (as in the function exhibits the "spirit" of idempotence).

damienwebdev: So upon further research, it is correct to say that 

```ts
type f = (x: SomeType, b: SomeOtherType): SomeOtherType;

If, for each x in SomeType, there exists a "c" in SomeOtherType, and f(x,b) = c and f(x,c) = c
``` 

Then it is correct to say that for a higher arity function, `f is idempotent over "b"` as `f` has a fixed point (that is to say it results to the same thing) for a given "a" and each `b`.

lderrickable: As I understand it, idempotence refers to an operation, not the argument passed in, so "which argument is idempotent" doesn't make sense to me. I would think that if a function took two of the same arguments and always returned the same result, the function is idempotent. If a function could take two arguments that were the same both times and return a different answer then the function is not idempotent. I don't think that identifying which argument causes the return value to change has any bearing on whether or not the function is idempotent.

lderrickable: I think a doc that just explains how it works may be more helpful:

This jasmine matcher tests idempotence by calling a given function two times. The first call is made with a given argument, while the second is made with the returned value of the first call. If the two return values are equal, then the function under test is idempotent. Here is an example use case:
```ts
multiplyByOne(arg) {
   return arg * 1;
}

expect((arg?: YourType) => (multiplyByOne(arg))).toBeIdempotent();
```

lderrickable: Also, I think I see what you were trying to say with the multiple arguments. That's basically how we will need to test a lot of the behaviors. But I think the explanation you gave might just make things more confusing, and I think it would be best just to trust that the dev will be able to extend the matcher for their usecase.

damienwebdev: Idempotence generally refers to the application of an operation to a certain argument in an arity one function.

In this case, formally you can write

Simple
===
image%20=%20b%20\implies%20f(b)%20=%20b)

The problem is that when you have a >=1 arity function

Which of the following are you referring to?

Arrity 2. right-argument idempotent (I refrain from writing left-arg for brevity)
===
image%20=%20b%27%20\implies%20f(a,b%27)%20=%20b%27)

Arrity 2. dual-argument idempotent
===
image%20=%20b%27%20\implies%20f(a%27,b%27)%20=%20b%27)

Apologies for the formal math here, it's just more precise at the loss of general legibility.

If we consider where we're planning on using this (in applied filters), it may be easier to consume.

Consider:

```ts
export const daffApplyFilter = (request: DaffCategoryFilterRequest, filters: Dict<DaffCategoryFilterRequest>): Dict<DaffCategoryFilterRequest> => {...};
```

We say that this function is idempotent over "filters" because we can feed the result filters back into the function with the same request and get the same result.

The confusion surrounds which argument makes sense to pass the result back into the function as an argument.

As an example of where this would matter, consider a trivial function where both arguments are the same type:

```ts
export const anIdempotentFunction = (a: number, b: number) => {
 return a * 2;
}
```

This function is idempotent over `b` but not idempotent over `a`, because we must choose which arg to feed the result back into (if the result type isn't a pair). If we choose to fill `a`, it's clearly not idempotent.

```ts
anIdempotentFunction(1,2) //2
anIdempotentFunction(anIdempotentFunction(1,2),2)//4
```

Yet, if we choose to fill `b`, it is idempotent.

```ts
anIdempotentFunction(1,2)//2
anIdempotentFunction(1,anIdempotentFunction(1,2))//2
```

griest024: > The function that always should have the same domain and co-domain is the function that wraps the function under test

This is correct, as the custom matcher can't be expected to know how to invoke the function under test. Each test is responsible for crafting the callback such that it correctly inherits the idempotence from the function under test.

> As I understand it, idempotence refers to an operation, not the argument passed in, so "which argument is idempotent" doesn't make sense to me.

That understanding is mostly right. At least for the "rigorous" definition of idempotence (to the extent that the rigorous definition of idempotence can be applied to higher-arity functions) it's very relevant to consider for which argument the function is idempotent. Its kinda like taking partial derivatives in multivar calc. The function can have very different behavior depending on which arg you target in the derivative.

The geometric understanding of the partial derivative is that you fix one arg and vary the other. The idempotence analog of this concept is that one arg is fixed (for example a filter request) while the one under idempotence test (for example the applied filter) gets set to the function result.

```ts
expect(idem => daffApplyFilter(request, idem || filter)).toBeIdempotent()
```

I wouldn't word it like "the argument is idempotent" however, I would say "_f_ is idempotent with respect to _x_" or "_f_ is idempotent over _x_ (or the domain specified by _x_'s type)".

I think @lderrickable makes a good point: this matcher needs some good examples. Specifically I think giving an example function definition for each case (1-arity, higher-arity) and then a case that fails would be helpful.

The 1-arity case is a simple example to familiarize the concept. The higher-arity case shows how and explains why the callback needs to be formulated in the way that it does. The example that fails idempotence will maybe be helpful to people who don't understand why idempotence is good.

I think these cases would be good for a unit test. We can point to them in the documentation as examples.

lderrickable: @damienwebdev ok, I see why we need examples for the higher-arity usecases. I still stand by what I said about the docs being more confusing than helpful though (not sure if I actually said that), and I agree with pretty much everything @griest024 said.

damienwebdev: I've added specs @lderrickable -- please take a look.

damienwebdev:
@griest024 I wasn't convinced after thinking about this package, that it belonged in the currently published core package, so I elected to move it into a tool.

libs/core/testing/jasmine/src/idempotence.ts
lderrickable: lgtm

lderrickable (approved):

griest024 (approved):

PR #1459, opened 2021-04-21
 feat(category): add kinded category requests

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Category requests assume they are always using the ID.

## What is the new behavior?
Category requests now specify what kind of identifier they use.

```
[x] Yes
[ ] No
```

`DaffCategoryPageRequestKind` now must be present on `DaffCategoryRequest` objects.

## Other information

### DIFF
--- a/libs/category/driver/magento/src/category.service.ts
+++ b/libs/category/driver/magento/src/category.service.ts
@@ -11,6 +11,7 @@ import {
 } from 'rxjs/operators';
 
 import {
+  DaffCategoryIdRequest,
   DaffGetCategoryResponse,
   daffApplyRequestsToFilters,
   DaffCategoryPageMetadata,
@@ -63,9 +64,9 @@ export class DaffMagentoCategoryService implements DaffCategoryServiceInterface 
   /**
    * Gets a category based on parameters. Default current_page is 1, and default page_size is 20.
    *
-   * @param categoryRequest A DaffCategoryRequest object.
+   * @param categoryRequest A DaffCategoryIdRequest object.
    */
-  get(categoryRequest: any): Observable<any> {
+  get(categoryRequest: DaffCategoryIdRequest): Observable<DaffGetCategoryResponse> {
     return combineLatest([
       this.apollo.query<MagentoGetACategoryResponse>({
         query: MagentoGetCategoryQuery,
@@ -91,7 +92,7 @@ export class DaffMagentoCategoryService implements DaffCategoryServiceInterface 
     );
   }
 
-  private getProductsQueryVariables(request: DaffCategoryRequest): MagentoGetProductsByCategoriesRequest {
+  private getProductsQueryVariables(request: DaffCategoryIdRequest): MagentoGetProductsByCategoriesRequest {
     const queryVariables = {
       filter: this.magentoAppliedFiltersTransformer.transform(request.id, request.filter_requests),
     };
--- a/libs/category/driver/magento/src/category.service.ts
+++ b/libs/category/driver/magento/src/category.service.ts
@@ -60,7 +60,6 @@ export class DaffMagentoCategoryService implements DaffCategoryServiceInterface 
   ) {}
 
   //todo the MagentoGetCategoryQuery needs to get its own product ids.
-  // TODO(griest024): fix return type
   /**
    * Gets a category based on parameters. Default current_page is 1, and default page_size is 20.
    *
--- a/libs/category/src/models/category-request.ts
+++ b/libs/category/src/models/category-request.ts
@@ -4,14 +4,62 @@ import { DaffSortDirectionEnum } from '@daffodil/core/state';
 import { DaffCategoryFilterRequest } from './filters/public_api';
 
 /**
+ * When you request a category page, there's two ways to do so, via URI or via ID.
+ */
+export enum DaffCategoryPageRequestKind {
+  URI = 'URI',
+  // eslint-disable-next-line @typescript-eslint/no-shadow
+  ID = 'ID'
+}
+
+/**
  * A model of the request sent to the {@link DaffCategoryServiceInterface} to retrieve information
  * about a category.
  */
-export interface DaffCategoryRequest {
-  id: ID;
+export interface DaffCategoryBaseRequest {
+  /**
+   * The properties by which to filter the items of the category being requested.
+   */
   filter_requests?: DaffCategoryFilterRequest[];
+
+  /**
+   * The option by which to sort a categories' items.
+   */
   applied_sort_option?: string;
+
+  /**
+   * The sort direction to sort the categories' items by.
+   */
   applied_sort_direction?: DaffSortDirectionEnum;
+
+  /**
+   * What page of the categories' items to retrieve.
+   */
   current_page?: number;
+
+  /**
+   * The number of items per-page to request.
+   */
   page_size?: number;
 }
+
+/**
+ * The request used to retrieve a category and its relevant items.
+ */
+export type DaffCategoryRequest = DaffCategoryIdRequest | DaffCategoryUriRequest;
+
+/**
+ * A request used to retrieve a category by its uid.
+ */
+export interface DaffCategoryIdRequest extends DaffCategoryBaseRequest {
+  id: ID;
+  kind: DaffCategoryPageRequestKind.ID;
+}
+
+/**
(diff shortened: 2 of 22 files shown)


libs/category/driver/magento/src/category.service.ts
```diff
@@ -63,9 +64,9 @@ export class DaffMagentoCategoryService implements DaffCategoryServiceInterface
   /**
```
lderrickable, line 64: Looks like the todo on line 63 can be removed

libs/category/src/models/category-request.ts
```diff
@@ -3,15 +3,63 @@ import { DaffSortDirectionEnum } from '@daffodil/core/state';
...
+   */
   filter_requests?: DaffCategoryFilterRequest[];
+
+  /**
+   * The option by which to sort a categories' items.
+   */
   applied_sort_option?: string;
+
+  /**
+   * The sort direction to sort the categories' items by.
```
lderrickable, line 31: This feels like a test

libs/category/src/models/category-request.ts
```diff
@@ -3,15 +3,63 @@ import { DaffSortDirectionEnum } from '@daffodil/core/state';
 
 import { DaffCategoryFilterRequest } from './filters/public_api';
 
+/**
+ * When you request a category page, there's two ways to do so, via URI or via ID.
+ */
+export enum DaffCategoryPageRequestKind {
```
lderrickable, line 9: Should this be `DaffCategoryRequestKind`?

libs/category/state/src/effects/category-page.effects.ts
```diff
@@ -72,47 +73,44 @@ export class DaffCategoryPageEffects<
   @Effect()
   changeCategoryPageSize$: Observable<any> = this.actions$.pipe(
     ofType(DaffCategoryPageActionTypes.CategoryPageChangeSizeAction),
-    withLatestFrom(
+    withLatestFrom<DaffCategoryPageChangePageSize, Observable<DaffCategoryPageMetadata>>(
```
lderrickable, line 76: why do you prefer casting here rather than typing the input args of the switchMap? I usually prefer typing args to casting. Seems neater

libs/category/src/models/category-request.ts
lderrickable: I also wonder if we actually need this enum at all. Because we can tell if a request is and id request if it has an ID and we can tell if it is a Url request if it has a URL. By adding `kind` as a required field, we make it so the user needs to add a `kind: DaffCategoryRequestKind.ID` to every action dispatch.

lderrickable (changes requested):

libs/category/state/src/effects/category-page.effects.ts
griest024: I thought it was needed for the build but I think it was just vs code being dumb, changed

libs/category/src/models/category-request.ts
griest024: > Should this be DaffCategoryRequestKind?

I would be fine with that change.

> we can tell if a request is and id request if it has an ID and we can tell if it is a Url request if it has a URL

I prefer having an enum on which we can switchcase. Most of the users will probably end up using the resolvers anyway.

libs/category/src/models/category-request.ts
griest024: changed

libs/category/driver/magento/src/category.service.ts
griest024: removed

libs/category/state/src/reducers/category/pure/build-metadata-from-request.ts
```diff
@@ -1,9 +1,9 @@
...
   DaffCategoryPageMetadata,
-  DaffCategoryRequest,
+  DaffCategoryIdRequest,
 } from '@daffodil/category';
 
-type MetadataRequestIntersection = keyof DaffCategoryPageMetadata & keyof DaffCategoryRequest;
+type MetadataRequestIntersection = keyof DaffCategoryPageMetadata & keyof DaffCategoryIdRequest;
 
 const allowedMetadataKeys: MetadataRequestIntersection[] = [
   'id',
```
damienwebdev, line 9: In retrospect, is `id` necessary?

libs/category/state/src/reducers/category/pure/build-metadata-from-request.ts
```diff
@@ -13,7 +13,7 @@ const allowedMetadataKeys: MetadataRequestIntersection[] = [
   'page_size',
 ];
 
-export const buildMetadataFromRequest = (request: DaffCategoryRequest): Partial<DaffCategoryPageMetadata> =>
+export const buildMetadataFromRequest = (request: DaffCategoryIdRequest): Partial<DaffCategoryPageMetadata> =>
```
damienwebdev, line 16: Why is this an `DaffCategoryIdRequest`?

libs/category/state/src/effects/category-page.effects.ts
```diff
@@ -75,10 +76,9 @@ export class DaffCategoryPageEffects<
     withLatestFrom(
       this.store.pipe(select(this.categorySelectors.selectCategoryPageMetadata)),
     ),
-    switchMap((
-      [action, categoryRequest]: [DaffCategoryPageChangePageSize, DaffCategoryRequest],
-    ) => this.processCategoryGetRequest({
+    switchMap(([action, categoryRequest]: [DaffCategoryPageChangePageSize, DaffCategoryPageMetadata]) => this.processCategoryGetRequest({
       ...categoryRequest,
+      kind: DaffCategoryPageRequestKind.ID,
```
damienwebdev, line 81: Why is the `kind` being spread in here shouldn't this be on the request?

damienwebdev (changes requested):

libs/category/src/models/category-request.ts
lderrickable: Ok, I think we should go with `DaffCategoryRequestKind`, since it will be used for both category and category page actions

libs/category/state/src/effects/category-page.effects.ts
griest024: that's the metadata being spread, not actually the request. I improving its name though

libs/category/state/src/reducers/category/pure/build-metadata-from-request.ts
griest024: See above comment

libs/category/state/src/reducers/category/pure/build-metadata-from-request.ts
griest024: Probably not, and that will allow the function to be typed on the request union.

griest024: changed

libs/category/src/models/category-request.ts
griest024: changed

libs/category/state/src/reducers/category/category.reducer.ts
```diff
@@ -54,6 +54,7 @@ export function daffCategoryReducer<U extends DaffGenericCategory<U>, W extends
         categoryPageMetadata: {
           ...initialState.categoryPageMetadata,
           ...buildMetadataFromRequest(action.request),
+          id: action.request.id,
```
damienwebdev, line 57: Should we even set this upon load, or only upon load success?

damienwebdev (changes requested):

libs/category/state/src/reducers/category/category.reducer.ts
griest024: How much of the requests do we want to optimistically apply to state? I thought everything but maybe it should only be the filters.

griest024: bump

damienwebdev: In retrospect, I'm not sure it matters.

lderrickable (approved):

damienwebdev (approved):

PR #1461, opened 2021-04-22
feat(category): add getByUri driver call

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/category/driver/magento/src/category.service.spec.ts
+++ b/libs/category/driver/magento/src/category.service.spec.ts
@@ -1,4 +1,9 @@
-import { TestBed } from '@angular/core/testing';
+import {
+  TestBed,
+  fakeAsync,
+  tick,
+  flush,
+} from '@angular/core/testing';
 import {
   ApolloTestingModule,
   ApolloTestingController,
@@ -14,6 +19,7 @@ import {
   DaffGetCategoryResponse,
   DaffCategoryRequestKind,
   DaffCategoryFilterRangeNumericRequest,
+  DaffCategoryUriRequest,
 } from '@daffodil/category';
 import {
   MagentoGetACategoryResponse,
@@ -183,7 +189,7 @@ describe('Driver | Magento | Category | CategoryService', () => {
     expect(categoryService).toBeTruthy();
   });
 
-  describe('get | getting a category', () => {
+  describe('get | getting a category by ID', () => {
     let result: Observable<DaffGetCategoryResponse>;
 
     beforeEach(() => {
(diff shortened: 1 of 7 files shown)


libs/category/driver/magento/src/category.service.spec.ts
```diff
@@ -250,6 +256,123 @@ describe('Driver | Magento | Category | CategoryService', () => {
...
+
+      const productsOp = controller.expectOne(MagentoGetProductsQuery);
+      productsOp.flushData(mockGetProductsResponse);
+
+      expect(productsOp.operation.variables.filter.category_id.eq).toEqual(String(mockMagentoCategory.id));
+
+      flush();
+    }));
+
+    describe('when filters are requested', () => {
```
lderrickable, line 330: I think this should read "when the request applied a filter" or something like that. Filters are always requested/retrieved from the backend.

lderrickable (changes requested):

griest024:
Blocked by #1462

libs/category/driver/magento/src/category.service.spec.ts
griest024: changed

lderrickable (dismissed):

libs/category/driver/magento/src/category.service.ts
```diff
@@ -91,6 +89,38 @@ export class DaffMagentoCategoryService implements DaffCategoryServiceInterface
     );
   }
 
+  getByUri(categoryRequest: DaffCategoryUriRequest): Observable<DaffGetCategoryResponse> {
+    const truncatedUri = categoryRequest.uri.replace('.html', '');
```
damienwebdev, line 93: I don't think we should be truncating the URI.

damienwebdev (changes requested):

libs/category/driver/magento/src/category.service.ts
griest024: I have a PR out for appending the suffix to the URL, which I think it required? Or do you think that we should have the URL suffix in the daff urls at all?

lderrickable: Why would magento expect a url without the suffix, but provide the suffix in the category response? Would it be better for daffodil to only deal with whatever format the platform expects in the uri query? It seems like with this PR and the suffix PR the end result is just us putting the suffix onto the uri only to remove it just before the request is made. Do we need the suffix for some reason?

griest024: yeah I was under the impression that there was a compelling reason to need the suffix app-side. There is some consequence for routing behavior but I don't remember what it is. @damienwebdev do you know?

damienwebdev: In retrospect, it's more correct to say that this truncating is configurable by the app-dev. It's not a safe assumption that the suffix is `.html`.

griest024: changed truncation behavior to be dynamic based on config option

damienwebdev (approved):
lgtm

PR #1472, opened 2021-04-27
feat(product): add uri resolver

### LINKED ISSUE #1471: add product uri resolver

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
Fixes #1471

### DIFF
--- a/libs/product/state/src/resolvers/product-page-uri/product-page-uri.resolver.ts
+++ b/libs/product/state/src/resolvers/product-page-uri/product-page-uri.resolver.ts
@@ -1,0 +1,55 @@
+import { isPlatformBrowser } from '@angular/common';
+import {
+  Inject,
+  Injectable,
+  PLATFORM_ID,
+} from '@angular/core';
+import {
+  ActivatedRouteSnapshot,
+  Resolve,
+} from '@angular/router';
+import { ofType } from '@ngrx/effects';
+import {
+  ActionsSubject,
+  Store,
+} from '@ngrx/store';
+import {
+  Observable,
+  of,
+} from 'rxjs';
+import {
+  mapTo,
+  take,
+} from 'rxjs/operators';
+
+import {
+  DaffProductPageLoadByUrl,
+  DaffProductPageActionTypes,
+} from '../../actions/public_api';
+import { DaffProductReducersState } from '../../reducers/public_api';
+
+/**
+ * Resolves product data for product pages, and will only resolve the url
+ * after a product request succeeds or fails. This resolver will take a full
+ * a url of the form `some/url.html` and attempt to resolve a product from it.
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class DaffProductPageUriResolver implements Resolve<Observable<boolean>> {
+  constructor(
+    @Inject(PLATFORM_ID) private platformId: string,
+    private store: Store<DaffProductReducersState>,
+    private dispatcher: ActionsSubject,
+  ) { }
+
+  resolve(route: ActivatedRouteSnapshot): Observable<boolean> {
+    this.store.dispatch(new DaffProductPageLoadByUrl(route.toString()));
+
+    return isPlatformBrowser(this.platformId) ? of(true) : this.dispatcher.pipe(
+      ofType(DaffProductPageActionTypes.ProductPageLoadSuccessAction, DaffProductPageActionTypes.ProductPageLoadFailureAction),
+      mapTo(true),
+      take(1),
+    );
+  }
+}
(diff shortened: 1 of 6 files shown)


libs/product/state/src/resolvers/product-page/product-page.resolver.ts
```diff
@@ -37,18 +16,10 @@ import { DaffProductReducersState } from '../../reducers/public_api';
 })
 export class DaffProductPageResolver implements Resolve<Observable<boolean>> {
```
lderrickable, line 17: Should this be deprecated?

griest024: yes, good catch

griest024: changed

lderrickable:
How exactly is this intended to be used? Because the app dev won't know what kind of url the user will give them (either domain/product/id or domain/product/product-name.html). I would expect that a common resolver would be needed to determine what kind of url is used, then that resolver could call productIdResolver.resolve or productUrlResolver.resolve. Is there another plan to handle this?

griest024:
> a common resolver would be needed to determine what kind of url is used, then that resolver could call productIdResolver.resolve or productUrlResolver.resolve

That's exactly how it works: https://github.com/graycoreio/daffodil/blob/develop/libs/external-router/routing/src/guard/existence.guard.ts

griest024:
Although the guard doesn't directly call the router, it adds a route to a specific place in the config and then that presumably is sufficient to invoke the correct resolver.

griest024:
@lderrickable hang on I'll write a guide. I should have already done this TBH.

griest024:
@lderrickable #1474

lderrickable:
@griest024 With the setup outlined in the docs, how would it handle this route: `domain/product:id`? Seems like it would get to the `DaffProductPageUriResolver` and always result in a failed request.

Maybe something like this?
```
@NgModule({
  imports: [
    DaffExternalRouterModule.forRoot({
      failedResolutionPath: 'not-found',
    }, [
      {
        type: 'product:id',
        route: {
          component: MyProductPageComponent,
          ..., // additional route config, but without path
          resolve: {
            product: DaffProductPageIdResolver
          }
        }
      },
      {
        type: 'product/**',
        route: {
          component: MyProductPageComponent,
          ..., // additional route config, but without path
          resolve: {
            product: DaffProductPageUriResolver
          }
        }
      }
    ]),
  ],
})
class AppModule {}
```

lderrickable:
actually, I'm not sure that would work, since the existence guard/resolver service would make the "resolvable route" a known route and therefore it wouldn't be caught by the `product/**`

griest024:
@lderrickable I'm not sure exactly what your question is. If you're referring to the use case of putting routes in nested trees, that is supported by `insertionStrategy`. I added some better docs about that to my PR.

Note that `type` in the config is an enum defined by the platform and is returned in the driver reponse. For example Magento only ever has [these three](https://github.com/graycoreio/daffodil/blob/develop/libs/external-router/driver/magento/src/model/resolution-types.ts), You seem to be using it like its a route path but its not.

lderrickable (dismissed):

lderrickable:
lgtm

libs/product/state/src/resolvers/product-page-uri/product-page-uri.resolver.ts
```diff
@@ -0,0 +1,55 @@
...
+})
+export class DaffProductPageUriResolver implements Resolve<Observable<boolean>> {
+  constructor(
+    @Inject(PLATFORM_ID) private platformId: string,
+    private store: Store<DaffProductReducersState>,
+    private dispatcher: ActionsSubject,
+  ) { }
+
+  resolve(route: ActivatedRouteSnapshot): Observable<boolean> {
+    this.store.dispatch(new DaffProductPageLoadByUrl(route.toString()));
```
damienwebdev, line 47: Seeing that you mock `toString` in the test, do you know that `toString` returns what you expect? e.g. the path after the domain.

damienwebdev (changes requested):

libs/product/state/src/resolvers/product-page-uri/product-page-uri.resolver.ts
griest024: fixed

damienwebdev (approved):

PR #1493, opened 2021-05-04
feat(core): handle missing leading slash in query fragment truncation

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

query param truncation fails if the uri does not have a leading slash

## What is the new behavior?

query param truncation works as expected if the uri does not have a leading slash

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/core/routing/src/uri/truncation/query-fragment.ts
+++ b/libs/core/routing/src/uri/truncation/query-fragment.ts
@@ -1,7 +1,6 @@
 /**
  * Truncates query parameters and fragments from a URI.
  * e.g. /foo/bar/baz.html?thing=test#id -> /foo/bar/baz.html
- * Note that this will not work as expected unless `uri` has a leading slash.
  */
 export const daffUriTruncateQueryFragment = (uri: string): string =>
-  (new URL(`http://fakedomain.com${uri}`)).pathname;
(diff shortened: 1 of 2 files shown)


libs/core/routing/src/uri/truncation/query-fragment.ts
```diff
@@ -1,6 +1,7 @@
 /**
  * Truncates query parameters and fragments from a URI.
  * e.g. /foo/bar/baz.html?thing=test#id -> /foo/bar/baz.html
+ * Note that this will not work as expected unless `uri` has a leading slash.
  */
 export const daffUriTruncateQueryFragment = (uri: string): string =>
```
damienwebdev, line 6: Rather than add a comment, can we handle the case by optionally adding a slash if the first char isn't a slash?

damienwebdev (changes requested):

libs/core/routing/src/uri/truncation/query-fragment.ts
griest024: yeah I think its better that way, changed

damienwebdev (approved):
lgtm

PR #1495, opened 2021-05-04
feat(external-router): truncate driver request and response

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

External router driver do not truncate the request or response

## What is the new behavior?
- Magento drivers truncate the query parameters and fragments from the request
- Magento drivers truncate the leading slash from the response
- The testing driver truncates query params, fragments, and the leading slash from the request URL before performing the lookup in its config. If the lookup succeeds, the truncated URL is used for the response.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/external-router/driver/magento/2.4.1/src/magento.service.spec.ts
+++ b/libs/external-router/driver/magento/2.4.1/src/magento.service.spec.ts
@@ -20,7 +20,8 @@ describe('@daffodil/external-router/driver/magento/2.4.1 | DaffExternalRouterMag
   let service: DaffExternalRouterMagentoDriver;
   let controller: ApolloTestingController;
   let scheduler: TestScheduler;
-  let url: string;
+  let responseUrl: string;
+  let requestUrl: string;
   let id: ID;
   let resolution: MagentoUrlResolverResponse;
   let resolvableUrl: DaffExternallyResolvableUrl;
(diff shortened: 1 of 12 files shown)


lderrickable:
how do we know for sure that there won't be any fragments or query parameters in the resolvable url? Should this just truncate urls in the same way that product and category do? I.e. by a passed in function via injection token.

griest024:
> how do we know for sure that there won't be any fragments or query parameters in the resolvable url

In the request or response URL?

lderrickable:
request

griest024:
I do truncate the query params and fragments from the request URI though. Do you mean what if the query params and fragments are significant and we should allow app devs to pass that info to Magento if they want?

lderrickable:
> Do you mean what if the query params and fragments are significant and we should allow app devs to pass that info to Magento if they want?

yea, that one

griest024:
yeah I guess I can add a config for the external router magento driver. @damienwebdev does this make sense to you?

griest024:
@lderrickable after some discussion we have decided to remove truncate on the request URL for Magento 2.4.2. This is because 2.4.2 can handle query params and fragments where 2.4.1 would crash. It doesn't make sense to give 2.4.1 configurable truncation since the requirements are pretty narrow.

lderrickable:
lgtm

lderrickable (approved):

damienwebdev (approved):
lgtm

PR #1514, opened 2021-05-10
feat(category): use locatable interface and change URI to URL

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
- uses `DaffLocatable`
- change `URI` to `URL` to reduce confusion

```
[x] Yes
[ ] No
```

changes all occurences of `URI` to `URL`

## Other information


lderrickable:
I see what you mean, but I'm still leaning toward uri. url isn't correct, while uri is vague. But the vagueness should be made up for with context clues. If a model has an id and a uri, then I don't think it is unreasonable to assume that they are supposed to represent different entities, because if the uri was just the id we wouldn't bother having a uri field. I am pretty torn though. @damienwebdev deciding vote?

damienwebdev:
I suspect that `url` is as correct (and user-friendly) as we can get. 

Comparatively:

* Magento struggles with this (calls it `url_key`)
* Shopify struggles with this (calls it `handle`). 

While it's technically incorrect to call it `URL`, I think in the context of what we (and users) will try to be using it for; providing a link to the page that renders the entity:

```html
<a [routerLink]="category.url">Some Link</a>

```  
I think it is safe for us to call this `url`.

As an aside, it's really more of an `href` -- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#href. The appropriate language here is likely that this is a `relative_url`, but since it is up to the client to define what the absolute URL is and Daffodil doesn't have any other concept of additional URL for entities, I think `url` is safe.

Issue #1499, opened 2021-05-05
[FEAT] Handle Redirects in External Router Existence Guard


griest024:
use response token: `import { RESPONSE } from '@nguniversal/express-engine/tokens'` to set HTTP code to `301` and set location header.

PR #1516, opened 2021-05-10
feat(external-router): process redirects and errors

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, redirects cause an infinite loop.

Fixes: #1499 

## What is the new behavior?
We no longer get stuck in an infinite loop.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/external-router/driver/src/public_api.ts
+++ b/libs/external-router/driver/src/public_api.ts
@@ -6,4 +6,4 @@ export {
   DaffExternalRouterDriverInterface,
   DaffExternalRouterDriver,
 } from './interfaces/external-router-driver.interface';
-export { NOT_FOUND_RESOLUTION } from './not-found-resolution';
+export { DAFF_EXTERNAL_ROUTER_NOT_FOUND_RESOLUTION } from './not-found-resolution';
--- a/libs/external-router/driver/src/public_api.ts
+++ b/libs/external-router/driver/src/public_api.ts
@@ -6,3 +6,4 @@ export {
   DaffExternalRouterDriverInterface,
   DaffExternalRouterDriver,
 } from './interfaces/external-router-driver.interface';
+export { NOT_FOUND_RESOLUTION } from './not-found-resolution';
--- a/libs/external-router/routing/src/errors/client-error.ts
+++ b/libs/external-router/routing/src/errors/client-error.ts
@@ -1,0 +1,16 @@
+import {
+  DaffError,
+  DaffInheritableError,
+} from '@daffodil/core';
+
+/**
+ * An error thrown when the driver encounters a 4xx error.
+ */
+export class DaffExternalRouterClientError extends DaffInheritableError
+  implements DaffError {
+	public readonly code: string = '4xx_CLIENT_ERROR';
+
+	constructor(message?: string) {
+	  super(message);
+	}
+}
--- a/libs/external-router/routing/src/guard/existence.guard.ts
+++ b/libs/external-router/routing/src/guard/existence.guard.ts
@@ -1,6 +1,8 @@
+import { isPlatformServer } from '@angular/common';
 import {
   Injectable,
   Inject,
+  PLATFORM_ID,
 } from '@angular/core';
 import {
   CanActivate,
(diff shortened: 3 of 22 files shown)


libs/external-router/driver/magento/src/processors/process-redirect.ts
```diff
@@ -0,0 +1,22 @@
...
+  DaffExternalRouterTemporaryRedirectError,
+} from '@daffodil/external-router/driver';
+
+import { MagentoUrlResolverResponse } from '../model/public_api';
+
+/**
+ * Processes Magento responses and throws the appropriate error so that External Router
+ * can handle redirects.
+ */
+export const processRedirectResponse = (response: ApolloQueryResult<MagentoUrlResolverResponse>) => {
```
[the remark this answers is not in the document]
griest024: I agree but I prefer `magentoProcessRedirectResponse` to stay consistent

damienwebdev:
@lderrickable @griest024 I've reworked this fairly significantly, please take another look. I like this version much better. It allows the drivers to act like regular HttpResponse results and defers managing those results to the guard.

libs/external-router/routing/src/guard/existence.guard.ts
```diff
@@ -54,12 +63,40 @@ export class DaffExternalRouterExistenceGuard implements CanActivate {
...
+      // Handle redirects
       catchError(error => {
+        if(
+          !(error instanceof DaffExternalRouterPermanentRedirectError) &&
+          !(error instanceof DaffExternalRouterTemporaryRedirectError)) {
+          throw error;
+        }
+
+        //On platform browser, we need to add a redirect for the original url and redirect to the new url.
+        this.externalRouter.addRedirect(state.url, error.redirectUrl);
```
damienwebdev, line 85: `state.url` may not be safe to insert here as it contains queries and hashes.

libs/external-router/routing/src/guard/existence.guard.ts
```diff
@@ -54,12 +63,40 @@ export class DaffExternalRouterExistenceGuard implements CanActivate {
...
+        return of(this.router.parseUrl(error.redirectUrl));
+      }),
+      // Handle "404"
+      catchError((error) => {
+        if(!(error instanceof DaffExternalRouterNotFoundError)) {
+          throw error;
+        }
+
+        this.externalRouter.addRedirect(state.url, this.config.failedResolutionPath);
```
damienwebdev, line 94: `state.url` may not be safe to insert here as it contains queries and hashes.

libs/external-router/routing/src/guard/existence.guard.ts
griest024: yeah because that means that the path will only match with the query strings, right?

griest024 (changes requested):

libs/external-router/routing/src/errors/client-error.ts
```diff
@@ -0,0 +1,17 @@
+import {
+  DaffError,
+  DaffInheritableError,
+} from '@daffodil/core';
+
+/**
+ * An error thrown when the driver needs a URI to be redirected permanently
+ * to another location.
```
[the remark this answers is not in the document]
damienwebdev: Fixed in abe8b76cc3adb9b8ac0ef4e98dff25987494542e

libs/external-router/routing/src/guard/existence.guard.ts
damienwebdev: Fixed in abe8b76cc3adb9b8ac0ef4e98dff25987494542e -- @griest024 can you give it a once over to verify.

libs/external-router/driver/src/public_api.ts
```diff
@@ -6,3 +6,4 @@ export {
   DaffExternalRouterDriverInterface,
   DaffExternalRouterDriver,
 } from './interfaces/external-router-driver.interface';
+export { NOT_FOUND_RESOLUTION } from './not-found-resolution';
```
griest024, line 9: this should have a daff or magento prefix since its exported

griest024 (changes requested):

libs/external-router/driver/src/public_api.ts
damienwebdev: 28c849b8

griest024 (approved):

Issue #1537, opened 2021-05-14
[BUG] Magento Add Bundle to Cart on v2.4.2 doesn't work


damienwebdev:
# :bug: Bug report

## Current Behavior
When I try to add a bundled item to my cart using product_uid's I get the error:
```
GraphQL error]: Message: Please specify product option(s)., Location: undefined, Path: undefined
```

## Expected Behavior
I don't expect to get an error.

## Minimal reproduction of the problem with instructions
This is happening because we changed to using `product.uid` instead of `product.id` and the old `addBundledProductsToCart` doesn't handle `uid`. We need to use[ `addProductsToCart `](https://devdocs.magento.com/guides/v2.4/graphql/mutations/add-products-to-cart.html)

## What is the motivation / use case for changing the behavior?
I'd like to be able to add bundle items to my cart.

## Environment

<pre><code>
Daffodil version: 0.32.0

griest024:
I'll take this one

PR #1540, opened 2021-05-14
fix(cart): cannot add composite products to cart

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #1537

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/cart/driver/magento/src/cart-item.service.ts
+++ b/libs/cart/driver/magento/src/cart-item.service.ts
@@ -20,20 +20,14 @@ import { DaffQueuedApollo } from '@daffodil/core/graphql';
 
 import { DAFF_MAGENTO_CART_MUTATION_QUEUE } from './injection-tokens/cart-mutation-queue.token';
 import { DAFF_CART_MAGENTO_EXTRA_CART_FRAGMENTS } from './injection-tokens/public_api';
-import { MagentoConfigurableCartItemInput } from './models/requests/cart-item';
+import { MagentoCartItemInput } from './models/requests/cart-item';
+import { addCartItem } from './queries/add-cart-item';
 import {
   listCartItems,
-  addConfigurableCartItem,
-  addBundleCartItem,
-  addSimpleCartItem,
   removeCartItem,
   updateCartItem,
 } from './queries/public_api';
-import {
-  MagentoAddSimpleCartItemResponse,
-  MagentoAddBundleCartItemResponse,
-  MagentoAddConfigurableCartItemResponse,
-} from './queries/responses/add-cart-item';
+import { MagentoAddCartItemResponse } from './queries/responses/add-cart-item';
 import { MagentoListCartItemsResponse } from './queries/responses/list-cart-items';
 import { MagentoUpdateCartItemResponse } from './queries/responses/public_api';
 import { MagentoRemoveCartItemResponse } from './queries/responses/remove-cart-item';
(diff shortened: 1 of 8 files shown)


libs/cart/driver/magento/src/cart-item.service.ts
```diff
@@ -115,42 +109,26 @@ export class DaffMagentoCartItemService implements DaffCartItemServiceInterface
...
-    }).pipe(
-      map(result => this.cartTransformer.transform(result.data.addConfigurableProductsToCart.cart)),
-    );
+    return this.addCartItemRequest(cartId, transformConfigurableCartItem(cartItemInput));
   }
 
   private addSimpleProduct(cartId: DaffCart['id'], cartItemInput: DaffCartItemInput): Observable<Partial<DaffCart>> {
-    return this.mutationQueue.mutate<MagentoAddSimpleCartItemResponse>({
-      mutation: addSimpleCartItem(this.extraCartFragments),
+    return this.addCartItemRequest(cartId, transformSimpleCartItem(cartItemInput));
```
lderrickable, line 120: The magento docs for (https://devdocs.magento.com/guides/v2.4/graphql/mutations/add-products-to-cart.html#input-attributes) say that all of the cart item inputs are required, which I admit sounds weird. But if it is the case, then adding a simple product to the cart like this will fail I think. Either `transformSimpleCartItem` needs to return something for all of the input fields or default values need to be added for the graphQl queries somehow.

lderrickable (changes requested):

libs/cart/driver/magento/src/cart-item.service.ts
griest024: I think the docs say only `sku` and `quantity` are required. In any case the following input works through postman:
```json
"input": {
    "quantity": 1,
    "sku": "Tee-Brush-Spring Blister"
}
```

griest024: Note that `[ID!]` is different from `[ID!]!`

damienwebdev (changes requested):

libs/cart/driver/magento/src/models/requests/cart-item.ts
```diff
@@ -1,20 +1,7 @@
...
-	quantity: number;
-	value: string[];
-}
-
-export interface MagentoConfigurableCartItemInput {
-	parentSku: string;
-	data: MagentoCartItemInput;
+  parent_sku?: string;
+  selected_options?: string[];
+  entereded_options?: string[];
```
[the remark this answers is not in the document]
griest024: no, fixed

libs/cart/driver/magento/src/models/requests/cart-item.ts
```diff
@@ -1,20 +1,7 @@
...
-export interface MagentoBundledCartItemOption {
-	id: number;
-	quantity: number;
-	value: string[];
-}
-
-export interface MagentoConfigurableCartItemInput {
-	parentSku: string;
-	data: MagentoCartItemInput;
+  parent_sku?: string;
```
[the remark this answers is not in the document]
griest024: I think we should match the platform models as closely as possible and its one type with optional in Magento. So no, I think it should be one model

libs/cart/driver/magento/src/cart-item.service.ts
lderrickable: I guess I took "The CartItemInput object must contain the following attributes:" too literally.

lderrickable (approved):

damienwebdev (approved):

PR #1555, opened 2021-05-21
feat(design): add nav-accordion-item component

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
A new `daff-nav-accordion-item` component was added for use in navigation contexts with nested accordion-items.

The plan is to use this for daff.io docs

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/design-land/src/app/accordion/accordion.component.html
+++ b/apps/design-land/src/app/accordion/accordion.component.html
@@ -1,31 +1,60 @@
-<h1>Daff Accordion</h1>
+<div class="design-land-accordion-example">
+	<h1>Daff Accordion</h1>
 
-<daff-accordion>
-  <daff-accordion-item initiallyActive="true">
-    <h3 daffAccordionItemTitle>Details</h3>
-    <div daffAccordionItemContent>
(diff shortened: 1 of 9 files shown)


xelaint:
I think `nested` may be a better property name for what this is trying to accomplish?

xelaint (changes requested):
Goodwill Industries of Central Florida

lderrickable:
I think I like `indent` better, since it directly describes what it's doing whereas `nested` does not. With `nested` we're claiming that we know exactly what everyone expects when an accordion item is nested. With `indent` it's very clear that the intention is. If you feel strongly about `nested` I'll change it though.

lderrickable:
On second thought, maybe this is an imperative vs declarative thing. Could be that `nested` is better because it is declarative. I'm not sure

lderrickable:
Ok, I just spoke to Damien about imperative vs declarative, and I do think that `indent` is probably the right call here for the reasons I previously stated.

damienwebdev (approved):
lgtm

PR #1568, opened 2021-06-02
feat(tools-dgeni): add processor for the @inheritdoc tag

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The jsdoc `@inheritdoc` tag doesn't do anything right now.

## What is the new behavior?
The jsdoc `@inheritdoc` tag now causes dgeni to copy member descriptions from the interface to the class that implements it.

```
[ ] Yes
[x] No
```

## Other information
While there is a dgeni-packages for jsdoc, there is no processor in it for the `@inheritdoc` tag. See the list of supported tags here: https://github.com/angular/dgeni-packages/tree/master/jsdoc/tag-defs
I also tried using it anyway just to see if it would process them secretly, but it doesn't. So I wrote a custom processor that looks for the `@inheritdoc` tag. An example of this working is on the daff.io app with the `DaffCategoryFacade`.

### DIFF
--- a/tools/dgeni/src/processors/addInheritedDocsContent.ts
+++ b/tools/dgeni/src/processors/addInheritedDocsContent.ts
@@ -1,0 +1,26 @@
+import { Processor, Document } from 'dgeni';
+
+/**
+ * Inherit docs content from parent interfaces.
+ */
+export class AddInheritedDocsContentProcessor implements Processor {
+	name = 'filterOutPrivateProperties';
+	$runAfter = ['docs-processed'];
+	$runBefore = ['rendering-docs'];
+
+	$process(docs: Document[]): Document[] {
+		return docs.map(doc => {
+			if(!doc.members || !doc.tags.tags.filter(tag => tag.tagName === 'inheritdoc').length) return doc;
+
+			//get all interfaces that the doc implements.
+			doc.moduleDoc.exports.filter(e => e.symbol.escapedName.includes('Interface'))
+				.map(i => {
+					i.members.map(member => {
+						//copy over the description from the interface to the doc.
+						doc.members.find(m => m.name === member.name).description = member.description;
+					})
+				})
+			return doc;
+		});
+	}
+}
--- a/tools/dgeni/src/processors/addInheritedDocsContent.ts
+++ b/tools/dgeni/src/processors/addInheritedDocsContent.ts
@@ -13,11 +13,12 @@ export class AddInheritedDocsContentProcessor implements Processor {
 			if(!doc.members || !doc.tags.tags.filter(tag => tag.tagName === 'inheritdoc').length) return doc;
 
 			//get all interfaces that the doc implements.
-			doc.moduleDoc.exports.filter(e => e.symbol.escapedName.includes('Interface'))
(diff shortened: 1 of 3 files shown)


tools/dgeni/src/processors/addInheritedDocsContent.ts
```diff
@@ -0,0 +1,26 @@
...
+	name = 'filterOutPrivateProperties';
+	$runAfter = ['docs-processed'];
+	$runBefore = ['rendering-docs'];
+
+	$process(docs: Document[]): Document[] {
+		return docs.map(doc => {
+			if(!doc.members || !doc.tags.tags.filter(tag => tag.tagName === 'inheritdoc').length) return doc;
+
+			//get all interfaces that the doc implements.
+			doc.moduleDoc.exports.filter(e => e.symbol.escapedName.includes('Interface'))
```
damienwebdev, line 16: Why `.includes` instead of a strict equality check?

tools/dgeni/src/transforms/daffodil-api-package/index.ts
```diff
@@ -24,6 +25,7 @@ export const apiDocs =  new Package('checkout', [
   .processor(new CleanSelectorsProcessor())
   .processor(new MakeTypesHtmlCompatibleProcessor())
   .processor(new FilterOutPrivatePropertiesProcessor())
+  .processor(new AddInheritedDocsContentProcessor())
```
damienwebdev, line 28: Can this be done with the existing processor from dgeni? https://github.com/angular/dgeni-packages/commit/cf8b7444d527d5c97ee531d1dc71241413b3bdb7

damienwebdev (changes requested):

tools/dgeni/src/processors/addInheritedDocsContent.ts
lderrickable: That field would be something like "DaffCartFacadeInterface"

tools/dgeni/src/transforms/daffodil-api-package/index.ts
lderrickable: no. This PR is just for classes that implement interfaces and getting the docs for those interface members into the class. It looks like that existing processor is only for classes that extend each other. We should probably use that as well in the future though.

tools/dgeni/src/processors/addInheritedDocsContent.ts
damienwebdev: Is there a prop on a dgeni doc that gives the symbols type which could be compared against? 

While today we use `*Interface` as the convention, I can imagine various scenarios where a different convention gets used. 

It would be nice to allow for such an edge-case.

lderrickable: I'll take a look

lderrickable: changed

damienwebdev (approved):

PR #1577, opened 2021-06-03
docs(external-router): add @inheritdoc tag to classes that should inh…

lderrickable:
…erit docs from their interfaces

## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Adds `@inheritdoc` tag to classes that should inherit docs from the interfaces they implement.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/external-router/routing/src/errors/client-error.ts
+++ b/libs/external-router/routing/src/errors/client-error.ts
@@ -5,6 +5,8 @@ import {
 
 /**
  * An error thrown when the driver encounters a 4xx error.
+ *
+ * @inheritdoc
  */
 export class DaffExternalRouterClientError extends DaffInheritableError
   implements DaffError {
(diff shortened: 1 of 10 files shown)


libs/external-router/routing/src/errors/client-error.ts
```diff
@@ -5,6 +5,8 @@ import {
 
 /**
  * An error thrown when the driver encounters a 4xx error.
+ *
+ * @inheritdoc
```
griest024, line 9: I don't think the errors need to inherit docs, @damienwebdev do you agree?

lderrickable: I think I'm just going to remove it, because right now the `@inheritdoc` will only work if the fields in the interface also exist in the class, which isn't the case with all of these errors. They ultimately extend from `Error`

griest024 (approved):

PR #1584, opened 2021-06-04
fix(tools-dgeni): AddInheritedDocsContentProcessor is overriding the …

lderrickable:
…FilterOutPrivatePropertiesProcessor

## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

TLDR: There were a number of bugs with the `AddInheritedDocsContentProcessor`. This PR makes the inheritdoc tag usage a lot more robust by requiring the names of the interfaces from which you want the class to inherit.

1. The AddInheritedDocsContentProcessor has the wrong name and it is overriding the FilterOutPrivatePropertiesProcessor.
2. Dgeni does not include interfaces in generated documentation objects for classes in which they are implemented if those interfaces are defined before the classes.
3. Dgeni includes too many interfaces as related to classes for some reason; e.g. the DaffProductFacade has every type of product facade interface associated with it (`DaffConfigurableProductFacadeInterface`, `DaffCompositeProductFacadeInterface`, etc). This results in shared fields (like `loading$`) being overwritten by the last interface when docs are generated. This must be resolved by either finding a better list of implemented interfaces within the dgeni-docs generated Doc object (which I couldn't find), or by searching for each interface name specifically (which makes sense because I needed to address problem #2 anyway).

```
[ ] Yes
[x] No
```

Notes:
The new inheritdoc tag usage is:
For a single interface: `@inheritdoc SomeInterfaceName`
For multiple interfaces: `@inheritdoc SomeInterfaceName, SomeOtherInterfaceName`

This dgeni processor will search for those interfaces in the global list of generated docs, then add member descriptions for any member matches that are found between the interface and class.

### DIFF
--- a/tools/dgeni/src/processors/addInheritedDocsContent.spec.ts
+++ b/tools/dgeni/src/processors/addInheritedDocsContent.spec.ts
@@ -15,20 +15,21 @@ describe("AddInheritedDocsContentProcessor", () => {
 				tags: {
 					tags: [{
 						tagName: 'inheritdoc',
-						description: 'SomeInterfaceName'
 					}]
 				},
 				members: [{
 					name: 'member1',
 					description: null,
-				}]
-			},
-			{
-				tags: { tags: [] },
-				name: 'SomeInterfaceName',
-				members: [{
-					name: 'member1',
-					description: stubInterfaceMemberDescription
+				}],
+				implementsClauses: [{
+					doc: {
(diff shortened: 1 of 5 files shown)


damienwebdev (approved):
lgtm

PR #1597, opened 2021-06-14
docs(category): add some api docs

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/category/src/models/category-breadcrumb.ts
+++ b/libs/category/src/models/category-breadcrumb.ts
@@ -7,7 +7,17 @@ import { DaffCategory } from './category';
  * describing the traversal to a particular category within a tree of categories.
  */
 export interface DaffCategoryBreadcrumb extends DaffLocatable {
+	/**
+	 * The id of the associated category.
+	 */
   id: DaffCategory['id'];
+	/**
+	 * A label to display the category name in the UI.
+	 */
   name: DaffCategory['name'];
+	/**
+	 * The number of nodes between this node and the root node for the category trie. For example, the level for a
+	 * "Men's Shirts" breadcrumb could be 2 if the breadcrumb trie looked like: "Men" => "Clothing" => "Men's Shirts".
+	 */
   level: number;
 }
--- a/libs/category/src/models/category-breadcrumb.ts
+++ b/libs/category/src/models/category-breadcrumb.ts
@@ -16,8 +16,8 @@ export interface DaffCategoryBreadcrumb extends DaffLocatable {
 	 */
   name: DaffCategory['name'];
 	/**
-	 * The number of nodes between this node and the root node for the category trie. For example, the level for a
-	 * "Men's Shirts" breadcrumb could be 2 if the breadcrumb trie looked like: "Men" => "Clothing" => "Men's Shirts".
+	 * The number of nodes between this node and the root node for the category tree. For example, the level for a
+	 * "Men's Shirts" breadcrumb could be 2 if the breadcrumb tree looked like: "Men" => "Clothing" => "Men's Shirts".
 	 */
   level: number;
 }
--- a/libs/category/src/models/generic-category.ts
+++ b/libs/category/src/models/generic-category.ts
@@ -11,7 +11,13 @@ import { DaffCategoryBreadcrumb } from './category-breadcrumb';
  * this type is a recursive generic.
  */
 export interface DaffGenericCategory<T extends DaffGenericCategory<T>> extends DaffLocatable, DaffIdentifiable {
+	/**
+	 * The name of the category that can be used as a UI label.
+	 */
 	name: string;
+	/**
+	 * A description of the category.
+	 */
 	description?: string;
 	/**
 	 * A title for the category for usage in the tab title of the webpage.
@@ -21,9 +27,24 @@ export interface DaffGenericCategory<T extends DaffGenericCategory<T>> extends D
 	 * An overview description of the category for search engine results.
 	 */
 	meta_description?: string;
+	/**
+	 * The number of child category nodes under this category.
+	 */
   children_count?: number;
+	/**
+	 * The number of products described by this category.
+	 */
   total_products?: number;
+	/**
+	 * The direct child categories of this category.
+	 */
   children?: T[];
+	/**
+	 * The ids for all of the products described by this category.
+	 */
   product_ids?: string[];
+	/**
+	 * The breadcrumb trail that leads from this category to its root category.
+	 */
   breadcrumbs?: DaffCategoryBreadcrumb[];
 }
--- a/libs/category/src/models/generic-category.ts
+++ b/libs/category/src/models/generic-category.ts
@@ -32,7 +32,7 @@ export interface DaffGenericCategory<T extends DaffGenericCategory<T>> extends D
 	 */
   children_count?: number;
 	/**
-	 * The number of products described by the category.
+	 * The number of products that belong to the category.
 	 */
   total_products?: number;
 	/**
@@ -40,7 +40,7 @@ export interface DaffGenericCategory<T extends DaffGenericCategory<T>> extends D
 	 */
   children?: T[];
 	/**
-	 * The ids for all of the products described by the category.
+	 * The ids for all of the products that belong to the category.
 	 */
   product_ids?: string[];
 	/**
--- a/libs/category/src/models/generic-category.ts
+++ b/libs/category/src/models/generic-category.ts
@@ -32,7 +32,7 @@ export interface DaffGenericCategory<T extends DaffGenericCategory<T>> extends D
 	 */
   children_count?: number;
 	/**
-	 * The number of products described by this category.
+	 * The number of products described by the category.
 	 */
   total_products?: number;
 	/**
@@ -40,7 +40,7 @@ export interface DaffGenericCategory<T extends DaffGenericCategory<T>> extends D
 	 */
   children?: T[];
 	/**
-	 * The ids for all of the products described by this category.
+	 * The ids for all of the products described by the category.
 	 */
   product_ids?: string[];
 	/**
--- a/libs/category/src/models/get-category-response.ts
+++ b/libs/category/src/models/get-category-response.ts
@@ -13,7 +13,16 @@ export interface DaffGetCategoryResponse<
 	V extends DaffGenericCategory<V> = DaffCategory,
 	W extends DaffProduct = DaffProduct
 > {
+	/**
+	 * The list of products described by the category.
+	 */
   products: W[];
+	/**
+	 * The category requested.
+	 */
   category: V;
+	/**
+	 * The properties of a category page.
+	 */
   categoryPageMetadata: DaffCategoryPageMetadata;
 }
--- a/libs/category/src/models/get-category-response.ts
+++ b/libs/category/src/models/get-category-response.ts
@@ -14,7 +14,7 @@ export interface DaffGetCategoryResponse<
 	W extends DaffProduct = DaffProduct
 > {
 	/**
-	 * The list of products described by the category.
+	 * The list of products described by the requested category.
 	 */
   products: W[];
 	/**
@@ -22,7 +22,7 @@ export interface DaffGetCategoryResponse<
 	 */
   category: V;
 	/**
-	 * The properties of a category page.
+	 * The properties of the requested category page.
 	 */
   categoryPageMetadata: DaffCategoryPageMetadata;
 }
--- a/libs/category/src/models/get-category-response.ts
+++ b/libs/category/src/models/get-category-response.ts
@@ -14,11 +14,11 @@ export interface DaffGetCategoryResponse<
 	W extends DaffProduct = DaffProduct
 > {
 	/**
-	 * The list of products described by the requested category.
+	 * The list of products that belong to the requested category.
 	 */
   products: W[];
 	/**
-	 * The category requested.
+	 * The requested category.
 	 */
   category: V;
 	/**
--- a/libs/category/state/src/facades/category-facade.interface.ts
+++ b/libs/category/state/src/facades/category-facade.interface.ts
@@ -20,7 +20,7 @@ import { DaffCategoryReducerState } from '../reducers/category/category-reducer-
 
 /**
  * A facade for interacting with the category state.
- * Exposes many parts of the state for easy access and allows dispatching of actions.
+ * This facade exposes many parts of the state for easy access and allows dispatching of actions.
  */
 export interface DaffCategoryFacadeInterface<
 	V extends DaffGenericCategory<V> = DaffCategory,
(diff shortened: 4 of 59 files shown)


libs/category/src/models/category-breadcrumb.ts
```diff
@@ -7,7 +7,17 @@ import { DaffCategory } from './category';
...
+	/**
+	 * The id of the associated category.
+	 */
   id: DaffCategory['id'];
+	/**
+	 * A label to display the category name in the UI.
+	 */
   name: DaffCategory['name'];
+	/**
+	 * The number of nodes between this node and the root node for the category trie. For example, the level for a
```
xelaint, line 19: is `trie` supposed to be `tree`?

libs/category/src/models/category-breadcrumb.ts
```diff
@@ -7,7 +7,17 @@ import { DaffCategory } from './category';
...
+	 * "Men's Shirts" breadcrumb could be 2 if the breadcrumb trie looked like: "Men" => "Clothing" => "Men's Shirts".
```
xelaint, line 20: is `trie` supposed to be `tree`?

libs/category/src/models/generic-category.ts
```diff
@@ -11,12 +11,36 @@ import { DaffCategoryBreadcrumb } from './category-breadcrumb';
...
+	/**
+	 * The name of the category that can be used as a UI label.
+	 */
 	name: string;
+	/**
+	 * A description of the category.
+	 */
 	description?: string;
+	/**
+	 * The number of child category nodes under this category.
```
xelaint, line 27: `the category`?

libs/category/src/models/generic-category.ts
```diff
@@ -11,12 +11,36 @@ import { DaffCategoryBreadcrumb } from './category-breadcrumb';
...
+	 */
   children_count?: number;
+	/**
+	 * The number of products described by this category.
```
xelaint, line 31: `the category`?

libs/category/src/models/generic-category.ts
```diff
@@ -11,12 +11,36 @@ import { DaffCategoryBreadcrumb } from './category-breadcrumb';
...
+	 */
   total_products?: number;
+	/**
+	 * The direct child categories of this category.
```
xelaint, line 35: `the category`?

libs/category/src/models/generic-category.ts
```diff
@@ -11,12 +11,36 @@ import { DaffCategoryBreadcrumb } from './category-breadcrumb';
...
+	 */
   children?: T[];
+	/**
+	 * The ids for all of the products described by this category.
```
xelaint, line 39: `the category`?

libs/category/src/models/generic-category.ts
```diff
@@ -11,12 +11,36 @@ import { DaffCategoryBreadcrumb } from './category-breadcrumb';
...
+	 */
   product_ids?: string[];
+	/**
+	 * The breadcrumb trail that leads from this category to its root category.
```
xelaint, line 43: `the category`?

libs/category/src/models/get-category-response.ts
```diff
@@ -13,7 +13,16 @@ export interface DaffGetCategoryResponse<
...
+	/**
+	 * The list of products described by the category.
+	 */
   products: W[];
+	/**
+	 * The category requested.
+	 */
   category: V;
+	/**
+	 * The properties of a category page.
```
xelaint, line 25: `the category`?

libs/category/state/src/facades/category-facade.interface.ts
```diff
@@ -19,6 +19,10 @@ import { DaffProduct } from '@daffodil/product';
 
 import { DaffCategoryReducerState } from '../reducers/category/category-reducer-state.interface';
 
+/**
+ * A facade for interacting with the category state.
+ * Exposes many parts of the state for easy access and allows dispatching of actions.
```
xelaint, line 24: `It exposes`?

libs/category/state/src/facades/category.facade.ts
```diff
@@ -28,6 +28,9 @@ import { getDaffCategorySelectors } from '../selectors/category.selector';
 import { DaffCategoryFacadeInterface } from './category-facade.interface';
 
 /**
+ * A facade for interacting with the category state.
+ * Exposes many parts of the state for easy access and allows dispatching of actions.
```
xelaint, line 32: `It exposes`?

libs/category/state/src/selectors/category.selector.ts
```diff
@@ -23,17 +23,41 @@ import {
...
+	/**
+	 * Selects the current category
+	 */
   selectSelectedCategory: MemoizedSelector<Record<string, any>, V>;
+	/**
+	 * Selects the products of the current category.
+	 */
   selectCategoryPageProducts: MemoizedSelector<Record<string, any>, W[]>;
+	/**
+	 * Selects a category by id.
```
xelaint, line 45: Selects `the`

xelaint (changes requested):

libs/category/src/models/category-breadcrumb.ts
lderrickable: I meant `trie` but I think I'm misusing the term. I'll switch to "tree"

libs/category/src/models/generic-category.ts
lderrickable: I'd rather use "this" when there are multiple categories "the" could be referring to. I could probably use "the" and presume that the reader will correctly assume I mean "this" (and they probably would), but I could also just use "this".

libs/category/state/src/selectors/category.selector.ts
lderrickable: I'd rather use "a" here. "the" sounds weird to me. I had to sit here for 10 minutes trying to figure out why, but I think I've got it. I think "Selects a category" suggests a one of many relationship, whereas "Selects the category" does not. When you add the rest of it "Selects a category by id." vs "Selects the category by id." they do seem to mean the same thing, but it feels like the "the" article is working against the rest of the statement. Does that make sense to you, or am I crazy?

libs/category/state/src/facades/category.facade.ts
lderrickable: changed

libs/category/state/src/facades/category-facade.interface.ts
lderrickable: changed

libs/category/src/models/get-category-response.ts
lderrickable: changed

libs/category/src/models/generic-category.ts
lderrickable: changed

libs/category/src/models/generic-category.ts
lderrickable: changed

libs/category/src/models/category-breadcrumb.ts
lderrickable: changed

libs/category/src/models/category-breadcrumb.ts
lderrickable: changed

libs/category/src/models/generic-category.ts
xelaint: In that case, `a category` would probably make more sense? What is `this` category?

libs/category/state/src/selectors/category.selector.ts
```diff
@@ -23,17 +23,41 @@ import {
...
   V extends DaffGenericCategory<V> = DaffCategory,
   W extends DaffProduct = DaffProduct
   > extends
   DaffCategoryFeatureMemoizedSelectors<V>,
   DaffCategoryPageMemoizedSelectors<V>,
   DaffCategoryEntitiesMemoizedSelectors<V> {
+	/**
+	 * Selects the current category
+	 */
   selectSelectedCategory: MemoizedSelector<Record<string, any>, V>;
```
xelaint, line 39: Kind of related to this PR because of the docs you wrote — wouldn't it make more sense for this property to be named `selectCurrentCategory`.

libs/category/state/src/selectors/category.selector.ts
xelaint: What if I phrased it to be selects the current category by id given that the other docs are referencing current category. selects a current category by id no longer makes sense.

libs/category/src/models/get-category-response.ts
```diff
@@ -13,7 +13,16 @@ export interface DaffGetCategoryResponse<
 	V extends DaffGenericCategory<V> = DaffCategory,
 	W extends DaffProduct = DaffProduct
 > {
+	/**
+	 * The list of products described by the requested category.
```
griest024, line 17: I would say `The list of products that belong to the requested category`

libs/category/src/models/get-category-response.ts
```diff
@@ -13,7 +13,16 @@ export interface DaffGetCategoryResponse<
...
+	 */
   products: W[];
+	/**
+	 * The category requested.
```
griest024, line 21: I prefer `The requested category`

libs/category/state/src/reducers/category-entities/category-entities.reducer.ts
```diff
@@ -16,6 +16,9 @@ import {
 } from '../../actions/category.actions';
 import { daffCategoryEntitiesAdapter } from './category-entities-adapter';
 
+/**
+ * Returns the state for category entities.
```
griest024, line 20: Say something about the reducer storing category entities in response to certain actions.

libs/category/state/src/selectors/category-page/category-page.selector.ts
```diff
@@ -22,27 +22,72 @@ import {
...
+	/**
+	 * Selects the sort options that may be applied to the current category.
+	 */
 	selectCategorySortOptions: MemoizedSelector<Record<string, any>, DaffCategoryPageMetadata['sort_options']['options']>;
+	/**
+	 * Selects the ids of all products in the current category page.
+	 */
 	selectCategoryPageProductIds: MemoizedSelector<Record<string, any>, DaffCategoryPageMetadata['product_ids']>;
+	/**
+	 * Selects whether or not the category page has no products.
```
griest024, line 64: Remove double negative

griest024 (changes requested):

libs/category/state/src/selectors/category.selector.ts
lderrickable: If it was changed to a "current category" selector, then I would agree that "the" makes sense, because "the" designates a particular category from the beginning, i.e. the current category. Since there can only ever be one current category, "a current category" wouldn't make sense. But that's not what this selector does. It has no idea what the current category is. It just returns whatever category you want based on the id provided.

I think your argument is this: if a doc has a broad meaning (selects a category by id), then it should also be able to cover a more specific circumstance too (if I give it the id of the current category, it should return the current category). Therefore, the doc should cover the more specific circumstance of giving it a particular value. The doc should be rewritten as "Selects the category by id".

This argument sounds congruent to your argument: if a doc has a broad meaning (selects a category by id), then it should also be able to cover a more specific circumstance too (if I give it the id "4", it should return the category related to the id "4"). Therefore, the doc should cover the move specific circumstance of giving it a "4" id. The doc should be rewritten to "Selects the category with id=4."

Given, the second argument seems obviously wrong, because it takes the first argument to the extreme. But I think they are essentially the same. They both try to justify a more narrow doc by supplying a more narrow example that the selector covers

libs/category/src/models/generic-category.ts
lderrickable: "This" category is the `DaffGenericCategory`. "a category" would make less sense, I think, because it would be even more arbitrary than "the".

Oh, that's just because I make mistakes sometimes (some might say, "often"). I'll change it to "this".

libs/category/state/src/selectors/category.selector.ts
lderrickable: I think so, yea. I don't think I should make breaking changes in this PR though. I can remove the doc if you'd prefer.

libs/category/src/models/generic-category.ts
lderrickable: Oh actually, nevermind, I meant to have that one be a "the". I chose "The number of products described by the category." because I know you prefer "the" to "this" and I didn't think "this" was necessary because there is no other category to get confused with as there is no reference to any other category in this particular description. So "the" category can only refer to the `DaffGenericCategory`.

libs/category/src/models/get-category-response.ts
lderrickable: changed

libs/category/src/models/get-category-response.ts
lderrickable: Changed

libs/category/state/src/reducers/category-entities/category-entities.reducer.ts
lderrickable: done

libs/category/state/src/selectors/category-page/category-page.selector.ts
lderrickable: done

griest024 (dismissed):

libs/category/state/src/selectors/category.selector.ts
xelaint: I think you can keep the doc, but make an issue about updating the property name?

xelaint:
There's two changes to `generic-category.ts` that still need to be addressed but lgtm after those changes + rebase.

libs/category/src/models/generic-category.ts
lderrickable: This is the same situation as the other discussion we were having: 
> I'd rather use "this" when there are multiple categories "the" could be referring to. I could probably use "the" and presume that the reader will correctly assume I mean "this" (and they probably would), but I could also just use "this".

libs/category/src/models/generic-category.ts
lderrickable: This is the same situation as the other discussion we were having:
> I'd rather use "this" when there are multiple categories "the" could be referring to. I could probably use "the" and presume that the reader will correctly assume I mean "this" (and they probably would), but I could also just use "this".

libs/category/state/src/selectors/category.selector.ts
lderrickable: added: https://github.com/graycoreio/daffodil/issues/1648

xelaint (dismissed):

lderrickable:
I think this needs a reapproval by @griest024

griest024 (approved):

griest024 (approved):

PR #1606, opened 2021-06-17
fix(daffio): the routing around api and guide docs is not correct

lderrickable:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Both guide docs and api docs use the same routing module, which ends up just rendering both with the api docs components. This was not noticed previously because the api docs and guide docs both looked the same. Now with the introduction of the table of contents only for guide docs, I noticed the bug.

## What is the new behavior?
The guide docs and api docs each have their own routing module.

```
[ ] Yes
[x] No
```


griest024:
The sidebar closes when navigating to an API docs page, is that intended?

lderrickable:
> The sidebar closes when navigating to an API docs page, is that intended?

I'm not sure. But I think I'd prefer that it does close.

griest024:
> > The sidebar closes when navigating to an API docs page, is that intended?
> 
> I'm not sure. But I think I'd prefer that it does close.

I'm not sure I agree with that but at least make sure the sidebar is closing for a good reason, not just because stuff is breaking.

damienwebdev:
I think we leave any UI concerns out until after the basic features are merged. The entire app is a travesty in terms of UI currently. We'll address these concerns once we have a working understanding of what the available capabilities are.

griest024:
okay I'll be more lenient about daff.io while its in heavy breaking development

griest024 (approved):

lderrickable:
> > > The sidebar closes when navigating to an API docs page, is that intended?
> > 
> > 
> > I'm not sure. But I think I'd prefer that it does close.
> 
> I'm not sure I agree with that but at least make sure the sidebar is closing for a good reason, not just because stuff is breaking.

I did check that it was that way in develop, so I don't think anything changed between develop and this branch. I'll check again to make sure though

Why would you want the sidebar to remain open when you navigate to a different page?

lderrickable:
@griest024 I think we just misunderstood each other. I originally thought you were suggesting that the sidebar behavior had changed (and I think you did mean this), so I checked develop and it didn't seem like the behavior changed to me. So then I just took your question at face value and were asking if the behavior was always intended that way, and I didn't know the answer to that. I kind of do think we should avoid unintentionally changing things if we can, but I'd be ok with the "full steam ahead" approach too.

griest024:
I did mean that the sidebar behavior changed between this branch and develop. Let me check again

griest024:
yeah the sidebar behavior is definitely different for me.

lderrickable:
huh. That's weird

PR #1618, opened 2021-06-18
feat(product): change driver interface to return response object

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
The product driver `get` and `getByUrl` calls now return the driver response model.

```
[x] Yes
[ ] No
```

driver interface changed

## Other information

### DIFF
--- a/libs/product/driver/shopify/src/product.service.ts
+++ b/libs/product/driver/shopify/src/product.service.ts
@@ -9,11 +9,15 @@ import {
 } from 'rxjs';
 import { map } from 'rxjs/operators';
 
+import { daffArrayToDict } from '@daffodil/core';
 import {
   DaffProduct,
   DaffProductTypeEnum,
 } from '@daffodil/product';
-import { DaffProductServiceInterface } from '@daffodil/product/driver';
+import {
+  DaffProductDriverResponse,
+  DaffProductServiceInterface,
+} from '@daffodil/product/driver';
 
 interface GetAllProductsResponse {
   shop?: ShopGraph;
@@ -112,9 +116,12 @@ export class DaffShopifyProductService implements DaffProductServiceInterface {
 
   constructor(private apollo: Apollo) {}
 
-  getByUrl(url: DaffProduct['url']): Observable<DaffProduct> {
+  getByUrl(url: DaffProduct['url']): Observable<DaffProductDriverResponse> {
     // TODO: implement
-    return of();
+    return of({
+      id: null,
+      products: {},
+    });
   }
 
   getAll(): Observable<DaffProduct[]> {
@@ -140,14 +147,23 @@ export class DaffShopifyProductService implements DaffProductServiceInterface {
     );
   }
 
-  get(productId: string): Observable<DaffProduct> {
+  /**
+   * A query for retrieving a particular product as an Observable<DaffProduct>.
+   *
+   * @param productId - A product ID
+   * @returns Observable<Product>
+   */
+  get(productId: string): Observable<DaffProductDriverResponse> {
     return this.apollo.query<GetAProductResponse>({
       query: GetAProduct,
       variables: {
         id: productId,
       },
     }).pipe(
-      map(result => DaffShopifyProductTransformer(result.data.node)),
+      map(result => ({
+        id: productId,
+        products: daffArrayToDict([DaffShopifyProductTransformer(result.data.node)], p => p.id),
+      })),
     );
   }
 }
--- a/libs/product/driver/shopify/src/product.service.ts
+++ b/libs/product/driver/shopify/src/product.service.ts
@@ -120,7 +120,7 @@ export class DaffShopifyProductService implements DaffProductServiceInterface {
     // TODO: implement
     return of({
       id: null,
-      products: {},
+      products: [],
     });
   }
 
@@ -156,7 +156,7 @@ export class DaffShopifyProductService implements DaffProductServiceInterface {
     }).pipe(
       map(result => ({
         id: productId,
-        products: daffArrayToDict([DaffShopifyProductTransformer(result.data.node)], p => p.id),
+        products: [DaffShopifyProductTransformer(result.data.node)],
       })),
     );
   }
--- a/libs/product/driver/shopify/src/product.service.ts
+++ b/libs/product/driver/shopify/src/product.service.ts
@@ -147,12 +147,6 @@ export class DaffShopifyProductService implements DaffProductServiceInterface {
     );
   }
 
-  /**
-   * A query for retrieving a particular product as an Observable<DaffProduct>.
-   *
-   * @param productId - A product ID
-   * @returns Observable<Product>
-   */
   get(productId: string): Observable<DaffProductDriverResponse> {
     return this.apollo.query<GetAProductResponse>({
       query: GetAProduct,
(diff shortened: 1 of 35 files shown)


libs/product/driver/testing/src/drivers/product.service.ts
```diff
@@ -46,11 +50,26 @@ export class DaffTestingProductService implements DaffProductServiceInterface {
...
 
-  get(productId: DaffProduct['id']): Observable<DaffProduct> {
-    return of(this.productFactory.create({ images: this.productImageFactory.createMany(5) }));
+  /**
+   * Get product by ID
+   *
+   * @param productId product ID
+   * @returns An Observable of a Product
+   */
+  get(productId: DaffProduct['id']): Observable<DaffProductDriverResponse> {
```
lderrickable, line 59: This doesn't need a doc

libs/product/state/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
```diff
@@ -49,9 +49,11 @@ export function daffCompositeProductEntitiesReducer<T extends DaffProduct, V ext
       );
     case DaffProductActionTypes.ProductLoadSuccessAction:
     case DaffProductPageActionTypes.ProductPageLoadSuccessAction:
-      if(action.payload.type === DaffProductTypeEnum.Composite) {
+      const product = action.payload.products[action.payload.id];
```
lderrickable, line 52: Since related and upsell products might be composite, don't you need to do something like: 
```
return adapter.upsertMany(
    action.payload.products.filter(product => product.type === DaffProductTypeEnum.Composite)
);
```
That would be if it was an array anyway. I guess that begs the question: why do we want this to be a dictionary instead of an array?

libs/product/state/src/reducers/configurable-product-entities/configurable-product-entities.reducer.ts
```diff
@@ -50,9 +50,11 @@ export function daffConfigurableProductEntitiesReducer<T extends DaffProduct, V
       );
     case DaffProductActionTypes.ProductLoadSuccessAction:
     case DaffProductPageActionTypes.ProductPageLoadSuccessAction:
```
lderrickable, line 52: Same here as the comment above, since more than one of the incoming products could be configurable.

libs/product/state/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
lderrickable: Oh, is it a dictionary because that assumes we'll need to access the main product more often than we'll need to iterate over all products? I'm not sure that's true though

lderrickable (changes requested):

libs/product/state/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
griest024: > why do we want this to be a dictionary instead of an array

Easy uniqueness assurances

griest024: So the regular entity state doesn't have a check for simple type. Does that mean that products are duplicated in different entity states? I do think your suggestion is necessary to implement but the multiple entity states are confusing.

griest024: nvm ignore that ^ I see that the states are just storing the extra fields that aren;t covered by the main entity state

griest024: changed

libs/product/driver/testing/src/drivers/product.service.ts
griest024: removed

lderrickable (changes requested):

libs/product/state/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
lderrickable: > > why do we want this to be a dictionary instead of an array
> 
> Easy uniqueness assurances

I think we're given the same uniqueness assurances by entity state. I see `DaffProductResponse` as just a vehicle to get products from the driver call into state. I think it probably makes more sense to use an array in the `DaffProductResponse` instead of a Dictionary, because you're just transforming from an array to a dictionary and then back into an array when the products are loaded into entity state. Then the `upsertMany` converts it back into a dictionary. I'm not even sure there will ever be a time when we'll need to get only the main product (which is when the dictionary will be useful). 

Additionally, if you're worries about uniqueness between the driver call and state, I don't think there will be any difference in practice between using an array vs dictionary. If it is a dictionary and there are two products with the same Id returned from the backend, then the first instance of the id will be overwritten in the driver layer. If it is an array, the overwriting will happen in the entity state layer instead of the driver.

libs/product/driver/shopify/src/product.service.ts
```diff
@@ -140,14 +147,23 @@ export class DaffShopifyProductService implements DaffProductServiceInterface {
...
   }
 
-  get(productId: string): Observable<DaffProduct> {
+  /**
+   * A query for retrieving a particular product as an Observable<DaffProduct>.
+   *
+   * @param productId - A product ID
+   * @returns Observable<Product>
+   */
+  get(productId: string): Observable<DaffProductDriverResponse> {
```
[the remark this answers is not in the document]
griest024: removed this and the others

libs/product/state/src/reducers/composite-product-entities/composite-product-entities.reducer.ts
griest024: That's well reasoned. I'm convinced but what does @damienwebdev think? And do we even try to ensure uniqueness in the products array coming from the driver?

griest024: changed

damienwebdev: I agree.

libs/product/driver/testing/src/drivers/product.service.spec.ts
```diff
@@ -52,23 +52,24 @@ describe('Driver | Testing | Product | ProductService', () => {
...
       const url = 'url';
 
-      productService.getByUrl(url).subscribe(product => {
-        expect(isProduct(product)).toBeTruthy();
-        expect(product.images).toEqual(stubProductImages);
-        expect(product.url).toEqual(url);
+      productService.getByUrl(url).subscribe(resp => {
+        expect(isProduct(Object.values(resp.products)[0])).toBeTruthy();
+        expect(Object.values(resp.products)[0].images).toEqual(stubProductImages);
+        expect(Object.values(resp.products)[0].url).toEqual(url);
```
lderrickable, line 72: All these usages of `Object.values` can be removed I think

griest024: removed

lderrickable (approved):

PR #1626, opened 2021-06-22
feat(product): add upsell and related products

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Adds related and upsell fields to DaffProduct

```
[ ] Yes
[x] No
```

## Other information
The reason why the Magento product preview contains so much info is to allow prices to be computed for composites and configurables


libs/product/driver/magento/src/queries/fragments/product-preview.ts
```diff
@@ -0,0 +1,30 @@
+import { gql } from 'apollo-angular';
+
+
+export const magentoProductPreviewFragment = gql`
+  fragment magentoProductPreview on ProductInterface {
```
damienwebdev, line 5: Does this type need the composite and configurable fragments?

damienwebdev (changes requested):
Have you tried this with the query complexity limiter of Magento?

lderrickable:
I think the `transformMagentoProductResponse` transformer needs updated too, so the upsell and related products can be added to the `DaffProductDriverResponse.products` array.

griest024:
> I think the `transformMagentoProductResponse` transformer needs updated too, so the upsell and related products can be added to the `DaffProductDriverResponse.products` array.

Yeah it does but I'm working through the infinite recursion that it causes

lderrickable:
> > I think the `transformMagentoProductResponse` transformer needs updated too, so the upsell and related products can be added to the `DaffProductDriverResponse.products` array.
> 
> Yeah it does but I'm working through the infinite recursion that it causes

I think I'm missing something. Wouldn't it just be: 
```
export function transformMagentoProductResponse(product: MagentoProduct, mediaUrl: string): DaffProductDriverResponse {
  const daffProduct = transformMagentoProduct(product, mediaUrl);
  return {
    id: daffProduct.id,
    products: [daffProduct, ...daffProduct.upsell_products, ...daffProduct.related_products],
  };
}
```
?

libs/product/driver/magento/src/queries/fragments/product-preview.ts
griest024: yes, added

griest024:
> > > I think the `transformMagentoProductResponse` transformer needs updated too, so the upsell and related products can be added to the `DaffProductDriverResponse.products` array.
> > 
> > 
> > Yeah it does but I'm working through the infinite recursion that it causes
> 
> I think I'm missing something. Wouldn't it just be:
> 
> ```
> export function transformMagentoProductResponse(product: MagentoProduct, mediaUrl: string): DaffProductDriverResponse {
>   const daffProduct = transformMagentoProduct(product, mediaUrl);
>   return {
>     id: daffProduct.id,
>     products: [daffProduct, ...daffProduct.upsell_products, ...daffProduct.related_products],
>   };
> }
> ```
> 
> ?

Yes it is, good call. However that's not the issue that causes the infinite recursion.

The existing transforms assume that the product is a full product, and not just a preview. Therefore I had to separate out the bundle and configurable transforms that dealt with the extra fields from the full transform that spread in the simple product.

I kind of hate my solution so suggestions about a more elegant solution is appreciated.

griest024:
@lderrickable oh sorry I misread your initial comment, you were only talking about the response transformer. The changes to the response transformer don't cause the recursion. Hopefully all the issues have been resolved.

libs/product/driver/magento/src/queries/fragments/product-preview.ts
griest024: * added to the main product fragment, not to the preview fragment but the resulting query is equivalent

libs/product/driver/magento/src/models/product-preview.interface.ts
```diff
@@ -0,0 +1,29 @@
+import { MagentoProduct } from './magento-product';
```
lderrickable, line 1: import should be removed

libs/product/driver/magento/src/queries/fragments/product-preview.ts
lderrickable: wouldn't we want these on the preview fragment? That way, the configurable and bundle fields will be on both previews and products. And we'll want those fields for upsell and related products.

lderrickable (changes requested):

lderrickable:
>The existing transforms assume that the product is a full product, and not just a preview. Therefore I had to separate out the bundle and configurable transforms that dealt with the extra fields from the full transform that spread in the simple product.

I kind of hate my solution so suggestions about a more elegant solution is appreciated.

I think I've got an idea. What do you think of this:
```
export function transformMagentoProduct(product: MagentoProduct, mediaUrl?: string): DaffProduct {
  switch(product.__typename) {
    case MagentoProductTypeEnum.BundledProduct:
      return transformMagentoBundledProduct(
         transformMagentoSimpleProduct(product, mediaUrl),
         <MagentoBundledProduct>product
      );
    case MagentoProductTypeEnum.ConfigurableProduct:
      return transformMagentoConfigurableProduct(
         transformMagentoSimpleProduct(product, mediaUrl),
         <MagentoConfigurableProduct>product,
      );
    default:
      return transformMagentoSimpleProduct(product, mediaUrl);
  }
}

export function transformMagentoBundledProduct(daffProduct: DaffProduct, product: MagentoBundledProduct, mediaUrl: string): DaffCompositeProduct {
  return {
    ...daffProduct,
    price: 0,
    discount: {
      amount: 0,
      percent: 0,
    },
    type: DaffProductTypeEnum.Composite,
    items: product.items.map(transformMagentoBundledProductItem),
  };
}

export function transformProductPreview(product: MagentoProductPreview, mediaUrl: string): DaffProduct {
  switch(product.__typename) {
    case MagentoProductTypeEnum.BundledProduct:
      return transformMagentoBundledProduct(
         transformMagentoSimpleProductPreview(product, mediaUrl),
         <MagentoBundledProduct>product
      );
    case MagentoProductTypeEnum.ConfigurableProduct:
      return transformMagentoConfigurableProduct(
         transformMagentoSimpleProductPreview(product, mediaUrl),
         <MagentoConfigurableProduct>product
      );
    default:
      return transformMagentoSimpleProductPreview(product, mediaUrl);
  }
}
```

griest024:
> wouldn't we want these on the preview fragment? That way, the configurable and bundle fields will be on both previews and products. And we'll want those fields for upsell and related products.

It's spread in the parent product fragment so the fields are there. It doesn't matter too much either way; I can change it if you feel strongly.

libs/product/driver/magento/src/models/product-preview.interface.ts
griest024: removed

griest024:
@lderrickable 

> I think I've got an idea. What do you think of passing an already transformed DaffProduct into the bundle/configurable transformer functions:

nice one, its certainly better than what I had. I also changed the switch case transform to accept a daff product so we only needed one switch case.

libs/product/driver/magento/src/queries/fragments/product.ts
```diff
@@ -48,11 +28,24 @@ export const magentoProductFragment = gql`
 		description {
 			html
 		}
+    upsell_products {
+      ...magentoProductPreview
+      ...magentoBundledProduct
+      ...magentoSimpleProduct
```
lderrickable, line 34: I don't think we want to spread the Simple product fragment here. Kind of defeats the purpose of having the preview fragment, right?

lderrickable:
> > wouldn't we want these on the preview fragment? That way, the configurable and bundle fields will be on both previews and products. And we'll want those fields for upsell and related products.
> 
> It's spread in the parent product fragment so the fields are there. It doesn't matter too much either way to me; I can change it if you feel strongly.

Oh, I see. I didn't realize you'd spread those fragments into the upsell and related fields. lgtm

libs/product/driver/magento/src/transforms/product-transformers.ts
```diff
@@ -8,27 +8,48 @@ import {
...
+ */
+export function transformProduct(product: MagentoProduct, mediaUrl: string): DaffProduct {
+  return transformMagentoProduct(transformMagentoSimpleProduct(product, mediaUrl), product);
+}
+
+/**
+ * Transforms a Magento product preview into a minimal Daffodil product.
+ * Handles all product types.
+ */
+export function transformProductPreview(product: MagentoProduct, mediaUrl: string): DaffProduct {
```
lderrickable, line 26: I think the names of `transformProduct` and `transformProductPreview` are a little confusing and I don't think they are necessary. I think it should just be the `transformMagentoProduct` function here:
```
export function transformMagentoProduct(product: MagentoProduct): DaffProduct {
  switch(product.__typename) {
    case MagentoProductTypeEnum.BundledProduct:
      return transformMagentoBundledProduct(transformMagentoSimpleProduct(product), <MagentoBundledProduct>product);
    case MagentoProductTypeEnum.ConfigurableProduct:
      return transformMagentoConfigurableProduct(transformMagentoSimpleProduct(product), <MagentoConfigurableProduct>product);
    case MagentoProductTypeEnum.SimpleProduct:
    default:
      return transformMagentoSimpleProduct(product);
  }
}
```
and then a congruent function in previews. Otherwise the api here is pretty confusing

libs/product/driver/magento/src/queries/fragments/product.ts
griest024: all the simple fragment does is add the name and AFAIK we do want the name. or do you mean it should be spread into the preview fragment?

libs/product/driver/magento/src/transforms/product-transformers.ts
griest024: changed

libs/product/driver/magento/src/queries/fragments/product.ts
lderrickable: oh whoops. I keep mixing up the differences between `DaffProduct` and `MagentoProduct`. I was thinking that `MagentoSimpleProduct` was the `MagentoProduct` (like it is in daffodil models). lgtm

libs/product/driver/magento/src/transforms/product-preview.ts
```diff
@@ -5,21 +5,39 @@ import {
...
-  [MagentoProductTypeEnum.BundledProduct]: DaffProductTypeEnum.Composite,
-  [MagentoProductTypeEnum.ConfigurableProduct]: DaffProductTypeEnum.Configurable,
-  [MagentoProductTypeEnum.SimpleProduct]: DaffProductTypeEnum.Simple,
-};
+/**
+ * Transforms a Magento product into a product preview.
+ * Handles all product types.
+ */
+export function transformMagentoProductPreview(product: MagentoProduct, mediaUrl: string): DaffProduct {
+  const preview = transformMagentoSimpleProductPreview(product, mediaUrl);
```
lderrickable, line 23: I think it's technically better to just call `transformMagentoSimpleProductPreview` in each switch case, since every `transformMagentoProductPreview` call only needs to call `transformMagentoSimpleProductPreview` one time. That way, it doesn't need to save the transformed preview in memory. 

Same for the `transformMagentoProduct` function

griest024: changed

lderrickable:
lgtm

lderrickable (approved):

PR #1634, opened 2021-06-24
feat(cart): deprecate DaffCartItem.item_id and extend DaffIdentifiabl…

lderrickable:
…e instead

## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The `DaffCartItem` has an id field that is named `item_id`. It could just be `id` instead and extend the `DaffIdentifiable` interface.

## What is the new behavior?
`DaffCartItem.item_id` is deprecated and `DaffCartItem` extends `DaffIdentifiable`. All instances of the `item_id` usages are replaced by `id`, but `item_id` is still populated.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/cart/testing/src/factories/cart-item/cart-item.factory.ts
+++ b/libs/cart/testing/src/factories/cart-item/cart-item.factory.ts
@@ -11,8 +11,8 @@ import { DaffProductImage } from '@daffodil/product';
 import { DaffProductImageFactory } from '@daffodil/product/testing';
 
 export class DaffMockCartItem implements DaffCartItem {
-	item_id = faker.datatype.uuid();
 	id = faker.datatype.uuid();
+	item_id = this.id;
 	type = DaffCartItemInputType.Simple;
   product_id = faker.datatype.uuid();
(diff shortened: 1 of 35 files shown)


libs/cart/testing/src/factories/cart-item/cart-item.factory.ts
```diff
@@ -12,6 +12,7 @@ import { DaffProductImageFactory } from '@daffodil/product/testing';
 
 export class DaffMockCartItem implements DaffCartItem {
 	item_id = faker.datatype.uuid();
+	id = faker.datatype.uuid();
```
griest024, line 15: This should be identical to `item_id`

griest024 (changes requested):

libs/cart/testing/src/factories/cart-item/cart-item.factory.ts
lderrickable: fixed

griest024 (approved):

PR #1646, opened 2021-06-28
feat(seo): add a service for managing description meta tags

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Adds a service for managing description meta tags, which are used by search engines to display the descriptions of each search result.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/seo/src/meta-description/meta-description.interface.ts
+++ b/libs/seo/src/meta-description/meta-description.interface.ts
@@ -1,0 +1,16 @@
+/**
+ * A meta description service is responsible for managing the meta description tag for the current page.
+ */
+export interface DaffMetaDescriptionServiceInterface {
+  /**
+   * Inserts or updates, as needed, a meta description tag on the current page.
+   *
+   * @param content The meta tag content
+   */
+  upsert(content: string): void;
+
+  /**
+   * Removes the meta tag, if it exists, from the current page.
+   */
+  remove(): void;
+}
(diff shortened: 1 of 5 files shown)


griest024 (commented):
Do you think its worth adding a spec for the meta tag behavior when changing pages? I don't know if its likely to be a problem but it might be worth it since its the behavior that breaks with the Angular `Meta` service.

lderrickable:
> Do you think its worth adding a spec for the meta tag behavior when changing pages? I don't know if its likely to be a problem but it might be worth it since its the behavior that breaks with the Angular `Meta` service.

I'd say that's already covered. The problem with the Angular `Meta` service is that when you call `addTag` twice, it will add an additional `meta` tag. So you'd need to call the `Meta.removeTag` function, then the `Meta.addTag` whenever you navigated to a different page. But this scenario is covered by the `when the description meta tag exists, it should update the existing meta element` test. When you navigate to a new page, it will upsert the meta tag rather than adding it.

griest024 (dismissed):

griest024 (changes requested):
Can you reimplement this by relying on the `Meta` service?

damienwebdev:
Closing as it appears that https://github.com/angular/angular/issues/42700 is a real bug in Angular and the `upsert`/`remove` behaviors created here can be replicated with features of the original `Meta` service of Angular.

https://stackblitz.com/edit/angular-ivy-zzejke?file=src%2Fapp%2Fapp.component.ts

Is a working example of such code.

A future PR for such a service should outline a better, more general Meta service with a more thought-out API than the original Angular Service.

PR #1652, opened 2021-06-30
docs(cart): add root cart package API docs

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information
There were a few outdated and confusing things that I added TODOs for, lmk if you can clear those up.

### DIFF
--- a/libs/cart/src/models/cart-coupon.ts
+++ b/libs/cart/src/models/cart-coupon.ts
@@ -1,7 +1,21 @@
 import { ID } from '@daffodil/core';
 
+/**
+ * A coupon applied to the cart.
+ * Also known as a promo code.
+ */
 export interface DaffCartCoupon {
+  /**
+   * The coupon ID.
+   */
+  // TODO: change to id
   coupon_id?: ID;
+  /**
+   * The coupon code.
+   */
   code: string;
+  /**
+   * An optional coupon description.
+   */
   description?: string;
 }
--- a/libs/cart/src/models/cart-coupon.ts
+++ b/libs/cart/src/models/cart-coupon.ts
@@ -8,7 +8,6 @@ export interface DaffCartCoupon {
   /**
    * The coupon ID.
    */
-  // TODO: change to id
   coupon_id?: ID;
   /**
    * The coupon code.
--- a/libs/cart/src/models/cart-item-input.ts
+++ b/libs/cart/src/models/cart-item-input.ts
@@ -54,7 +54,7 @@ export interface DaffCompositeCartItemInputOption {
   /**
    * An identifier of the option for which a value has been chosen.
    */
-  // TODO: change to id?
+  // TODO: change to ID type?
 	code: string | number;
   /**
    * How many of the specified option should be added to the cart.
--- a/libs/cart/src/models/cart-item-input.ts
+++ b/libs/cart/src/models/cart-item-input.ts
@@ -1,34 +1,75 @@
 import { ID } from '@daffodil/core';
 import { DaffProduct } from '@daffodil/product';
 
+/**
+ * An enum for the type of the cart item.
+ */
+// TODO: remove "input" designation
 export enum DaffCartItemInputType {
 	Simple = 'simple',
 	Composite = 'composite',
 	Configurable = 'configurable'
 }
 
+/**
+ * A set of fields specifying how to add a product to the cart.
+ */
 export interface DaffCartItemInput {
+  /**
+   * The type of cart item that correspondes to the product being added to the cart.
+   */
 	type: DaffCartItemInputType;
+  /**
+   * The ID of the product to add to the cart.
+   */
   productId: DaffProduct['id'];
+  /**
+   * How many of the product should be added to the cart.
+   */
 	qty: number;
 }
 
+/**
+ * A more constrained cart item input for simple products.
+ */
 export interface DaffSimpleCartItemInput extends DaffCartItemInput {
 	type: DaffCartItemInputType.Simple;
 }
 
+/**
+ * An cart item input for composite products.
+ */
 export interface DaffCompositeCartItemInput extends DaffCartItemInput {
 	type: DaffCartItemInputType.Composite;
+  /**
+   * The selected composite product options to be added to the cart.
+   */
 	options: DaffCompositeCartItemInputOption[];
 }
 
+/**
+ * Designates a particular composite product option to add to the cart.
+ */
 export interface DaffCompositeCartItemInputOption {
+  // TODO: what's this for? can we remove it?
 	code: string | number;
+  /**
+   * How many of the specified option should be added to the cart.
+   */
 	quantity: number;
+  /**
+   * The value identifying the particular composite product option.
+   */
 	value: string;
 }
 
+/**
+ * Designates a particular configurable product variant to add to the cart.
+ */
 export interface DaffConfigurableCartItemInput extends DaffCartItemInput {
 	type: DaffCartItemInputType.Configurable;
+  /**
+   * The value identifying the particular configurable product variant.
+   */
 	variantId: ID;
 }
--- a/libs/cart/src/models/cart-item-input.ts
+++ b/libs/cart/src/models/cart-item-input.ts
@@ -51,14 +51,17 @@ export interface DaffCompositeCartItemInput extends DaffCartItemInput {
  * Designates a particular option for the composite product to add to the cart.
  */
 export interface DaffCompositeCartItemInputOption {
-  // TODO: what's this for? can we remove it?
+  /**
+   * An identifier of the option for which a value has been chosen.
+   */
+  // TODO: change to id?
 	code: string | number;
   /**
    * How many of the specified option should be added to the cart.
    */
 	quantity: number;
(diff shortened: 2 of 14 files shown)


libs/cart/src/models/cart-coupon.ts
```diff
@@ -1,7 +1,21 @@
...
 
+/**
+ * A coupon applied to the cart.
+ * Also known as a promo code.
+ */
 export interface DaffCartCoupon {
+  /**
+   * The coupon ID.
+   */
+  // TODO: change to id
```
lderrickable, line 11: I tried to change this to use the DaffIdentifiable interface, but it ended up being a little tricky because coupon_id is optional. Maybe we make `coupon_id` the same as the code with Magento (which doesn't seem to have an id other than the code)?

griest024: okay I'll remove this for now and we can address it later

griest024: removed

lderrickable: Hmm, yea the more I think about it, the more I think it makes sense. Added an issue for it here: https://github.com/graycoreio/daffodil/issues/1656

libs/cart/src/models/cart-item-input.ts
```diff
@@ -1,34 +1,75 @@
...
 	Composite = 'composite',
 	Configurable = 'configurable'
 }
 
+/**
+ * A set of fields specifying how to add a product to the cart.
+ */
 export interface DaffCartItemInput {
+  /**
+   * The type of cart item that correspondes to the product being added to the cart.
```
lderrickable, line 19: typo

libs/cart/src/models/cart-item-input.ts
```diff
@@ -1,34 +1,75 @@
...
 
+/**
+ * A more constrained cart item input for simple products.
+ */
 export interface DaffSimpleCartItemInput extends DaffCartItemInput {
 	type: DaffCartItemInputType.Simple;
 }
 
+/**
+ * An cart item input for composite products.
```
lderrickable, line 40: "An" => "A"

libs/cart/src/models/cart-item-input.ts
```diff
@@ -1,34 +1,75 @@
...
+ */
 export interface DaffCompositeCartItemInput extends DaffCartItemInput {
 	type: DaffCartItemInputType.Composite;
+  /**
+   * The selected composite product options to be added to the cart.
```
lderrickable, line 45: This sounds like the options themselves are being added directly to the cart. It should be reworded so it's clear the options are for a product that is being added to the cart.

libs/cart/src/models/cart-item-input.ts
```diff
@@ -1,34 +1,75 @@
...
+   */
 	options: DaffCompositeCartItemInputOption[];
 }
 
+/**
+ * Designates a particular composite product option to add to the cart.
```
lderrickable, line 51: Same here as the above comment:
> This sounds like the options themselves are being added directly to the cart. It should be reworded so it's clear the options are for a product that is being added to the cart.

libs/cart/src/models/cart-item-input.ts
griest024: fixed

libs/cart/src/models/cart-item-input.ts
griest024: fixed

libs/cart/src/models/cart-item-input.ts
griest024: changed

libs/cart/src/models/cart-item-input.ts
griest024: changed

libs/cart/src/models/cart-item-input.ts
```diff
@@ -1,34 +1,75 @@
...
+ */
 export interface DaffCompositeCartItemInputOption {
+  // TODO: what's this for? can we remove it?
 	code: string | number;
```
lderrickable, line 55: For an example product like a toolbox composite product, the `code` could be a hammer that you could optionally add to the final product. Then the `value` would be like a sledge hammer or a claw hammer. So the code has to do with the parent "item" within the child "option".

lderrickable: That being said, it might be possible to remove it, because options are always nested under items. I'm not sure if that would make our lives more difficult or not though

lderrickable: Oh, nevemind. We definitely need it. I was thinking this was the `DaffCompositeProductItemOption`

libs/cart/src/models/cart-item.ts
```diff
@@ -10,22 +10,50 @@ import {
...
+ * A cart item.
+ * Cart item types correspond to product types.
+ * They hold information about the product to which they correspond and how many of said product is in the cart.
+ */
 export interface DaffCartItem extends DaffLocatable, DaffIdentifiable {
 	/**
 	 * @deprecated use id instead.
 	 */
 	item_id: ID;
 	type: DaffCartItemInputType;
```
lderrickable, line 37: needs a doc

libs/cart/src/models/cart-payment.ts
```diff
@@ -1,4 +1,14 @@
+/**
+ * The kind of payment used for the cart.
+ */
 export interface DaffCartPaymentMethod {
+  /**
+   * Identifies the kind of payment used.
```
lderrickable, line 6: I think this needs an example, since the text is the same as the interface doc.

libs/cart/src/models/cart.ts
```diff
@@ -17,18 +22,45 @@ export interface DaffCart extends DaffIdentifiable {
 	 * @deprecated use totals instead
 	 */
   grand_total: number;
+  /**
+   * A list of coupon applied to the cart.
```
lderrickable, line 26: *coupons

libs/cart/src/models/cart.ts
```diff
@@ -17,18 +22,45 @@ export interface DaffCart extends DaffIdentifiable {
...
+  /**
+   * A list of totals for the cart.
+   */
   totals: DaffCartTotal[];
+  /**
+   * The selected shipping method.
+   */
   shipping_information: DaffCartShippingInformation | null;
+  /**
+   * A list of possible shipping methods.
```
lderrickable, line 54: Even though it would be using the same word, I think "A list of available shipping methods" sounds better

libs/cart/src/models/cart.ts
```diff
@@ -17,18 +22,45 @@ export interface DaffCart extends DaffIdentifiable {
...
+   */
   available_shipping_methods: DaffCartShippingRate[];
+  /**
+   * A list of possible payment methods.
```
lderrickable, line 58: Even though it would be using the same word, I think "A list of available payment methods" sounds better

libs/cart/src/models/composite-cart-item.ts
```diff
@@ -5,15 +5,31 @@ import {
...
+ */
 export interface DaffCompositeCartItem extends DaffCartItem {
+  /**
+   * A list of composite cart item options.
+   */
 	options: DaffCompositeCartItemOption[];
 }
 
+/**
+ * A composite product option in the cart item.
```
lderrickable, line 20: I think this should be "a cart item"

libs/cart/src/models/configurable-cart-item.ts
```diff
@@ -1,11 +1,22 @@
...
+ * A cart item for a configurable product.
+ */
 export interface DaffConfigurableCartItem extends DaffCartItem {
+  /**
+   * A list of the chosen configurable attributes.
+   */
 	attributes: DaffConfigurableCartItemAttribute[];
 }
 
 export interface DaffConfigurableCartItemAttribute {
```
lderrickable, line 13: needs a doc

griest024: changed

libs/cart/src/models/composite-cart-item.ts
griest024: changed

libs/cart/src/models/cart.ts
griest024: changed

libs/cart/src/models/cart.ts
griest024: changed

libs/cart/src/models/cart-item.ts
griest024: added

libs/cart/src/models/cart-item-input.ts
griest024: added some docs for it. I think it also makes sense to change it to the `ID` type. I don't like the `string | number` union

lderrickable (approved):

PR #1662, opened 2021-07-04
docs(dgeni): add link tags to daffodil models, classes, etc

lderrickable:
## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

For api docs, any generic types for fields are just typed as `V` or `T` from the `V extends SomeModel` in the typeParameters of the doc. Furthermore, none of the daffodil member types are linked to their associated doc.

## What is the new behavior?
Type parameters are added for each model, class, etc. (this happens when they have generics). Links are added for references to other daffodil models in the member types and type parameters.

```
[ ] Yes
[x] No
```

### DIFF
--- a/tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
+++ b/tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
@@ -9,15 +9,40 @@ export class AddLinkTagToDaffodilReferencesProcessor implements Processor {
 	$runBefore = ['rendering-docs'];
 
 	$process(docs: Document[]): Document[] {
+		let docDictionary = docs.reduce((acc, doc) => ({
+			...acc,
+			[doc.name]: true
+		}), {});
+
+		
 		return docs.map(doc => {
-			doc.typeParams = doc.typeParams?.slice(1, doc.typeParams.length-1).replace(/(Daff([A-Z]|[a-z])*)/g, '{@link $1}');
-
-			doc.members?.map(member => {
-				member.type = member.type.replace(/(Daff([A-Z]|[a-z])*)/, '{@link $1}');
-				return member;
-			})
-
-			return doc;
+			return this.addLinksToMemberTypes(
+				this.addLinksToTypeParams(doc, docDictionary), 
+				docDictionary
+			);
 		});
+	}
+
+	addLinksToTypeParams(doc, docDictionary) {
+		const typeParams = doc.typeParams?.match(/(\w*)/g).reduce((acc, match) => ({
+			...acc,
+			[match]: docDictionary[match] ? `{@link ${match}}` : match
+		}), {});
+		doc.typeParams = doc.typeParams?.slice(1, doc.typeParams.length-1).replace(/(\w*)/g, (match) => typeParams[match]);
+		return doc;
+	}
+
+	addLinksToMemberTypes(doc, docDictionary) {
+		doc.members?.map(member => {
+			const typeNames = member.type.match(/(\w*)/g).reduce((acc, match) => ({
+				...acc,
+				[match]: docDictionary[match] ? `{@link ${match}}` : match
+			}), {});
+			member.type = member.type.replace(/(\w*)/g, (match) => typeNames[match]);
+
+			return member;
+		})
+
+		return doc;
 	}
 }
--- a/tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
+++ b/tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
@@ -1,7 +1,7 @@
 import { Processor, Document } from 'dgeni';
 
 /**
- * Inherit docs content from parent interfaces.
+ * Adds a link tag ({@link }) around daffodil models, classes, etc.
  */
 export class AddLinkTagToDaffodilReferencesProcessor implements Processor {
 	name = 'addLinkTagToDaffodilReferences';
--- a/tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
+++ b/tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
@@ -1,0 +1,23 @@
+import { Processor, Document } from 'dgeni';
+
+/**
+ * Inherit docs content from parent interfaces.
+ */
+export class AddLinkTagToDaffodilReferencesProcessor implements Processor {
+	name = 'addLinkTagToDaffodilReferences';
+	$runAfter = ['docs-processed'];
+	$runBefore = ['rendering-docs'];
+
+	$process(docs: Document[]): Document[] {
+		return docs.map(doc => {
+			doc.typeParams = doc.typeParams?.slice(1, doc.typeParams.length-1).replace(/(Daff([A-Z]|[a-z])*)/g, '{@link $1}');
+
+			doc.members?.map(member => {
+				member.type = member.type.replace(/(Daff([A-Z]|[a-z])*)/, '{@link $1}');
+				return member;
+			})
+
+			return doc;
+		});
+	}
+}
--- a/tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
+++ b/tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
@@ -9,40 +9,26 @@ export class AddLinkTagToDaffodilReferencesProcessor implements Processor {
 	$runBefore = ['rendering-docs'];
 
 	$process(docs: Document[]): Document[] {
-		let docDictionary = docs.reduce((acc, doc) => ({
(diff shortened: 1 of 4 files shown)


tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Adds a link tag ({@link }) around daffodil models, classes, etc.
+ */
+export class AddLinkTagToDaffodilReferencesProcessor implements Processor {
+	name = 'addLinkTagToDaffodilReferences';
+	$runAfter = ['docs-processed'];
+	$runBefore = ['rendering-docs'];
+
+	$process(docs: Document[]): Document[] {
+		return docs.map(doc => {
+			doc.typeParams = doc.typeParams?.slice(1, doc.typeParams.length-1).replace(/(Daff([A-Z]|[a-z])*)/g, '{@link $1}');
```
griest024, line 13: Are all the interfaces necessarily going to begin with `Daff`? I think no, for example `T extends MagentoProduct`. I guess the others will be added in a different processor later?

lderrickable: I think that everything that will be part of daffodil's official public api will start with `Daff`. We had a conversation a few weeks ago about possibly removing the `MagentoProduct` etc. from the api, but you pointed out that the Magento factories we made wouldn't work. But I still think these platform specific models shouldn't be included in our docs, so instead of adding links to platform specific models later, I plan on excluding them from the docs entirely.

griest024: there are still a few edge cases like `Dict`. Something to keep in mind but it shouldn't hold up this PR

griest024 (approved):

tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
damienwebdev, line 13: I think the better way to do this is to limit exposure on a case-by-case basis with `@docs-private`, so that we throw out the docs we don't want to showcase in the public API. 

By doing this, this simply becomes a doc lookup by doc name. 

This gives us two benefits:

1. Better alerting if a doc doesn't exist for a type.
2. It's not dependent on doc name like `Daff` and therefore `Dict` would work.

damienwebdev (changes requested):

tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
griest024: So you mean build a list of type names from `docs` and look for the name in there?

lderrickable: How do you tell the difference between `Dict` and `Dictionary`? Or `DaffProduct` from `boolean`? Maybe I just don't understand what you mean. It sounds like you're saying this should pick up everything that isn't marked as @docs-private, but we can't add any tags to the member.type definitions in the code itself. I'm pretty sure that needs to be added by dgeni. So if we don't tell dgeni exactly what to pick up (like things prefixed with `Daff`), how can we keep dgeni from creating a link for everything (like `boolean`)?

lderrickable: I think that would work with the member types, but we would need to add A LOT of extra documentation code. I'm not sure it would work very well with the type parameters. For members, it would need to be a new tag that is strictly for the type of the doc.member like this:
```
export class SomeClass {
    /**
     * Some member doc
     * @exclude-type-link
     * /
     someBooleanMember: boolean;
```

So we'd need to add that on everything that we don't want to link, which is a lot of things. Is this what you meant? Or were you thinking of something else?

damienwebdev: I think I misunderstood what this code was doing. This runs before the link processor, not instead of. So, anything with `Daff` in it becomes a link.

griest024: I think looking for the presence of the type in `docs` is sufficient to determine if we can link to it or not.

lderrickable: > Better alerting if a doc doesn't exist for a type.
I also don't think this would give us better alerting. It would tell us if we didn't mark a member as `@exclude-type-link` when it should be, but since the libraries themselves will fail to compile if we try to use a type that doesn't exist, the dgeni process will succeed whether or not we have good documentation for that type. For example:
```
/**
 * A well documented class
 * /
export class SomeClass {
    /**
     * Some member doc
     * /
     someDaffMember: DaffType;
}

export class DaffType {}
```
The DaffType needs to be defined somewhere, or the build will fail. Dgeni will pick it up and create a link for it regardless as to whether or not we have good docs for DaffType.

griest024: > Dgeni will pick it up and create a link for it regardless as to whether or not we have good docs for DaffType.

Isn't this even more reason to try and do a doc lookup instead of always creating a link for `Daff*` types?

lderrickable: That's a good point. I wonder if we should just have a specific tag for things like `Dict`, since I don't think there should be many non-Daff-prefixed models in the future.

lderrickable: > > Dgeni will pick it up and create a link for it regardless as to whether or not we have good docs for DaffType.
> 
> Isn't this even more reason to try and do a doc lookup instead of always creating a link for `Daff*` types?

No, the doc will always exist for all `Daff*` types. What I'm saying is that the doc will exist for anything defined in the daffodil libs. But it might be blank if we haven't written documentation for it. So dgeni would alway link to a doc, but that doc wouldn't necessary have anything in it. I think the confusion is that the dgeni object for anything defined in your api is called a "doc", but that "doc" doesn't necessarily have "documentation" written for all the members inside it. The "doc" is just an object with the class name and a list of all the member names (plus the documentation of that class and members if it is there).

lderrickable: > I think looking for the presence of the type in `docs` is sufficient to determine if we can link to it or not.

I think this might be right. That would solve the `Dict` problem. I'll give it a try

griest024: > dgeni would alway link to a doc, but that doc wouldn't necessary have anything in it.

I think that's okay. As long as the links generated don't 404, I think its okay to link to incomplete docs.

lderrickable: > > dgeni would alway link to a doc, but that doc wouldn't necessary have anything in it.
> 
> I think that's okay. As long as the links generated don't 404, I think its okay to link to incomplete docs.

I agree. I was just saying that this will not give us "Better alerting if a doc doesn't exist for a type."

griest024: Maybe if that doc lookup fails we just log something to console? Not really an alert but better than nothing.

lderrickable: I think the idea would be for many lookups to fail, because I'd need to lookup every type as if it might be a doc (so stuff like "boolean"). And if the lookup fails, then that's fine, we just don't add a link tag for it. Probably "fail" is the wrong word.

griest024: yeah you're right, that stuff isn't worth logging

lderrickable: > I think looking for the presence of the type in docs is sufficient to determine if we can link to it or not.

Change to this

griest024 (changes requested):

tools/dgeni/src/processors/addLinkTagToDaffodilReferences.ts
```diff
@@ -0,0 +1,48 @@
...
+			[doc.name]: true
+		}), {});
+
+		
+		return docs.map(doc => {
+			return this.addLinksToMemberTypes(
+				this.addLinksToTypeParams(doc, docDictionary), 
+				docDictionary
+			);
+		});
```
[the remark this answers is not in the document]
lderrickable: changed

griest024 (approved):

damienwebdev (approved):

Issue #1691, opened 2021-07-15
[FEAT] Injectable Magento Extra Product Preview GraphQL fragment


griest024:
# :bulb: Feature request

## Feature Name

Magento Product Preview GraphQL fragment

## The Desired Behavior

I want to be able to differentiate between the full product fragment and the product preview fragment.

## Your Use Case

When related products is an external package, it needs to include the injected product preview fragments in its provided product fragment without an infinite recursion.

PR #1692, opened 2021-07-15
feat(product): add product preview fragment injection token to magent…

griest024:
…o driver

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #1691

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/product/driver/magento/src/injection-tokens/fragments/product-preview.token.spec.ts
+++ b/libs/product/driver/magento/src/injection-tokens/fragments/product-preview.token.spec.ts
@@ -1,0 +1,43 @@
+import { TestBed } from '@angular/core/testing';
+import { gql } from 'apollo-angular';
+import { DocumentNode } from 'graphql';
+
+import {
+  daffProvideProductMagentoExtraProductPreviewFragments,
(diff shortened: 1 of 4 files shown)


lderrickable:
I was just thinking that we needed this, so nice. I think it needs to be passed down into the preview fragment though. Otherwise, it won't be on upsell or related products

griest024:
> I think it needs to be passed down into the preview fragment though. Otherwise, it won't be on upsell or related products

Once related is external, it will handle that passing down. Doing it here would complicate the implementation somewhat (because the fragments would have to be wrapped in functions) and just be removed soon so I omitted that part.

lderrickable:
> > I think it needs to be passed down into the preview fragment though. Otherwise, it won't be on upsell or related products
> 
> Once related is external, it will handle that passing down. Doing it here would complicate the implementation somewhat (because the fragments would have to be wrapped in functions) and just be removed soon so I omitted that part.

oh nice, that makes sense

lderrickable (approved):

PR #1707, opened 2021-07-28
feat(related-products): add reducers and meta-reducers

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?

- Adds reducer for storing related product IDs
- Adds meta reducer for deduping related products from main product state

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/related-products/state/src/reducers/related-products/reducer.ts
+++ b/libs/related-products/state/src/reducers/related-products/reducer.ts
@@ -1,0 +1,35 @@
+import { DaffProduct } from '@daffodil/product';
+import {
+  DaffProductPageActions,
+  DaffProductPageActionTypes,
+} from '@daffodil/product/state';
+
+import { DaffRelatedProductsReducerState } from './reducer-state.interface';
+
+/**
+ * Initial values of the related product state.
+ */
+export const initialState: DaffRelatedProductsReducerState = {
+  relatedProductIds: [],
+};
+
+/**
+ * Reducer function that catches actions and changes/overwrites related product state.
+ *
+ * @param state current State of the redux store
+ * @param action a product action
+ * @returns product state
+ */
+export function daffRelatedProductsReducer<T extends DaffProduct>(state = initialState, action: DaffProductPageActions<T>): DaffRelatedProductsReducerState {
+  switch (action.type) {
+    case DaffProductPageActionTypes.ProductPageLoadSuccessAction:
+      const product = action.payload.products.filter(({ id }) => id === action.payload.id)[0];
+
+      return {
+        ...state,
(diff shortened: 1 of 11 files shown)


libs/related-products/state/src/state.module.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * A module that provides the default reducers for the related product redux state.
+ */
+@NgModule({
+  imports: [
+    StoreModule.forFeature(DAFF_RELATED_PRODUCTS_STORE_FEATURE_KEY, daffRelatedProductsReducers),
+  ],
+  providers: [
+    ...daffProductProvideMetaReducers(daffRelatedProductsDedupeMetaReducer),
```
lderrickable, line 20: I think we should continue keeping everything compatible with Pre angular 9 for now, so this would need to manually provide

libs/related-products/state/src/reducers/related-products/reducer.ts
```diff
@@ -0,0 +1,31 @@
...
+ * Reducer function that catches actions and changes/overwrites related product state.
+ */
+export function daffRelatedProductsReducer<T extends DaffProduct>(state = initialState, action: DaffProductPageActions<T>): DaffRelatedProductsReducerState {
+  switch (action.type) {
+    case DaffProductPageActionTypes.ProductPageLoadSuccessAction:
+      const product = action.payload.products.filter(({ id }) => id === action.payload.id)[0];
+
+      return {
+        ...state,
+        relatedProductIds: product.related?.map(({ id }) => id) || [],
```
lderrickable, line 26: won't the related products already be removed from the base product at this point, since the meta reducers run before this reducer?

lderrickable (changes requested):

libs/related-products/state/src/reducers/related-products/reducer.ts
griest024: the meta reducer is immutable (see the test case) and only the main product reducers should be affected by the provided meta reducers

libs/related-products/state/src/state.module.ts
griest024: changed

libs/related-products/state/src/reducers/related-products/reducer.ts
lderrickable: > the meta reducer is immutable
But the related product reducer uses the action to set its state, not existing state in `@daffodil/product`.

> only the main product reducers should be affected by the provided meta reducers
Ok, that makes sense

lderrickable (approved):

Issue #1720, opened 2021-08-01
[BUG] Changes to `@daffodil/design` cause build failures during incremental builds


damienwebdev:
# :bug: Bug report

## Current Behavior
https://github.com/graycoreio/daffodil/pull/1717
https://github.com/graycoreio/daffodil/pull/1716

## Expected Behavior
These builds should not fail.

## Minimal reproduction of the problem with instructions
Check the build steps for https://github.com/graycoreio/daffodil/pull/1716

## What is the motivation / use case for changing the behavior?
Passing builds.

## Why might this be occurring?

So, after reviewing the build/test graph, the CI runs like this:

```txt
├── Build @daffodil/design
│   ├── Build @daffodil/branding
│   |    ├── Build @daffodil/daffio
│   └── Build @daffodil/demo
├── Test @daffodil/design
│   ├── Test @daffodil/branding
│   |    ├── Test @daffodil/daffio <-- Problem
│   |    ├── Test @daffodil/design-land <-- Problem
│   └── Test @daffodil/demo
```

The builds error for two reasons:

1. Timeouts from parallelized `ngcc` when `@daffodil/daffio` and `@daffodil/design-land` are run at the same time.
2. Building a package (e.g. `@daffodil/design`), running ngcc, rebuilding the same package WITHOUT clearing away the contents, and then running `ngcc` again.

We can fix this a few ways:

1. Fully parallelize our tests, there's no reason to await the graph as we've already built the libs necessary
2. `ngcc` immediately after builds are done to ensure that the libs are ivy-compatible
   * Note that I haven't validated that this would work, as I suspect we'd need to update all the library `tsconfig.spec.json` to use `enableIvy: true`.
3. Clear out the contents of `dist/{package}` when we build that package.

PR #1717, opened 2021-07-29
feat(design): adds docs example for DaffModalComponent

damienwebdev:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/master/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?
Currently, modal docs are embedded directly into design-land as opposed to using the example viewer. This PR moves them to the new example pattern.

Touches https://github.com/graycoreio/daffodil/issues/1388

## What is the new behavior?
This adds a new `createCustomElementFromExample` which bulds upon https://github.com/graycoreio/daffodil/pull/1712.

Currently, without providing the module injector we get an error saying `No provider for DaffModalService`.

DaffModalModule provides a service that needs to be available in the module injector as the basic example.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #1721, opened 2021-08-01
Parallelize tests

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[x] CI related changes
[ ] Other... Please describe:
```

Currently, we don't parallize our tests, which means two things:

1. When we have non-dependent apps that run their tests at the same time, we incur parallelized `ngcc` against the `dist` folder. This results in [failed builds](https://github.com/graycoreio/daffodil/pull/1717)
2. When we run tests in parallel on the same node, the tests exhibit port contention resulting in slower builds that intermittently time out.

Fixes: #1720 

## What is the new behavior?
I've now parallelized all test runs.

```
[ ] Yes
[x] No
```

## Other information


.azure/azure-pipelines.yml
```diff
@@ -50,9 +53,6 @@ stages:
 
       - job: Build_And_Test
         displayName: Build And Test
```
griest024, line 55: This should be just "build" now

.azure/azure-pipelines.yml
```diff
@@ -97,27 +88,97 @@ stages:
...
+
+        - script: npx lerna run reportcoverage --scope="$(name)"
+          displayName: Report coverage for $(name)
+
+        - script: |
+              ./cc-test-reporter sum-coverage coverage/cc.*.json
+              ./cc-test-reporter -r $token upload-coverage
+          displayName: Report Code Climate
+          env:
+            token: $(CODECLIMATE_TOKEN_DAFFODIL)
```
griest024, line 180: Can you add a TODO here for moving to its own job?

griest024 (changes requested):

.azure/azure-pipelines.yml
damienwebdev: I left the job name as "build and test" because it's a unique id and would invalidate certain GH configurations for branch protection.

griest024 (dismissed):
lgtm but I think there is a test failing

EDIT: nvm should be fixed from the rebase

griest024 (approved):

.azure/azure-pipelines.yml
```diff
@@ -49,10 +52,7 @@ stages:
 
 
       - job: Build_And_Test
```
lderrickable, line 54: I think this should be called "Build"

griest024: https://github.com/graycoreio/daffodil/pull/1721#discussion_r681743786

.azure/azure-pipelines.yml
lderrickable: There are only two of us who can merge to the protected branches, so you could just remove the requirement for the "Build and Test" job and add a requirement for the "Build" job after this is merged

lderrickable:
I think I see how the 2nd problem gets fixed, but I don't think I understand how the 1st one does:
> When we have non-dependent apps that run their tests at the same time, we incur parallelized ngcc against the dist folder.

Does running the tests on different nodes solve this problem too?

.azure/azure-pipelines.yml
damienwebdev: Agreed, but I'll do it BC as a solo commit after this is merged.

damienwebdev:
@lderrickable yes. That's correct. The process used to look like:

```
├── Build @daffodil/design
│   ├── Build @daffodil/branding
│   |    ├── Build @daffodil/daffio
│   └── Build @daffodil/demo
├── Test @daffodil/design
│   ├── Test @daffodil/branding
│   |    ├── Test @daffodil/daffio <-- ngcc
│   |    ├── Test @daffodil/design-land <-- ngcc
│   └── Test @daffodil/demo <--- ngcc
```

All on the same node. It now looks like:

```
├── Build @daffodil/design
│   ├── Build @daffodil/branding
│   |    ├── Build @daffodil/daffio
│   └── Build @daffodil/demo

Archive pre-ivy `view-engine` code and share a copy of this archive with the next processes

├── Test @daffodil/design
├── Test @daffodil/branding
├── Test @daffodil/daffio <-- ngcc
├── Test @daffodil/design-land <-- ngcc
├──@daffodil/demo <--- ngcc
```

.azure/azure-pipelines.yml
lderrickable: what does BC mean?

lderrickable (approved):

.azure/azure-pipelines.yml
damienwebdev: Backward Compatible. This PR doesn't alter any configuration changes in Github. A future PR can make that backwards-incompatible alteration.

lderrickable: Oh, that's what backwards compatible means.

PR #1747, opened 2021-08-16
feat(category): add extra preview fragments to category products query

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The category get products query does not spread in extra product preview fragments. Optional product features that provide to extra product preview fragments (composite, configurable) will therefore be missing from the category products response.

## What is the new behavior?
The extra product preview fragments are spread in and the category products response includes all the info needed to render product preview cards.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/category/driver/magento/src/category.service.ts
+++ b/libs/category/driver/magento/src/category.service.ts
@@ -3,6 +3,7 @@ import {
   Inject,
 } from '@angular/core';
 import { Apollo } from 'apollo-angular';
+import { DocumentNode } from 'graphql';
 import {
   Observable,
   combineLatest,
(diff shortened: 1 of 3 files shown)


libs/category/driver/magento/src/category.service.ts
```diff
@@ -23,6 +24,7 @@ import {
 import { DaffCategoryServiceInterface } from '@daffodil/category/driver';
 import {
   DaffProductMagentoDriverConfig,
+  DAFF_PRODUCT_MAGENTO_EXTRA_PRODUCT_PREVIEW_FRAGMENTS,
```
damienwebdev, line 27: I don't think that this should be the same fragment token for the extension point. These two features are used differently and therefore need to be targeted separately.

damienwebdev (changes requested):

libs/category/driver/magento/src/category.service.ts
griest024: So category would have its own token and composite and configurable would depend on category?

damienwebdev (approved):
lgtm

libs/category/driver/magento/src/category.service.ts
damienwebdev: I misunderstood the intent here -- and I think this is a failing of docs for `@daffodil/product/driver/magento`. I'll add an issue clarifying this.

PR #1944, opened 2021-12-31
ci: skip test runs when they are not necessary

damienwebdev:
I noticed that https://github.com/graycoreio/daffodil/pull/1943/checks?check_run_id=4674581199 incurred a test run when it should not have.

### DIFF
--- a/.azure/azure-pipelines.yml
+++ b/.azure/azure-pipelines.yml
@@ -127,7 +127,7 @@ stages:
           name: packages
       - job: test
         displayName: "Test"
-        condition: and(succeeded(), ne(length(dependencies.compute_packages.outputs['packages.packagesToTest']), 0))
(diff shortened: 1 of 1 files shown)


griest024 (approved):

.azure/azure-pipelines.yml
```diff
@@ -127,6 +127,7 @@ stages:
           name: packages
       - job: test
         displayName: "Test"
+        condition: and(succeeded(), ne(length(dependencies.compute_packages.outputs['packages.packagesToTest'] ]), 0))
```
griest024, line 130: you didn't remove the closing bracket

griest024 (changes requested):

griest024 (changes requested):
The upload code coverage job should probably get the same condition. It doesn't make sense to publish code coverage when no tests were run.

griest024:
the test job is still running

griest024 (approved):

Issue #1998, opened 2022-02-23
[FEAT] Set Product Composite Config from URL Query Param


griest024:
# :bulb: Feature request

## Feature Name

Set Composite Config from URL Query Param

## The Desired Behavior

I want to be able to configure a composite product page via a query param in the URL

## Your Use Case

As a user, I would like to be able to visit a URL like `myapp.com/myproduct?composite_config=<base64 encoded config>`
As an app developer, I want to be able to configure the query param that holds this functionality.

## Prior Work

The value of the query param should decode to an object of the form: `{itemId: [optionId]}`

griest024:
@damienwebdev do you think we should add a selector for converting the current selection to a base64 selection? This would help apps show a "bookmark or share this collection with this link:" message.

PR #2001, opened 2022-02-24
fix(product-composite): preselect item options based on configurable query param

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #1998

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/product-composite/routing/src/effects/product-page.effects.ts
+++ b/libs/product-composite/routing/src/effects/product-page.effects.ts
@@ -25,6 +25,8 @@ import {
 } from '@daffodil/product';
 import {
   DaffCompositeProduct,
+  DaffCompositeProductItem,
+  DaffCompositeProductItemOption,
   DaffProductCompositeSelectionPayload,
 } from '@daffodil/product-composite';
 import { DaffCompositeProductApplyOption } from '@daffodil/product-composite/state';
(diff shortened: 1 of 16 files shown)


libs/product-composite/routing/src/effects/product-page.effects.ts
```diff
@@ -0,0 +1,100 @@
...
+
+      let selection: DaffProductCompositeSelectionPayload;
+
+      try {
+        selection = this.config.compositeSelectionQueryParamDecode(queryParam, product);
+      } catch (error) {
+        return EMPTY;
+      }
+
+      return of(...Object.keys(selection).reduce<DaffCompositeProductApplyOption<T>[]>(
```
damienwebdev, line 87: Can you extract this into a function and clarify it? I'm having a pretty difficult time understanding what this is doing.

damienwebdev (changes requested):

libs/product-composite/routing/src/effects/product-page.effects.ts
griest024: I've changed the implementation: using `URLSearchParams`, extracted some logic into separate function, optimized some array traversals

libs/product-composite/routing/src/effects/product-page.effects.ts
```diff
@@ -39,39 +41,51 @@ import {
...
-// TODO: replace with URLSearchParams when IE support is dropped
-const getQueryParams = (path: string): Record<string, string> => {
-  const params = {};
-  path.replace(
-    /[?&]+([^=&]+)=([^&]*)/gi,
-    (_, key, value) => params[key] = decodeURIComponent(value),
-  );
-  return params;
-};
+function buildApplyActions<T extends DaffCompositeProduct = DaffCompositeProduct>(product: DaffCompositeProduct, itemId: DaffCompositeProductItem['id'], selectedOptions: DaffCompositeProductItemOption['id'][]): DaffCompositeProductApplyOption<T>[] {
```
damienwebdev, line 46: Can this function be a lambda? Can you write a test covering what happens when the incoming URL is a legitimate base64 object but it doesn't match one of the items/selections?

damienwebdev (changes requested):

libs/product-composite/routing/src/effects/product-page.effects.ts
griest024: it can be but its more performant and expressive as it is, IMO. I've added that test case

PR #2186, opened 2022-09-12
docs(tools-dgeni,daffio): handle case-sensitive file names

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, we have some broken links in daff.io as a result of "shared-names" as a result of file-name clobbering between file names that only differ in capital letters.

Fixes: N/A

## What is the new behavior?
Previously these files looked like MyFile. They now look like m_yf_ile. This was based upon a behavior that @griest024 and I saw in the Angular docs.

```
[ ] Yes
[x] No
```

### DIFF
--- a/apps/daffio/src/app/docs/services/docs.service.ts
+++ b/apps/daffio/src/app/docs/services/docs.service.ts
@@ -3,14 +3,13 @@ import {
   Inject,
   Injectable,
 } from '@angular/core';
+import { crossOsFilename } from '@daffodil/docs-utils';
 import { Observable } from 'rxjs';
 
 import { DaffioDoc } from '../models/doc';
 import { DaffioGuideList } from '../models/guide-list';
 import { DAFFIO_DOCS_PATH_TOKEN } from './docs-path.token';
 import { DaffioDocsServiceInterface } from './docs-service.interface';
-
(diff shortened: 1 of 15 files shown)


apps/daffio/src/app/docs/services/docs.service.ts
```diff
@@ -10,6 +10,8 @@ import { DaffioGuideList } from '../models/guide-list';
 import { DAFFIO_DOCS_PATH_TOKEN } from './docs-path.token';
 import { DaffioDocsServiceInterface } from './docs-service.interface';
 
+export const computeCrossEnvSafePath = (path: string) => path.replace(/[A-Z]/g, (a) => a.toLocaleLowerCase().concat('_'));
```
griest024, line 13: Is there a good central place to export this so the logic isn't duplicated here and in dgeni?

griest024 (changes requested):

apps/daffio/src/app/docs/services/docs.service.ts
damienwebdev: Moved to a new `docs-utils` package.

griest024 (approved):

PR #2254, opened 2022-11-29
feat(forms): add `@daffodil/forms` and `daffFormNotEmptyValidator`

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Adds a validator for checking if a form has any field filled out.

```
[ ] Yes
[x] No
```

## Other information
This is useful for updates that operate on partials. Any single field can make the whole form valid but there must be at least one valid field.

### DIFF
--- a/libs/forms/src/validators/not-empty.ts
+++ b/libs/forms/src/validators/not-empty.ts
@@ -6,7 +6,8 @@ import {
 } from '@angular/forms';
 
 /**
- * A validator for a form group that returns an error if all of its immediate children don't have a truthy value.
+ * A form validator that passes validation when the Form control or
+ * at least one of the immediate children of a Form group have a value.
  * Empty strings are considered falsy.
  */
 // TODO: handle nested form groups
--- a/libs/forms/src/validators/not-empty.ts
+++ b/libs/forms/src/validators/not-empty.ts
@@ -1,0 +1,22 @@
+import {
+  AbstractControl,
+  FormGroup,
+  ValidationErrors,
+  ValidatorFn,
+} from '@angular/forms';
+
+/**
+ * A validator for a form group that returns an error if all of its immediate children don't have a truthy value.
+ * Empty strings are considered falsy.
+ */
+// TODO: handle nested form groups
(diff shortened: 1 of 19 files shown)


griest024:
put in new forms package

griest024:
@damienwebdev moved to forms package and improved to handle form controls as well as groups

libs/forms/src/validators/not-empty.ts
```diff
@@ -0,0 +1,22 @@
...
+  FormGroup,
+  ValidationErrors,
+  ValidatorFn,
+} from '@angular/forms';
+
+/**
+ * A validator for a form group that returns an error if all of its immediate children don't have a truthy value.
+ * Empty strings are considered falsy.
+ */
+// TODO: handle nested form groups
```
damienwebdev, line 12: Is this Todo still correct?

libs/forms/src/validators/not-empty.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  AbstractControl,
+  FormGroup,
+  ValidationErrors,
+  ValidatorFn,
+} from '@angular/forms';
+
+/**
+ * A validator for a form group that returns an error if all of its immediate children don't have a truthy value.
```
damienwebdev, line 9: This doesn't return an error.

What do you think about:

"A form validator that passes validation when the Form control or at least one of the immediate children of a Form group have a value."

damienwebdev (changes requested):

libs/forms/src/validators/not-empty.ts
griest024: yeah I think ideally this validator would handle nested form groups. Do you think it should be implemented for this PR? I don't have an immediate use case for it.

libs/forms/src/validators/not-empty.ts
griest024: changed

PR #2256, opened 2022-12-02
feat(core): add operation state

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Adds state, adapter, selectors, and facade for a basic operation state. This can be used as the basis for most driver-oriented state packages, significantly DRYing up the codebase.

```
[ ] Yes
[x] No
```

## Other information


libs/core/state/src/operation/facade.ts
```diff
@@ -0,0 +1,75 @@
...
+   * Whether the operation state has any errors.
+   * If so, it should be considered to be in an "error" state.
+   */
+  hasErrors$: Observable<boolean>;
+}
+
+/**
+ * @inheritdoc
+ */
+export abstract class DaffOperationStateFacade<
```
damienwebdev, line 48: The abstract class here bloats the API, imo.

The interface itself is sound, but the rest of the code exists solely to reduce code but at the cost of encouraging inheritance, which I think will make usages more brittle.

I think we keep the interface, but delete the abstract class.

libs/core/state/src/operation/selectors.ts
```diff
@@ -0,0 +1,88 @@
...
+   */
+  selectHasErrors: MemoizedSelector<TRootState, boolean>;
+}
+
+/**
+ * Creates a set of selectors for an operation state.
+ *
+ * @param selectState The feature selector for the operation state.
+ */
+export const daffOperationStateSelectorFactory = <
```
damienwebdev, line 49: This is a dev time function only, it only provides organizational use to us. Is there a way to leverage the compiler's code folding feature to remove this function before runtime?

libs/core/state/src/operation/state.ts
```diff
@@ -0,0 +1,23 @@
...
+   */
+  loading: DaffState;
+
+  /**
+   * The errors generated by the most recent operation.
+   */
+  errors: DaffStateError[];
+}
+
+export const daffOperationInitialState: DaffOperationState = {
```
damienwebdev, line 20: Same idea here with code folding, this is purely organizational code.

damienwebdev (changes requested):
I'm convinced that the DRYing that this PR does is good, but I'm not convinced in the implementation.

I have a feeling that we should be able to abuse code folding https://angular.io/guide/aot-compiler#code-folding to be able to limit the runtime performance impact of some of this code.

libs/core/state/src/operation/selectors.ts
griest024: the selector factory is the entrypoint into the selector portion of this feature. Are you suggesting that I memoize this factory and only export the memoized version?

libs/core/state/src/operation/state.ts
griest024: this is used in tests and does have a use case in consuming code. If I were to unexport this, I would just delete the token entirely and inline the code manually.

libs/core/state/src/operation/adapter.ts
```diff
@@ -0,0 +1,52 @@
+import { DaffState } from '../states/public_api';
+import { DaffOperationState } from './state';
+
+/**
+ * An adapter for modifying {@link DaffOperationState}.
+ */
+export class DaffOperationStateAdapter<T extends DaffOperationState = DaffOperationState> {
```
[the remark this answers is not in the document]
griest024: agreed, removed

libs/core/state/src/operation/facade.ts
griest024: agreed, removed

libs/core/state/src/operation/selectors.ts
griest024: changed to named function

PR #2349, opened 2023-03-05
 feat(contact): strengthen contact driver types

damienwebdev:
## PR Checklist

Previously, the `DaffContactServiceInterface` was extremely weak in both request and response.
    
This hardens the type, ironing down the possible send arguments and eliminating the genericity of the response type.
    
BREAKING CHANGE: The contact driver type changed, if you're providing an implementation, please see the new `DaffContactServiceInterface`. If you had a use-case for a generic response type, I would love to hear your feedback.

Note for @griest024 this has two other diffs that are nice and this pr should be rebase merged.

### DIFF
--- a/libs/contact/state/src/actions/contact.actions.ts
+++ b/libs/contact/state/src/actions/contact.actions.ts
@@ -1,5 +1,6 @@
 import { Action } from '@ngrx/store';
 
+import { DaffContactRequest } from '@daffodil/contact/driver';
 import { DaffStateError } from '@daffodil/core/state';
 
 export enum DaffContactActionTypes {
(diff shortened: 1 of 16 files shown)


libs/contact/state/src/actions/contact.actions.ts
```diff
@@ -37,7 +38,7 @@ export class DaffContactReset implements Action {
   readonly type = DaffContactActionTypes.ContactResetAction;
 }
 
-export type DaffContactActions<T> =
+export type DaffContactActions<T extends DaffContactRequest> =
```
griest024, line 41: you should set a default type param here

griest024 (changes requested):

PR #2419, opened 2023-05-07
ci: move build and tests to Github actions 

damienwebdev:
ci: move build and tests to Github actions

This PR makes several improvements upon the existing Azure Pipelines-based
build pipeline. First-and-foremost, we now:

1. Replace Azure Pipelines with Github actions
1. Leverage the `use-stamp-cache` of `setup-node`, lowering the `npm ci`
    install time substantially.
2. Leverage Build-caching as a result of `lerna 6/nx`

Secondarily, as a result of a build, we used to publish the `dist`
folder as a distributable. We currently no longer publish this*.

Additionally, share the old `dist` archive to the tests right before they ran.
You might think that now as a result of deleting it, we either rebuild
or the tests simply error. Instead, we now publish the `nx_cache`
as a cache during the build. Then, when the tests run, we restore from the cache,
causing `nx` to cache hit and rebuild the tested packages dependencies
packages near-instantly, resulting in the same output into the `dist`
folder upon re-build right before the tests run.

Interestingly, this process exposed a few bugs where a packages' tests
were improperly depending upon packages that they were not
supposed to depend upon, or were stricitly missing dependencies
in package.json.

* That archive may be restored under a different name in the future.

Co-authored-by: Hanmin Kim <hanminkim123@gmail.com>


damienwebdev:
Closed via: 576c0167bfe7bbee2c9d92f0be5c1d1fa86cdc01

PR #2776, opened 2024-04-26
feat(design): add immutable tree transform

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, we only have `daffTransformTreeInPlace` which does pre-order, right-to-left, in-place transformation. This isn't always pleasant to work with and often causes side-effects.

Fixes: N/A

## What is the new behavior?
There is a new `daffTransformTree` which is immutable and pre-order, left-to-right, which is likely more expected.

```
[ ] Yes
[x] No
```

### DIFF
--- a/libs/design/tree/src/utils/transform.ts
+++ b/libs/design/tree/src/utils/transform.ts
@@ -1,0 +1,40 @@
+import { RecursiveTreeKeyOfType } from '../interfaces/recursive-key';
+import { DaffTreeData } from '../interfaces/tree-data';
+
+/**
+ * Transform a tree-like structure into a {@link DaffTreeData}.
+ * Traverses the tree pre-order, left to right.
+ *
+ * @param tree - The data structure representing tree-like data.
+ * @param transformFn - A user-supplied function that will transform the user
+ *  type into a {@link DaffTreeData}
+ * @param key - The property of the your tree that indicates which
+ *  key contains the "children" of your tree structure.
+ *
+ */
+export const daffTransformTree = <
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  T extends Record<any,any>,
+>(
+  tree: T,
+  transformFn: (type: T) => DaffTreeData<unknown>,
+  key: RecursiveTreeKeyOfType<T>,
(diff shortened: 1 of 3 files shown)


libs/design/tree/src/utils/transform.ts
```diff
@@ -0,0 +1,39 @@
...
+    const childItems = node[key];
+    for (const child of <T[]>childItems) {
+      const transformedChild: DaffTreeData<unknown> = transformFn(child);
+      parent.items.push(transformedChild);
+      queue.push({ node: child, parent: transformedChild });
+    }
+  }
+
+  return transformedTree;
+};
```
griest024: All returns of this function are `DaffTreeData<unknown>`. I think a return type param should be added. Its unlikely to ever need to be explicitly provided as TS should infer it from the passed transform arg. It improves the DX of this function IMO.

```suggestion
export const daffTransformTree = <
  // eslint-disable-next-line @typescript-eslint/ban-types
  T extends Record<any,any>,
  R = unknown
>(
  tree: T,
  transformFn: (type: T) => DaffTreeData<R>,
  key: RecursiveTreeKeyOfType<T>,
): DaffTreeData<R> => {

  const transformedTree: DaffTreeData<R> = transformFn(tree);

  const queue: { node: T; parent: DaffTreeData<R> }[] = [{ node: tree, parent: transformedTree }];

  while (queue.length > 0) {
    const { node, parent } = queue.shift();

    const childItems = node[key];
    for (const child of <T[]>childItems) {
      const transformedChild: DaffTreeData<R> = transformFn(child);
      parent.items.push(transformedChild);
      queue.push({ node: child, parent: transformedChild });
    }
  }

  return transformedTree;
};
```

griest024 (changes requested):

libs/design/tree/src/utils/transform.ts
damienwebdev: Agreed! bf2b1ae519ecc59ddb61e3af5bef89ad25eaa81b

PR #2777, opened 2024-04-26
feat(design)!: add tree renderMode

damienwebdev:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Previously, "hidden" child nodes were entirely removed from the document. 

Fixes: N/A

## What is the new behavior?
This introduces a new Input on DaffTreeComponent called "renderMode" that accepts "not-in-dom" or "in-dom", with "in-dom" being the default.

- "in-dom" - The nodes are now retained in the document, but are hidden via css. 
- "not-in-dom" - The nodes removed from the document. (The old behavior.)

```
[x] Yes
[ ] No
```

BREAKING CHANGE: Previously, the now-named renderMode "not-in-dom" was the default. This is no longer the case. If one was rendering immensely large trees and you now notice a significant slow-down, you can change the renderMode like:

```ts
<ul daff-tree ... renderMode="not-in-dom">
 ...
</ul>
```

Which will restore the original behavior.

## Other information


libs/design/tree/src/tree/tree.component.scss
```diff
@@ -12,6 +12,10 @@
   --tree-padding: 16px;
 }
 
+.hidden {
```
xelaint: needs to be nested inside `.daff-tree`

libs/design/tree/src/interfaces/tree-mode.ts
```diff
@@ -0,0 +1,6 @@
+/**
```
xelaint: Can you rename this file to `tree-render-mode.ts` ?

xelaint (changes requested):
Not all `<li>` have the hidden class set. Seems like it's only not working for level-1 `<li>`s

Screen Shot 2024-04-26 at 2 39 59 PM

libs/design/tree/src/tree/tree.component.scss
damienwebdev: f644cf0b2

libs/design/tree/src/interfaces/tree-mode.ts
damienwebdev: dd11deb72

damienwebdev:
@xelaint 

> Not all `<li>` have the hidden class set. Seems like it's only not working for level-1 `<li>`s

That's by intent.

The idea is that for a tree like:

image

We currently render:

```html
<li level="1">@daffodil/analytics</li>
<li level="1">analytics-provider-data-layer</li>
<li level="1">@daffodil/auth</li>
<li level="1">@daffodil/authorize.net</li>
```

We now render:

```html
<li level="1">@daffodil/analytics</li>
<li level="2" class="hidden">Overview</li>
<li level="2" class="hidden">Configuration</li>
<li level="1">analytics-provider-data-layer</li>
<li level="1">@daffodil/auth</li>
<li level="1">@daffodil/authorize.net</li>
```

The hidden elements remain rendered in the DOM, they just aren't visible.

The "level-1"'s are essentially always visible because they're the "top" level that needs to be interactable to open the tree.

xelaint (approved):

Issue #763, opened 2020-04-03
[FEAT] Normalize input interfaces and attributes for design components


xelaint:
# :bulb: Feature request

## Feature Name
Normalize input interfaces and attributes for design components.

## The Desired Behavior
Create a list of input interfaces and attributes that will be used across the board in @daffodil/design.

### Properties to AVOID (too similar to CSS properties):
- align
- alignment
- horizontalAlignment
- fontSize
- fontFamily

**Exceptions:**
- color

### Known Valid Properties:
- `opened: boolean;`
- `fullscreen: boolean;`
- `size: xs | sm | md | lg | xl`

**Size (global interface):**
Not all sizes need to be exported and used.
```
export type DaffSizeSmallType = 'sm'
export type DaffSizeMediumType = 'md'
export type DaffSizeLargeType = 'lg'
export type DaffSizeXLargeType = 'xl'

export type DaffSizeAllType = DaffSizeXSmallType | DaffSizeSmallType | DaffSizeMediumType | DaffSizeLargeType | DaffSizeXLargeType
```

### Global Interfaces to Create
- [x] `size`
- [x] `skeleton: boolean;` | `DaffSkeletonble`
- [x] `mutating: boolean;` | `DaffMutable`
- [x] `DaffCompactable`
- [ ] `DaffOpenable`
- [ ] `DaffSelectable`
- [x] `Disablable` and `Focusable` interfaces for `daff-form-field`

### Card Types
- [x] `daff-card`
- [x] `daff-raised-card`
- [x] `daff-stroked-card`

### Link Set
- Basic & stylized variants (names TBD, after components are updated to reflect correct style)

### Feature
- [x] Deprecate DaffFeatureComponent

### Need to Deprecate
- [x] DaffHeroLayout
- [x] DaffHeroSize (size will be a global interface)
- [x] DaffCalloutLayout
- [x] DaffCalloutSize (size will be a global interface)
- [x] DaffListAppearance

### `center` type? **Fixed with textAlignment** (#1502)
- We need a way to provide centering option (horizontally). The `inline-block` update to hero/callout did not work.

## Your Use Case
As a developer, I'd like to maintain consistency in defining input interfaces and attributes throughout the design package.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

xelaint:
[...]

PR #2908, opened 2024-07-12
feat(design): create `DaffOpenableDirective`

### LINKED ISSUE #3020: [FEAT] `DaffOpenable` directive
# :bulb: Feature request

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

xelaint:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Part of: #763 
Fixes: #3020

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/design/src/core/openable/openable.directive.ts
+++ b/libs/design/src/core/openable/openable.directive.ts
@@ -3,16 +3,20 @@ import {
   EventEmitter,
   HostBinding,
   Input,
+  OnChanges,
   Output,
+  SimpleChanges,
+  isDevMode,
 } from '@angular/core';
 
 import { DaffOpenable } from './openable';
+import { DaffOpenableStateError } from './utils/state-error';
 
 /**
  * A directive that opens or closes a component. It should only be used as an [Angular Host Directive](https://angular.dev/guide/directives/directive-composition-api).
- * It supports both a state and stateless implementation, but only one version should be used within a component.
+ * This directive is stateless by default, but it supports both a state and stateless implementation. Only one version should be used within a component.
  *
- * @usage
+ * ## Example
  * ```ts
  * import {
  *  Component,
@@ -58,20 +62,25 @@ import { DaffOpenable } from './openable';
   standalone: true,
 })
 
-export class DaffOpenableDirective implements DaffOpenable {
+export class DaffOpenableDirective implements DaffOpenable, OnChanges {
   /** Whether or not a component implementing the directive is open */
   @Input() @HostBinding('class.daff-open') open = false;
 
+  private _setOpen(v: boolean) {
+    if(!this.stateless) {
+      this.open = v;
+    }
+  }
+
   /** Whether or not a component should handle state
    *
-   * @usage
    * ```ts
    * constructor(private openDirective: DaffOpenableDirective) {
-   *  this.openDirective.stateless = true;
+   *  this.openDirective.stateless = false;
    * }
    * ```
    */
-  stateless = false;
+  stateless = true;
 
   /**
    * Event fired when a component is opened (true) or closed (false)
@@ -82,9 +91,7 @@ export class DaffOpenableDirective implements DaffOpenable {
    * Open the component
    */
   reveal() {
-    if(!this.stateless) {
-      this.open = true;
-    }
+    this._setOpen(true);
 
     this.toggled.emit(true);
   }
@@ -93,9 +100,7 @@ export class DaffOpenableDirective implements DaffOpenable {
    * Close the component
    */
   hide() {
-    if(!this.stateless) {
-      this.open = false;
-    }
+    this._setOpen(false);
 
     this.toggled.emit(false);
   }
@@ -106,10 +111,21 @@ export class DaffOpenableDirective implements DaffOpenable {
   toggle() {
     const state = !this.open;
 
-    if(!this.stateless) {
-      this.open = state;
-    }
+    this._setOpen(state);
 
     this.toggled.emit(state);
+  }
+
+  ngOnChanges(changes: SimpleChanges) {
+    /**
+     * Throw an error if open is set in a component that is not stateless
+     */
+    if(changes.open.currentValue && !this.stateless) {
+      this.open = changes.open.previousValue;
+
+      if(isDevMode()) {
+        throw new Error(DaffOpenableStateError);
+      }
+    }
   }
 }
--- a/libs/design/src/core/openable/openable.directive.ts
+++ b/libs/design/src/core/openable/openable.directive.ts
@@ -8,6 +8,51 @@ import {
 
 import { DaffOpenable } from './openable';
 
+/**
+ * A directive that opens or closes a component. It should only be used as an [Angular Host Directive](https://angular.dev/guide/directives/directive-composition-api).
+ * It supports both a state and stateless implementation, but only one version should be used within a component.
+ *
+ * @usage
+ * ```ts
+ * import {
+ *  Component,
+ *  ChangeDetectionStrategy,
+ * } from '@angular/core';
+ *
+ * import { DaffOpenableDirective } from '@daffodil/design';
+ *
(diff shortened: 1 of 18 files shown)


xelaint:
@damienwebdev Need to finalize what the output naming is for the openable directive. Are we sticking to `toggled`?

libs/design/src/core/openable/openable.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+
+@Directive({
+  selector: '[daffOpenable]',
+  standalone: true,
+})
+
+export class DaffOpenableDirective implements DaffOpenable {
+  @Input() @HostBinding('class.daff-open') open = false;
+
+  @Output() toggled: EventEmitter<boolean> = new EventEmitter<boolean>();
```
damienwebdev, line 88: What are your thoughts on `opened`? I know we originally said no to this, but I really dislike `toggled`.

libs/design/src/core/openable/openable.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * An interface for giving a component the ability to display an open UI.
+ * In order to be openable, the class must implement this property.
+ */
+export interface DaffOpenable {
+  open: boolean;
```
damienwebdev, line 6: Needs docs.

libs/design/src/core/openable/openable.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+
+import { DaffOpenable } from './openable';
+
+@Directive({
+  selector: '[daffOpenable]',
+  standalone: true,
+})
+
+export class DaffOpenableDirective implements DaffOpenable {
+  @Input() @HostBinding('class.daff-open') open = false;
```
damienwebdev, line 67: Docs.

libs/design/src/core/openable/openable.directive.ts
```diff
@@ -0,0 +1,35 @@
...
+  Output,
+} from '@angular/core';
+
+import { DaffOpenable } from './openable';
+
+@Directive({
+  selector: '[daffOpenable]',
+  standalone: true,
+})
+
```
damienwebdev, line 64: Remove the whitespace.

damienwebdev (changes requested):

libs/design/src/core/openable/openable.directive.ts
xelaint: I think we said no because this output handles opened or closed, so calling it opened is incorrect. @griest024 any thoughts on a better name?

libs/design/src/core/openable/openable.directive.ts
xelaint: Disagree. There should be an empty new line?

Screen Shot 2024-08-12 at 9 14 39 AM

libs/design/src/core/openable/openable.directive.ts
xelaint: added

libs/design/src/core/openable/openable.ts
xelaint: added

libs/design/src/core/openable/openable.directive.ts
griest024: `toggled` does seem more accurate than `opened`. The only other idea I have is `openedUpdate` or `openedChange` which I dislike more than `toggled`

xelaint:
@griest024 we made some updates according to your concerns. lmk what you think.

libs/design/src/core/openable/openable.directive.ts
griest024: honestly this could just be `changed` because the implementing component has to manually bind the outputs anyway right? So I feel like we can use a generic event name without fear of collision.

libs/design/src/core/openable/openable.directive.ts
```diff
@@ -0,0 +1,115 @@
...
+ * ```
+ */
+@Directive({
+  selector: '[daffOpenable]',
+  standalone: true,
+})
+
+export class DaffOpenableDirective implements DaffOpenable {
+  /** Whether or not a component implementing the directive is open */
+  @Input() @HostBinding('class.daff-open') open = false;
```
griest024, line 67: I think there should be an additional guard in the setting of an input to ensure that the internal state is only updated from the bound input in stateless mode.
```suggestion
  private _open = false;
  
  @Input()
  set open(v: boolean) {
    if (this.stateless) {
      this._open = v
    } else {
      // log a warning?
    }
  }
  @HostBinding('class.daff-open')
  get open(): boolean {
    return this._open
  }
```

libs/design/src/core/openable/openable.ts
```diff
@@ -0,0 +1,17 @@
...
+
+  /** Sets reveal on the component */
+  reveal: () => void;
+
+  /** Sets hide on the component */
+  hide: () => void;
+
+  /** Sets toggle on the component */
+  toggle: () => void;
+}
```
griest024: These docs aren't helpful. Especially for `toggle`, they should say something about the expected behavior of the method.

griest024 (changes requested):

libs/design/src/core/openable/openable.directive.ts
griest024: and then in the later methods, only mutate `_open` not `open`

libs/design/src/core/openable/openable.directive.ts
xelaint: decided to stick with `toggled`

libs/design/src/core/openable/openable.directive.ts
xelaint: made updates to this

libs/design/src/core/openable/openable.ts
xelaint: updated

libs/design/src/core/openable/openable.ts
```diff
@@ -1,17 +1,16 @@
...
-  /** Sets reveal on the component */
+  /** Reveal the component */
   reveal: () => void;
 
-  /** Sets hide on the component */
+  /** Hide the component */
   hide: () => void;
 
-  /** Sets toggle on the component */
+  /** Toggles the component from open to not open */
```
griest024: This will also work in reverse from not open to open, correct? The comment should reflect that converse behavior.

libs/design/src/core/openable/openable.directive.ts
```diff
@@ -58,20 +62,25 @@ import { DaffOpenable } from './openable';
...
 })
 
-export class DaffOpenableDirective implements DaffOpenable {
+export class DaffOpenableDirective implements DaffOpenable, OnChanges {
   /** Whether or not a component implementing the directive is open */
   @Input() @HostBinding('class.daff-open') open = false;
 
+  private _setOpen(v: boolean) {
+    if(!this.stateless) {
+      this.open = v;
```
griest024, line 71: this should be mutating `this._open` not `this.open`. The setter for `this.open` contains the opposite check for `stateless` so this will never do anything.

griest024 (changes requested):

libs/design/src/core/openable/openable.directive.ts
xelaint: `_open` was removed, can you re-review?

libs/design/src/core/openable/openable.ts
xelaint: updated

griest024 (dismissed):

xelaint:
@griest024 had to fix some tests

PR #2909, opened 2024-07-12
feat: render design examples in daffio

griest024:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Fixes: #2903

## What is the new behavior?
- moves the example viewer components to a `docs-components` package
- changes the example rendering from a custom element based one to a component outlet one
- adds custom element support for the example viewer container so it can be rendered when the markdown is rendered with `innerHtml`
- converts example components to standalone so that they can be dynamically rendered
- extracts internal package ignore constant in dgeni and adds new packages

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

PR #2924, opened 2024-07-19
feat(design)!: make design examples standalone

damienwebdev:
Extract from https://github.com/graycoreio/daffodil/pull/2909


libs/design/notification/examples/src/notification-with-actions/notification-with-actions.component.html
```diff
@@ -1,14 +1,19 @@
 <button daff-button (click)="toggleNotification()">
-	<ng-container *ngIf="showNotification">Show Notification</ng-container>
-	<ng-container *ngIf="!showNotification">Hide Notification</ng-container>
+  @if (showNotification) {
```
xelaint: This is backwards. It's saying "Hide Notification" when the notification example is hidden and "Show Notification" when the button is clicked.

xelaint (changes requested):

libs/design/notification/examples/src/notification-with-actions/notification-with-actions.component.html
damienwebdev: I've handled this by removing the extraneous conditions.

PR #2926, opened 2024-07-21
feat(daffio,design): highlight active tree item in sidebar

### LINKED ISSUE #2854: [FEAT] Add selected functionality for tree component
# :bulb: Feature request

## Feature Name
Add selected UI for tree component

## The Desired Behavior
A `[daff-tree-item]` that is active/selected should have a `selected` class added to it to properly style it.

## Your Use Case
As a designer, I would like to indicate to users if a tree item in a tree is currently selected/active.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #2854 

```
[ ] Yes
[ ] No
```

## Other information
The way the options must be set is unfortunate. There is no way to set the options from the `    DaffTreeItemDirective` because of https://github.com/angular/angular/issues/50700#issuecomment-1590896233

### DIFF
--- a/apps/daffio/src/app/docs/components/docs-list/docs-list.component.html
+++ b/apps/daffio/src/app/docs/components/docs-list/docs-list.component.html
@@ -5,7 +5,7 @@
     </ng-template>
 
     <ng-template #daffTreeItemTpl let-node>
-      <a [daffTreeItem]="ROUTER_LINK_ACTIVE_CONFIG" [node]="node" [routerLink]="node.url">{{ node.title }}</a>
+      <a daffTreeItem [selected]="rla.isActive" routerLinkActive [routerLinkActiveOptions]="ROUTER_LINK_ACTIVE_CONFIG" #rla="routerLinkActive" [node]="node" [routerLink]="node.url">{{ node.title }}</a>
     </ng-template>
   </ul>
 }
--- a/apps/daffio/src/app/docs/components/docs-list/docs-list.component.html
+++ b/apps/daffio/src/app/docs/components/docs-list/docs-list.component.html
@@ -5,7 +5,7 @@
     </ng-template>
 
     <ng-template #daffTreeItemTpl let-node>
-      <a [daffTreeItem]="activeRouterLinkConfiguration" [node]="node" [routerLink]="node.url">{{ node.title }}</a>
+      <a [daffTreeItem]="ROUTER_LINK_ACTIVE_CONFIG" [node]="node" [routerLink]="node.url">{{ node.title }}</a>
     </ng-template>
   </ul>
 }
--- a/apps/daffio/src/app/docs/components/docs-list/docs-list.component.html
+++ b/apps/daffio/src/app/docs/components/docs-list/docs-list.component.html
@@ -5,7 +5,7 @@
     </ng-template>
 
     <ng-template #daffTreeItemTpl let-node>
-      <a daffTreeItem [node]="node" [routerLink]="node.url">{{ node.title }}</a>
+      <a [daffTreeItem]="activeRouterLinkConfiguration" [node]="node" [routerLink]="node.url">{{ node.title }}</a>
     </ng-template>
   </ul>
 }
--- a/apps/daffio/src/app/docs/components/docs-list/docs-list.component.ts
+++ b/apps/daffio/src/app/docs/components/docs-list/docs-list.component.ts
@@ -56,7 +56,7 @@ export class DaffioDocsListComponent implements OnInit {
   }
 
   tree$: Observable<DaffTreeData<unknown>>;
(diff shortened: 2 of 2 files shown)


xelaint:
So the tree technically doesn't handle the "active" state unless `ROUTER_LINK_ACTIVE_CONFIG` is set in the implementation?

Is there a different way to do this?

griest024:
> So the tree technically doesn't handle the "active" state unless `ROUTER_LINK_ACTIVE_CONFIG` is set in the implementation?
> 
> Is there a different way to do this?

no it does but the "overview" link will be active when a child guide is selected so you'll see 2 active items at once

xelaint:
Also, getting this error in console in design-land:

```
template.html:1 ERROR Error: NG0311: Directive _RouterLinkActive does not have an input with a public name of daffTreeItem.
    at validateMappings (
```

griest024:
> Also, getting this error in console in design-land:
> 
> ```
> template.html:1 ERROR Error: NG0311: Directive _RouterLinkActive does not have an input with a public name of daffTreeItem.
>     at validateMappings (
> ```

I cannot repro that

xelaint:
> > So the tree technically doesn't handle the "active" state unless `ROUTER_LINK_ACTIVE_CONFIG` is set in the implementation?
> > Is there a different way to do this?
> 
> no it does but the "overview" link will be active when a child guide is selected so you'll see 2 active items at once

Why are the links below overview treated as child guides of overview? Shouldn't they be sibling to overview?

griest024:
@xelaint I'll use cart as an example. the cart driver guide is at `/docs/packages/cart/drivers`. the cart "overview" guide has the link `/docs/packages/cart`. The angular router considers `/docs/packages/cart/drivers` to be a "child" of `/docs/packages/cart` and therefore considers `/docs/packages/cart` to be active on the route `/docs/packages/cart/drivers` *unless* `exact: true` is set.

libs/design/tree/src/tree-item/tree-item.directive.ts
```diff
@@ -6,6 +6,7 @@ import {
   Inject,
   Input,
 } from '@angular/core';
+import { RouterLinkActive } from '@angular/router';
```
damienwebdev: This makes the `design` library dependent on `@angular/router` which I've intentionally disallowed historically.

apps/daffio/src/app/docs/components/docs-list/docs-list.component.html
```diff
@@ -5,7 +5,7 @@
     </ng-template>
 
     <ng-template #daffTreeItemTpl let-node>
-      <a daffTreeItem [node]="node" [routerLink]="node.url">{{ node.title }}</a>
+      <a [daffTreeItem]="ROUTER_LINK_ACTIVE_CONFIG" [node]="node" [routerLink]="node.url">{{ node.title }}</a>
```
damienwebdev: From a review of the directive's docs:

> <a routerLink="/user/bob" routerLinkActive #rla="routerLinkActive">
>   Bob {{ rla.isActive ? '(already open)' : ''}}
> </a>

Can you do:

> <a [selected]="rla.isActive"  #rla="routerLinkActive" [node]="node" [routerLink]="node.url">{{ node.title }}</a>

damienwebdev (changes requested):

libs/design/tree/src/tree-item/tree-item.directive.ts
griest024: Then this feature cannot be implemented as requested. The best that we can do is set `daffTreeItem routerLinkActive="selected"` in app code.

apps/daffio/src/app/docs/components/docs-list/docs-list.component.html
griest024: yes I did something like that

libs/design/tree/src/tree-item/tree-item.directive.ts
griest024: removed everything from design

damienwebdev (approved):
lgtm

xelaint:
I'd expect a parent tree item to be open if there is an active/selected tree item within it. Should this be done in this PR or create a separate issue for it?

griest024:
> I'd expect a parent tree item to be open if there is an active/selected tree item within it. Should this be done in this PR or create a separate issue for it?

separate PR

xelaint (approved):

PR #2938, opened 2024-07-23
feat(design): expand selected tree items on initial render

### LINKED ISSUE #2931: [FEAT] A parent tree item should be expanded if there is an active/selected tree item within it
# :bulb: Feature request

## Feature Name
A parent tree item should be expanded if there is an active/selected tree item within it

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #2931

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/design/tree/src/tree-item/tree-item.directive.ts
+++ b/libs/design/tree/src/tree-item/tree-item.directive.ts
@@ -1,10 +1,9 @@
-/* eslint-disable quote-props */
-import { DOCUMENT } from '@angular/common';
 import {
   Directive,
-  Inject,
   Input,
 } from '@angular/core';
+
+import { collect } from '@daffodil/core';
 
 import { DaffTreeNotifierService } from '../tree/tree-notifier.service';
 import { DaffTreeFlatNode } from '../utils/flatten-tree';
@@ -32,7 +31,7 @@ import { DaffTreeFlatNode } from '../utils/flatten-tree';
 @Directive({
   selector: '[daffTreeItem]',
   host: {
-    'class': 'daff-tree-item',
+    class: 'daff-tree-item',
     '[class.selected]': 'selected',
     '[class.parent]': 'isParent',
     '[class.open]': 'open',
@@ -99,7 +98,6 @@ export class DaffTreeItemDirective {
   @Input() selected = false;
 
   constructor(
-    @Inject(DOCUMENT) private document: any,
     private treeNotifier: DaffTreeNotifierService,
   ) {}
 
@@ -107,38 +105,46 @@ export class DaffTreeItemDirective {
    * @docs-private
    */
   onEscape() {
-    this.toggleParent(this.node);
+    this.toggleParent();
   }
 
   /**
    * @docs-private
    */
   onClick() {
-    if(this.node.hasChildren) {
-      this.toggleTree(this.node);
+    if (this.node.hasChildren) {
+      this.toggleTree();
     }
+  }
+
+  /**
+   * Opens parent and parent of parent all the way to the root of the tree.
+   */
+  openAncestors() {
+    collect(this._node._treeRef, (node) => [node.parent], this._node.level).forEach((node) => node.open = true);
     this.treeNotifier.notify();
   }
 
   /**
    * Toggle the open state of the tree's parent.
    */
-  toggleParent(node: DaffTreeFlatNode) {
-    if(node._treeRef?.parent.parent === undefined) {
+  toggleParent() {
+    if (this.node._treeRef?.parent.parent === undefined) {
       return;
     }
-    node._treeRef.parent.open = !node._treeRef.parent.open;
-    (<Document>this.document).getElementById('tree-' + node._treeRef.parent.id).focus();
+    this.node._treeRef.parent.open = !this.node._treeRef.parent.open;
+    this.treeNotifier.notify();
   }
 
   /**
    * Toggle the open state of this specific subtree tree.
    */
-  toggleTree(node: DaffTreeFlatNode) {
-    if(node._treeRef.open === false) {
-      node._treeRef.open = true;
+  toggleTree() {
+    if (this.node._treeRef.open === false) {
+      this.node._treeRef.open = true;
     } else {
-      node._treeRef.open = false;
+      this.node._treeRef.open = false;
     }
+    this.treeNotifier.notify();
   }
 }
--- a/libs/design/tree/src/tree/tree.component.ts
+++ b/libs/design/tree/src/tree/tree.component.ts
@@ -4,15 +4,23 @@ import {
   NgTemplateOutlet,
 } from '@angular/common';
 import {
+  afterNextRender,
+  AfterViewInit,
   ChangeDetectionStrategy,
+  ChangeDetectorRef,
   Component,
   ContentChild,
+  contentChildren,
+  effect,
+  Injector,
   Input,
   OnChanges,
   OnInit,
+  runInInjectionContext,
   SimpleChanges,
   TemplateRef,
   ViewEncapsulation,
+  AfterContentInit,
 } from '@angular/core';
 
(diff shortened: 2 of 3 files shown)


griest024:
@xelaint can you review the behavior to make sure that you're happy with it?

xelaint:
I'm not sure if I like the scroll to view behavior. Can we remove that for now?

libs/design/tree/src/tree-item/tree-item.directive.ts
```diff
@@ -135,30 +138,51 @@ export class DaffTreeItemDirective {
...
+
+  scrollIntoView() {
+    (<HTMLElement>this.elementRef.nativeElement).scrollIntoView();
+  }
+
+  /**
+   * Opens parent and parent of parent all the way to the root of the tree.
+   */
+  openAncestors() {
+    const openParent = (tree: DaffTreeUi<unknown>) => {
```
damienwebdev: Remove the recursion and replace with iteration.

libs/design/tree/src/tree-item/tree-item.directive.ts
```diff
@@ -135,30 +138,51 @@ export class DaffTreeItemDirective {
...
+      if (tree?.parent.parent === undefined) {
+        return;
+      }
+      tree.parent.open = true;
+      openParent(tree.parent);
+    };
+    openParent(this._node._treeRef);
+    (<Document>this.document).getElementById('tree-' + this.node._treeRef.parent.id).focus();
```
damienwebdev: Do this without dom access. This likely requires a `focus` function on treeItemDirectives.

libs/design/tree/src/tree-item/tree-item.directive.ts
```diff
@@ -135,30 +138,51 @@ export class DaffTreeItemDirective {
...
     this.treeNotifier.notify();
   }
 
   /**
    * Toggle the open state of the tree's parent.
    */
-  toggleParent(node: DaffTreeFlatNode) {
-    if(node._treeRef?.parent.parent === undefined) {
+  toggleParent() {
+    if(this.node._treeRef?.parent.parent === undefined) {
```
damienwebdev: Nice catch.

libs/design/tree/src/tree-item/tree-item.directive.ts
```diff
@@ -135,30 +138,51 @@ export class DaffTreeItemDirective {
...
       return;
     }
-    node._treeRef.parent.open = !node._treeRef.parent.open;
-    (<Document>this.document).getElementById('tree-' + node._treeRef.parent.id).focus();
+    this.node._treeRef.parent.open = !this.node._treeRef.parent.open;
+    (<Document>this.document).getElementById('tree-' + this.node._treeRef.parent.id).focus();
```
damienwebdev: Do this without dom access. This likely requires a focus function on treeItemDirectives.

libs/design/tree/src/tree/tree.component.ts
```diff
@@ -154,4 +170,22 @@ export class DaffTreeComponent extends _daffTreeBase implements OnInit, OnChange
       this.flatTree = flattenTree(this._tree, this.renderMode === 'not-in-dom');
     });
   }
+
+  ngAfterViewInit(): void {
+    this.items.changes.pipe(
+      takeUntilDestroyed(this.destroyRef),
+    ).subscribe((items: QueryList<DaffTreeItemDirective>) => {
+      runInInjectionContext(this.injector, () => afterNextRender(() => {
```
damienwebdev: Is the `selected` state only available after `ngAfterViewInit`?  Can this be done sooner? Perhaps `ngAfterContentInit`?

If this needs `ngAfterViewInit` can we do this without a subscription?

griest024:
> I'm not sure if I like the scroll to view behavior. Can we remove that for now?

okay removed for now but the user can easily end up in a position where they cannot see the active item on initial render

griest024:
@xelaint note that the focusing behavior of tree item actually appears to scroll the element into view. This doesn't work for deeply nested items though (`docs/packages/search/driver/federated`)

damienwebdev (changes requested):

libs/design/tree/src/tree-item/tree-item.directive.ts
damienwebdev: In thinking through this, we shouldn't be messing with user's focus upon the initial tree render in any circumstance. It would be too jarring.

libs/design/tree/src/tree/tree.component.ts
griest024: the `ContentChildren` aren't available in `ngAfterContentInit` because they're rendered with `ngTemplateOutlet`. yeah I could do it without the sub

libs/design/tree/src/tree/tree.component.ts
```diff
@@ -154,4 +170,18 @@ export class DaffTreeComponent extends _daffTreeBase implements OnInit, OnChange
...
+        const activeTreeItem = items.find((treeItem) => treeItem.selected);
+        if (activeTreeItem) {
+          activeTreeItem.openAncestors();
+          this.cd.markForCheck();
```
damienwebdev: I don't believe that a `markForCheck` is sufficient here. It may be that we have to do a full `detectChanges`. I would expect Angular to error here with a `Expression changed after is has been checked`.

griest024: I haven't noticed that error in the console. I can still change it if you want though

libs/design/tree/src/tree/tree.component.ts
griest024: actually no, this doesn't work. The `selected` value of the tree items isn't updated

libs/design/tree/src/tree/tree.component.ts
damienwebdev: I think the task queue shift created by the observable is likely hiding the error.

libs/design/tree/src/tree-item/tree-item.directive.ts
griest024: what list would be iterated? Recursion is naturally suited to tree structures

xelaint:
The UI LGTM.

damienwebdev:
I haven't re-reviewed this yet, but please do not merge it until I do.

libs/design/tree/src/tree-item/tree-item.directive.ts
griest024: I updated this to use `collect` (which is itself recursive but has the added protection of  a max depth) and iteration.

griest024:
bump @damienwebdev

libs/design/tree/src/tree/tree.component.ts
```diff
@@ -152,4 +166,27 @@ export class DaffTreeComponent implements OnInit, OnChanges {
       this.flatTree = flattenTree(this._tree, this.renderMode === 'not-in-dom');
     });
   }
+
+  /**
+   * @docs-private
+   */
+  ngAfterContentInit(): void {
+    runInInjectionContext(this.injector, () => {
```
damienwebdev, line 174: ?

damienwebdev (changes requested):

libs/design/tree/src/tree/tree.component.ts
griest024: this is WIP, sorry didn't mean to request a review

griest024:
this just doesn't work, we'll need a different solution

PR #2955, opened 2024-08-05
fix(external-router): extract daff paths ignoring full path

griest024:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`daffExtractDaffPathData` ignores all routes in the path except the last, incorrectly building the path when the last route is an empty path route.

## What is the new behavior?
`daffExtractDaffPathData` computes the path from all routes in the path from root.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/external-router/src/util/extract-daff-path-data.ts
+++ b/libs/external-router/src/util/extract-daff-path-data.ts
@@ -8,6 +8,6 @@ import {
  * RouteSnapshot.
  */
 export const daffExtractDaffPathData = <T extends Data = Data, K extends keyof T = keyof T>(snapshot: ActivatedRouteSnapshot, key: K): T[K] => {
-  const pathFromRoot = snapshot.pathFromRoot[snapshot.pathFromRoot.length - 1].url.map((seg) => seg.path).join('/');
+  const pathFromRoot = snapshot.pathFromRoot.flatMap((route) => route.url).map((seg) => seg.path).join('/');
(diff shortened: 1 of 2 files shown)


griest024:
@damienwebdev is this the behavior that we actually want?

libs/external-router/src/util/extract-daff-path-data.ts
```diff
@@ -8,6 +8,6 @@ import {
  * RouteSnapshot.
  */
 export const daffExtractDaffPathData = <T extends Data = Data, K extends keyof T = keyof T>(snapshot: ActivatedRouteSnapshot, key: K): T[K] => {
-  const pathFromRoot = snapshot.pathFromRoot[snapshot.pathFromRoot.length - 1].url.map((seg) => seg.path).join('/');
+  const pathFromRoot = snapshot.pathFromRoot.flatMap((route) => route.url).map((seg) => seg.path).join('/');
```
damienwebdev, line 11: What was previously happening? Can you add a test for the prior state that errored? Or, does the associated test cover that state?

damienwebdev (changes requested):

libs/external-router/src/util/extract-daff-path-data.ts
griest024: the test covers that state

damienwebdev (approved):
lgtm

PR #2987, opened 2024-08-15
feat(daffio): add effect to sub to activated route

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/apps/daffio/src/app/core/router/effects.ts
+++ b/apps/daffio/src/app/core/router/effects.ts
@@ -1,0 +1,10 @@
+import { inject } from '@angular/core';
+import { createEffect } from '@ngrx/effects';
+
+import { DaffRouterActivatedRoute } from '@daffodil/router';
+
+// TODO: should this be moved to `@daffodil/router/state`?
+export const daffioRouterActivatedRoute = createEffect(
+  (service = inject(DaffRouterActivatedRoute)) => service.route$,
+  { dispatch: false, functional: true },
+);


apps/daffio/src/app/core/router/effects.ts
```diff
@@ -0,0 +1,10 @@
+import { inject } from '@angular/core';
+import { createEffect } from '@ngrx/effects';
+
+import { DaffRouterActivatedRoute } from '@daffodil/router';
+
+// TODO: should this be moved to `@daffodil/router/state`?
+export const daffioRouterActivatedRoute = createEffect(
```
damienwebdev, line 7: What's the difference between this and a regular observable of the route? That it can be loaded via EffectModule?

damienwebdev (changes requested):

apps/daffio/src/app/core/router/effects.ts
griest024: yeah it felt appropriate to put it in an effect but there's no real functional benefit over just subbing in the `AppModule`

damienwebdev: Is there a way that we can make something like "pseudoeffects" that automatically sub in the AppModule? I'd like to keep a safe distance ( as best we can) between @ngrx and anything router related.

griest024: oh yeah we can use an `APP_INITIALIZER`, good idea

griest024:
closed in favor of #2988

PR #3060, opened 2024-09-11
feat(dgeni): render interface and type inherited members

### LINKED ISSUE #3048: [FEAT] Display inherited members in separate section of API docs
# :bulb: Feature request

## Feature Name

Display inherited members in separate section of API docs

## The Desired Behavior

The inherited members of a type should be displayed below the own members of a type on the API docs page of daff.io

```
# <type>

## members
### own
...
### inherited from <parent type>
...
```
classes implementing interfaces and type extension should likely be handled separately
## Your Use Case

As an app dev, I want to easily see which members are inherited and from which type they are inherited

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #3048

## What is the new behavior?
- types and interfaces get full list of ancestors set on doc
- all ancestors get rendered for types and interfaces

an example of what this looks like (once `@inheritdoc` is added):
image

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/tools/dgeni/src/processors/addInheritedDocsContent.spec.ts
+++ b/tools/dgeni/src/processors/addInheritedDocsContent.spec.ts
@@ -8,9 +8,10 @@ describe('AddInheritedDocsContentProcessor', () => {
     stubInterfaceMemberDescription = 'some member description';
   });
 
-  it('should copy member descriptions from inherited docs', () => {
+  it('should copy member descriptions from inherited docs for class docs', () => {
     const docs = [
       {
+        docType: 'class',
         tags: {
           tags: [{
             tagName: 'inheritdoc',
--- a/tools/dgeni/src/processors/addInheritedDocsContent.ts
+++ b/tools/dgeni/src/processors/addInheritedDocsContent.ts
@@ -3,13 +3,13 @@ import {
   Document,
 } from 'dgeni';
 
-interface DocumentWithParents extends Document {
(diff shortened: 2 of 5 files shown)


griest024:
@xelaint any comment or updates that you want to make to the UI?

griest024:
@damienwebdev do we want to overload the `@inheritdoc` tag like this? Maybe we should just do this by default to all types and interfaces regardless of whether they have the `@inheritdoc` tag?

damienwebdev:
I would prefer to only list the references to the other types, I don't think it makes sense to duplicate this content.

griest024:
> I would prefer to only list the references to the other types, I don't think it makes sense to duplicate this content.

I personally find it annoying to have to click through a bunch of different pages to find the fields available. I like having all of the fields on one page.

Edit: maybe we put the inherited properties in a collapsible spoiler tag?

damienwebdev:
@griest024 my issue is more about content duplication for search engines. If we have lots of pages that reference `ID` for example, that content is duplicated over and over again. I prefer just making the site fast enough that page traversal is an after thought.

damienwebdev:
Closing as https://github.com/graycoreio/daffodil/issues/3048 is no longer valid.

PR #3056, opened 2024-09-11
feat(eslint): add linting rules for test describe depth

gracetxgao:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Fixes: #122

## What is the new behavior?

Maximum describe() depth of 3 set for tests.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

PR #3070, opened 2024-09-17
fix(core): allow shuffle to return first element in first index

### LINKED ISSUE #1780: [BUG] `randomSubset` does not include first element
# :bug: Bug report

## Current Behavior

`randomSubset` does not include first element when its called with a length of 1.

## Expected Behavior

I expect all array elements to be selected evenly.

## Minimal reproduction of the problem with instructions

call `randomSubset([0, 1, 2, 3], 1)` many times and notice that it never returns `[0]`

Others:

</code></pre>
This might actually be a bug in `shuffle`

gracetxgao:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

``randomSubset`` with length 1 never returns the first element. This is because ``shuffle`` makes it impossible for the first element of the shuffled array to be the same as the first element of the original array, and ``randomSubset`` slices the shuffled array up to ``length`` to return the subset.

Fixes: #1780

## What is the new behavior?

``shuffle`` can now potentially return an array with the first element being the first element of the original array, so calling ``randomSubset`` with length 1 can return the first element.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/libs/core/src/utils/shuffle.spec.ts
+++ b/libs/core/src/utils/shuffle.spec.ts
@@ -11,4 +11,16 @@ describe('Core | Utils | shuffle', () => {
     });
     expect(initialArray).not.toEqual(result);
   });
+
+  it('should return array containing the same single element if input contains one element', () => {
+    const initialArray = [1];
+    const result = shuffle(initialArray);
+    expect(initialArray).toEqual(result);
+  });
+
+  it('should return an empty array if input is an empty array', () => {
+    const initialArray = [];
+    const result = shuffle(initialArray);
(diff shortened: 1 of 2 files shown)


damienwebdev (changes requested):
@gracetxgao this needs a rebase. It includes commits from https://github.com/graycoreio/daffodil/pull/3056

Additionally, can you add a test covering the empty array case as well as the single-element case?

damienwebdev (approved):
lgtm

Issue #1022, opened 2020-08-06
[FEAT] DaffSwitchComponent


xelaint:
# :bulb: Feature request

## Feature Name
DaffSwitchComponent

## The Desired Behavior
DaffSwitchComponent is form control that toggles the state of a single setting "on/off" by clicking it.

## Your Use Case
As an enduser, I'd like to be able to toggle between two states.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

damienwebdev:
@xelaint can you post a link to the figma design for this component?

xelaint:
@damienwebdev There isn't a design for this yet.

PR #3143, opened 2024-10-04
feat(design): add basic switch component

gracetxgao:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Starter for #1022

## What is the new behavior?

A basic switch component is set up.

```
[ ] Yes
[x] No
```

## Other information


xelaint:
To do:
- [x] Loading state
- [x] Disabled state (ability to disable a switch, or set disabled when toggle is also loading)
- [x] Show error message if switch fails to toggle
- [x] A label position property that changes the orientation of the label from being displayed left or right of the toggle UI
- [x] ARIA considerations (https://www.w3.org/WAI/ARIA/apg/patterns/switch/)
- [x] Keyboard interactions

apps/design-land/src/app/app.module.ts
```diff
@@ -41,6 +42,8 @@ import { DesignLandTemplateModule } from './core/template/template.module';
     FontAwesomeModule,
     DesignLandNavModule,
     DesignLandTemplateModule,
+    DaffToastModule,
```
xelaint: this import is causing an error and is not needed (refer to comment on switch component)

libs/design/switch/src/switch/switch.component.ts
```diff
@@ -7,14 +9,80 @@ import {
...
   @Input() checked = false;
+  @Input() loading = false;
+  @Input() position: 'left' | 'right' | 'top' | 'bottom' = 'left';
+
   @Output() toggle = new EventEmitter<boolean>();
+
+  constructor(private toastService: DaffToastService) {}
+
+  onToggle() {
+    try {
```
xelaint: An error while toggling a switch component should not use the global toast push notification UI. Use the `DaffErrorMessageComponent` instead by adding a `<ng-content select="daff-error-message"></ng-content>` to the switch template.

libs/design/switch/src/switch/switch.component.scss
```diff
@@ -12,6 +12,34 @@
   justify-content: space-between;
   gap: 1rem;
 
+  padding: 10px;
```
xelaint: This component should not determine the external spacing in the event that more than one switch is placed next to each other. It's up to the implementation (in the example component) to correctly space out the switches.

libs/design/switch/src/switch/switch.component.ts
```diff
@@ -7,14 +9,80 @@ import {
...
   standalone: true,
+  imports: [
+    CommonModule,
+    DAFF_LOADING_ICON_COMPONENTS,
+  ],
 })
 export class DaffSwitchComponent {
   @Input() checked = false;
+  @Input() loading = false;
+  @Input() position: 'left' | 'right' | 'top' | 'bottom' = 'left';
```
xelaint: Refer to `DaffNavPlacement` on how we implement input properties like this. Also, update name to `labelPosition` for better clarity.

libs/design/switch/src/switch/switch.component.ts
```diff
@@ -7,14 +9,80 @@ import {
...
+        message: error.message,
+        status: 'danger',
+      },
+      {
+        duration: 5000,
+      });
+    }
+  }
+
+  @HostBinding('class.daff-switch--position-left') get isLeftPosition() {
```
xelaint: These `HostBindings` are unnecessary if the class is being applied to the `<label class="daff-switch">` element.

libs/design/switch/src/switch/switch.component.html
```diff
@@ -1,12 +1,23 @@
-<label class="daff-switch">
+<label class="daff-switch" 
+  [class.daff-switch--disabled]="disabled"
+  [class.daff-switch--position-left]="position === 'left'"
```
xelaint: I'm slowly making updates to changes in class names so they're not so long. These can be updated to `class.disabled`, `class.left`, etc.

libs/design/switch/src/switch/switch.component.html
```diff
@@ -1,12 +1,23 @@
...
+  [class.daff-switch--position-top]="position === 'top'"
+  [class.daff-switch--position-bottom]="position === 'bottom'"
+>
   <div class="daff-switch__text">
     <ng-content></ng-content>
   </div>
   <div class="daff-switch__toggle" [class.checked]="checked">
-    <div class="daff-switch__activator"></div>
-    <input type="checkbox" role="switch" class="cdk-visually-hidden"
+    <div *ngIf="!loading" class="daff-switch__activator"></div>
```
xelaint: Prefer this template:

```
<div class="daff-switch__activator">
  <daff-loading-icon *ngIf="loading"></daff-loading-icon>
</div>
```

libs/design/switch/examples/src/basic-switch/basic-switch.component.html
```diff
@@ -1 +1,7 @@
-<daff-switch [checked]="checked" (toggle)="changeSwitch()">Label</daff-switch>
\ No newline at end of file
+<daff-switch [checked]="checked">Default</daff-switch>
+<daff-switch [checked]="checked" position="left">Left Label</daff-switch>
```
xelaint: Move the position examples to a separate component called `switch-label-positions` and use `<select>` to select positions.

libs/design/switch/src/switch/switch.component.scss
```diff
@@ -12,6 +12,34 @@
...
+  &.daff-switch--position-left {
+    flex-direction: row;
+  }
+
+  &.daff-switch--position-right {
+    flex-direction: row-reverse;
+  }
+
+  &.daff-switch--position-top {
+    flex-direction: column;
```
xelaint: add `align-items: flex-start;`

libs/design/switch/src/switch/switch.component.scss
```diff
@@ -12,6 +12,34 @@
...
+  }
+
+  &.daff-switch--position-bottom {
+    flex-direction: column-reverse;
```
xelaint: add `align-items: flex-start;`

libs/design/switch/src/switch/switch.component.ts
```diff
@@ -7,14 +9,80 @@ import {
...
+import { DaffToastService } from '@daffodil/design/toast';
+
 @Component({
   selector: 'daff-switch',
   changeDetection: ChangeDetectionStrategy.OnPush,
   templateUrl: './switch.component.html',
   styleUrls: ['./switch.component.scss'],
   standalone: true,
+  imports: [
+    CommonModule,
```
xelaint: Don't import `CommonModule` in standalone components. Only import what's necessary.

xelaint (changes requested):

xelaint:
To-do ARIA considerations (https://www.w3.org/WAI/ARIA/apg/patterns/switch/):

- [x] A visible label referenced by the value of [aria-labelledby](https://w3c.github.io/aria/#aria-labelledby) set on the element with role switch.
- [x] [aria-label](https://w3c.github.io/aria/#aria-label) set on the element with role switch.
- [x] When on, the switch element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to true.
- [x] When off, the switch element has state [aria-checked](https://w3c.github.io/aria/#aria-checked) set to false.

xelaint:
@gracetxgao I went ahead and made updates to the aria labels. Lmk if you have any questions.

In the basic switch example, clicking on the disabled or loading switches toggles the default switch. Clicking on the switch in `switch-label-positions-example` also toggles the default switch. I think you may need to add form controls to examples.

xelaint (changes requested):
Keyboard functionalities:
- [x] `tab` key should tab to a switch. Switches are currently not tabbable.
- [x] `space` key to toggle the switch.

libs/design/switch/src/switch/switch.component.ts
```diff
@@ -33,6 +34,16 @@ export class DaffSwitchComponent {
   @Input() loading = false;
   @Input() labelPosition: DaffLabelPosition = DaffLabelPositionEnum.LEFT;
 
+  @HostBinding('attr.tabindex') tabindex = 0;
```
xelaint: tabindex should be set to `-1` if switch is disabled or loading

libs/design/switch/examples/src/basic-switch/basic-switch.component.html
```diff
@@ -1,3 +1,3 @@
-<daff-switch [checked]="checked">Default</daff-switch>
+<daff-switch [checked]="checked" [formControl]="defaultSwitchExample" ngDefaultControl>Default</daff-switch>
 <daff-switch [checked]="checked" disabled>Disabled</daff-switch>
```
xelaint: Can you create separate examples for disabled and loading? There should be another example to show switch with an error message too.

xelaint (changes requested):

libs/design/switch/src/switch/switch.component.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  @Output() toggle = new EventEmitter<boolean>();
+
+  onToggle() {
+    if (!this.disabled) {
+      try {
+        this.checked = !this.checked;
+        this.toggle.emit(this.checked);
+      } catch (error) {
+        throw new Error(DaffSwitchErrorMessage);
```
xelaint: I don't think we need to try and catch the error here? Should be up to the implementation

i.e.
```
<daff-switch [checked]="checked" [formControl]="defaultSwitchExample">
  Default
  <daff-error-message *ngIf="defaultSwitchExample.invalid">Custom error message</golf-error-message>
</daff-switch>
```

xelaint (changes requested):

libs/design/switch/examples/src/disabled-switch/disabled-switch.component.html
```diff
@@ -0,0 +1,2 @@
+<daff-switch [checked]="checked" disabled>Disabled</daff-switch>
+<daff-switch [checked]="checked" loading="true">Loading</daff-switch>
```
xelaint: Sorry, should've clarified more. Loading should be its own example as well.

xelaint (changes requested):

libs/design/switch/src/switch/switch.component.ts
xelaint: I made updates here [0e182c5](https://github.com/graycoreio/daffodil/pull/3143/commits/0e182c51aab72c81a575367fdeea56b5ab534b50) to how the error message should be used.

Also, switches that have errored should still be toggle-able.

xelaint (changes requested):
There seems to be something funny with tabindex and focus. When I tab to the first switch and then tab again, the focus goes somewhere not visible. Focus goes to the HTML button upon the 3rd tab.

Tabbing after focus is on the switch should move focus to the next available element (in this case is the HTML button). Something else that's not visible within the component seems to be tabbable.

damienwebdev:
@gracetxgao please remove the `package-lock.json` from this diff.

xelaint (dismissed):

libs/design/switch/src/switch/switch.component.ts
```diff
@@ -81,12 +107,12 @@ export class DaffSwitchComponent {
 
   @Input() id: string = 'daff-switch-' + switchUniqueId++;
 
-  @Output() toggle = new EventEmitter<boolean>();
+  @Output() toggleSwitch = new EventEmitter<boolean>();
```
damienwebdev: I kinda liked `toggle`, do you prefer toggleSwitch?

libs/design/switch/README.md
```diff
@@ -2,7 +2,7 @@
 A switch allows users to toggle the state of a single setting.
 
 ## Overview
```
damienwebdev, line 4: Can you add the examples to the readme, as done in the other component guides?

damienwebdev (changes requested):

libs/design/switch/src/switch/switch.component.ts
gracetxgao: I was getting this error `Output bindings, including aliases, should not be named as standard DOM events`

damienwebdev: Ok, perhaps `switch` then? Thoughts @xelaint?

xelaint: `toggled`?

xelaint:
@gracetxgao please rebase on develop since we pushed updates to linting, so there may be a few linting errors in this PR.

damienwebdev (approved):
lgtm

PR #3328, opened 2024-11-01
feat(design): move each button type to its own component

### LINKED ISSUE #3196: [FEAT] Refactor Button Component
# :bulb: Feature request

## Feature Name
Refactor Button Component

## The Desired Behavior
Every button variant should be its own component, rather than it all being in one single component. They still share the same css class of `.daff-button`, with an extension of the variant name, i.e. `.daff-flat-button`

## Your Use Case
As a developer, I would like to reduce the file size of the button component and unnecessary import of all button variants.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

xelaint:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #3196 

```
[ ] Yes
[x] No
```

## Other information


libs/design/button/src/button/stroked/stroked.component.ts
```diff
@@ -0,0 +1,55 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
+  imports: [
+    NgIf,
+    DaffPrefixSuffixModule,
+    DAFF_LOADING_ICON_COMPONENTS,
+  ],
+})
+export class DaffStrokedButtonComponent
+  extends DaffButtonBaseDirective {
```
damienwebdev, line 48: Since we define the `DaffButtonBaseDirective` as a directive, not a component, can we just add that directive as a host directive?

damienwebdev (changes requested):

libs/design/button/src/button/stroked/stroked.component.ts
xelaint: I don't think it'll work with `DaffPrefixDirective` and `DaffSuffixDirective` not being able to be used as a hostDirective.

damienwebdev: Yep. Thanks for the reminder. From empirical testing and https://github.com/angular/angular/issues/51599 it looks like you can't access content/view children in host directives.

xelaint:
@damienwebdev The `@if` for prefix and suffix works, but the issue with splitting the components up is that `DAFF_BUTTON_COMPONENTS,` handles exporting the `DaffPrefixSuffixModule` so that the usage component doesn't have to import it.

Since each component example is only importing the specific button type (i.e. DaffFlatButtonComponent), the `DaffPrefixSuffixModule` import here doesn't actually work.

```
@Component({
  // eslint-disable-next-line @angular-eslint/component-selector
  selector: '' +
    'button[daff-flat-button]' + ',' +
    'a[daff-flat-button]',
  templateUrl: '../button-base.component.html',
  styleUrl: './flat.component.scss',
  encapsulation: ViewEncapsulation.None,
  changeDetection: ChangeDetectionStrategy.OnPush,
  standalone: true,
  imports: [
    DaffPrefixSuffixModule,
    DAFF_LOADING_ICON_COMPONENTS,
  ],
})
```

damienwebdev:
@xelaint I think that's perfectly fine. While the prefix suffix code is tiny, we still don't want to encourage people to import code they don't use. If they want to use prefix and suffix, they can import it.

damienwebdev (approved):

damienwebdev (approved):
lgtm

PR #3343, opened 2024-11-08
feat(ssr): add link asset preload header service

### LINKED ISSUE #3340: [FEAT] SSR Image Preloading via Header
# :bulb: Feature request

## Feature Name

SSR Image Preloading via Header

## The Desired Behavior

I want a service that allows me to add [image preload link headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link) to the SSR document reponse.

## Your Use Case

As an app dev, I'd like to be able to specify important, LCP-impacting images in various priorities.

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #3340 

## What is the new behavior?
- Adds a header service that will be a noop on the browser but has an express implementation
- Adds a asset preload service that invokes the swappable header service

```
[ ] Yes
[x] No
```

## Other information


libs/ssr/src/headers/service.token.ts
```diff
@@ -0,0 +1,14 @@
+import { createSingletonInjectionToken } from '@daffodil/core';
+
+import { DaffSsrHeaderService } from './service.type';
+
+export const {
+  /**
+   * A token to hold the `DaffSsrHeaderService`.
+   */
+  token: DAFF_SSR_HEADER_SERVICE,
```
damienwebdev: This token should  be `providedIn: 'root'` with the no-op service.

libs/ssr/src/headers/browser.provider.ts
```diff
@@ -0,0 +1,10 @@
+import { DaffSsrHeaderNoopService } from './noop.service';
+import { provideDaffSsrHeaderService } from './service.token';
+
+/**
+ * Provides the `DaffSsrHeaderNoopService`.
+ */
+export const provideDaffSsrHeaderBrowserService = () => [
```
damienwebdev: This provider shouldn't exist, it should just use the no-op service by default. Then, we can override the implementation in SSR.

libs/ssr/express/src/response.token.ts
```diff
@@ -0,0 +1,14 @@
...
+export const {
+  /**
+   * A token to hold the express response during SSR.
+   */
+  token: DAFF_SSR_EXPRESS_RESPONSE,
+  /**
+   * Provider for `DAFF_SSR_EXPRESS_RESPONSE`.
+   */
+  provider: provideDaffSsrExpressResponse,
+} = createSingleInjectionToken<Response>('DAFF_SSR_EXPRESS_RESPONSE');
```
damienwebdev: I think this response type should be our own and not explicitly from express.

We should support solely a tiny subset of the `Response` methods that are relevant for Angular developers. I'd suggest starting with only the response methods that we currently use and omitting the rest.

At the outset, I think:

```ts
export interface Response {
  get(header: string): string;
  set(headers: Dict<string, Array<string> | string>): void;
  append(string, Array<string> | string>): void;
  status(code: number): Response
}
```

This would replace `DaffSsrHeaderExpressService`

libs/ssr/src/headers/link-asset-preload/service.ts
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Adds a link preload header to the SSR response.
+   *
+   * @param uri The URI of the asset to preload. Take care that this is percent-encoded. @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Link#encoding_urls
+   * @param extra Any extra values to add to the header. The first value of the tuple is the name of the param, the second is the value.
+   */
+  addHeader(uri: string, kind: DaffSsrHeadersLinkPreloadAssetKind, priority: DaffSsrHeadersLinkPreloadAssetPriority, extra: Record<string, string> = {}): void {
+    this._counter++;
+    if (this._counter > 6) {
+      console.warn('Adding more than 6 preloaded assets is discouraged. Most browsers have a default parallel HTTP request limit of 6 so more than 6 preloads may have unexpected behavior.');
```
damienwebdev: After doing some reading, this appears like its potentially only relevant with HTTP1.1, not HTTP2.

I think we can safely remove this warning.

damienwebdev (changes requested):

libs/ssr/express/src/response.token.ts
griest024: What's your preference for the translation layer between our type and the express response?

damienwebdev: I think this is simple enough:

```ts
import { Response as ExpressResponse } from 'express';

export interface DaffResponse {
  get(header: string): string;
  set(headers: Record<string, Array<string> | string>): void;
  append(header: string, value: Array<string> | string): void;
  status(code: number): Response
}

export class Response implements DaffResponse {
  constructor(
     private response: ExpressResponse,
  ) {}

  get(header: string): string {
    return this.response.get(header);
  } 

  set(headers: Record<string, string | string[]>): void {
    this.response.set(headers);
  }

  append(header: string, value: string | string[]): void {
    this.response.append(header, value);
  }

  status(code: number): Response {
    this.response.status(code);
    return this;
  }
}
```

griest024: Does `status` need to be chainable?

libs/ssr/src/headers/service.token.ts
griest024: changed to a response token, which defaults to the noop response

libs/ssr/src/headers/browser.provider.ts
griest024: removed

libs/ssr/express/src/response.token.ts
griest024: changed

libs/ssr/src/headers/link-asset-preload/service.ts
griest024: removed

damienwebdev (approved):

PR #3385, opened 2024-12-30
feat(dgeni): render API source block

### LINKED ISSUE #3306: [FEAT] Source-style API symbol page
# :bulb: Feature request

## Feature Name

Source-style API symbol page

## The Desired Behavior

image

griest024:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #3306

## What is the new behavior?
One of the major outstanding issues with this is that the dgeni TS package doesn't correctly infer types of everything that TS can. For example, injection tokens and providers that are destructured from the factory function are untyped. Functions that do not have explicit return annotations have untyped return types even when TS correctly infers the return type.

Some of the `const`s that are function do have some of their type correctly inferred though.

```
[ ] Yes
[x] No
```

## Other information


griest024:
@xelaint @damienwebdev this is ready to preview and give feedback

xelaint:
Is it possible to make the individual properties within the source block linkable to the properties section? When a user is on a page with a long source block section (i.e. `/docs/api/cart/state/DaffCartFacade`), the individual properties take a while to scroll to.

tools/dgeni/src/templates/api/base.template.html
```diff
@@ -2,5 +2,7 @@
 <h1>{$ doc.name $}</h1>
 {% endblock %}
 <p>{$ doc.description $}</p>
+<h2>API</h2>
```
xelaint, line 5: I don't think it's necessary to have this heading?

griest024:
> Is it possible to make the individual properties within the source block linkable to the properties section? When a user is on a page with a long source block section (i.e. `/docs/api/cart/state/DaffCartFacade`), the individual properties take a while to scroll to.

yes this is my eventual goal. I'm not sure if it should be in this PR though since it will require changes to how the properties are laid out

tools/dgeni/src/templates/api/base.template.html
griest024: While I'm not attached to this specific heading, I think some separator between the description and the source block is necessary. When the descriptions includes some code blocks at the end (which is fairly common), I think its not immediately apparent that the source block is not part of the description. See `/docs/api/auth/state/daffAuthProvideUnauthenticatedHooks`:
image

xelaint: Yeah, I agree there should be a separator then. Is the `providers` code block a usage example? If so, I think we should also label it as usage example as well, or split up usage example and api into tabs like how angular does it?

xelaint:
Do we need to render the API blocks for design?

Screen Shot 2024-12-31 at 11 04 05 AM

griest024:
> Do we need to render the API blocks for design?

I like them but I'll leave the final decision up to you and @damienwebdev. I have them here mostly as an example of what the source blocks will look like for components.

tools/dgeni/src/templates/api/base.template.html
griest024: yes the provider block is a usage example. I agree, we should tag it as [`@example`](https://jsdoc.app/tags-example) and render it separately (tabs sound good)

xelaint: > yes the provider block is a usage example. I agree, we should tag it as [`@example`](https://jsdoc.app/tags-example) and render it separately (tabs sound good)

Sounds good. @damienwebdev likes the idea of tabs as well.

xelaint (approved):

PR #3457, opened 2025-01-21
feat(demo): update shopify driver usage in demo 

leozhang14:
## PR Checklist

## PR Type

Implementation of the Shopify product driver.

```
[ ] Bugfix
[y] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Old code that does not work. Uses shop to query (not supported in Shopify Storefront 2025/01).

Fixes: N/A

## What is the new behavior?

get (url), getAll, getByURL, and getBestSellers should all function properly.

```
[y] Yes
[ ] No
```

Has effects on the demo app, as well as the product library since the shopify driver service files have been modified.

## Other information

IMPORTANT:

I don't know how to test the get and getByURL methods as they return observables. I will have to look into this more. I am also running into an issue with the testing for getAll and get, although the daffodil demo seems to work?

Chrome 131.0.0.0 (Mac OS 10.15.7) Driver | Shopify | Product | ProductService getAll | getting a list of products should should return an observable array of 20 products by default FAILED
        Error: Expected 0 to equal 20.

Chrome 131.0.0.0 (Mac OS 10.15.7) Driver | Shopify | Product | ProductService get | getting a single product should return an observable single product FAILED
        Error: Expected undefined to equal 'bd0de172-3fbd-4e17-8e61-f6d73ff876c3'


libs/driver/in-memory/src/backend/delegate.class.ts
```diff
@@ -31,10 +31,7 @@ export class DaffInMemoryBackendDelegate implements DaffInMemoryDataServiceInter
       // allow child backends to return nully values
       // which will fall back to default request behavior (https://github.com/angular/angular/tree/main/packages/misc/angular-in-memory-web-api#http-request-handling)
       ? request.call(backend, reqInfo)
-      : reqInfo.utils.createResponse$(() => ({
-        status: STATUS.NOT_FOUND,
-        statusText: `Backend ${reqInfo.collectionName} not found or does not support the ${method} request method`,
-      }));
+      : undefined;
```
griest024, line 34: What is the motivation for this change?

libs/product/driver/shopify/src/product.service.ts
```diff
@@ -70,35 +82,62 @@ export const GetAllProductsQuery = gql`
...
-export const GetAProduct = gql`
-  query ShopifyGetAProduct($id: ID!){
-    node(id: $id) {
-      id
-      ... on Product {
-        title
-      }
-    }
-  }
+export const GetProductByIdQuery = gql`
```
griest024: Queries should have variables and response typed, see https://github.com/graycoreio/daffodil/blob/develop/libs/cart/driver/magento/src/queries/add-item/query.ts for an example

libs/product/driver/shopify/src/product.service.ts
```diff
@@ -70,35 +82,62 @@ export const GetAllProductsQuery = gql`
...
  * Transforms a ProductNode into a different object.
  *
  * @param node - ProductNode object
  * @returns A Product object
  */
-export const DaffShopifyProductTransformer = (node: ProductNode): DaffProduct => ({
-  id: node.id,
-  url: null,
+
+const DaffShopifyProductTransformer = (node: ProductNode): DaffProduct => ({
```
griest024: Change this to camelCase. I know it was PascalCase before but the convention for `const`s and `function`s is camelCase.

libs/product/driver/shopify/src/product.service.ts
```diff
@@ -110,54 +149,71 @@ export const DaffShopifyProductTransformer = (node: ProductNode): DaffProduct =>
...
-        id: productId,
-        products: [DaffShopifyProductTransformer(result.data.node)],
-      })),
-    );
+  getByUrl(url: DaffProduct['url']): Observable<DaffProductDriverResponse> {
+    return this.apollo
+      .query<GetAllProductsResponse>({
+        query: GetAllProductsQuery,
+        variables: { length: this.defaultLength }, // Adjust the number to fit max number of products in store
+      })
```
griest024: Is there no better way to do this? Can we not just query the product directly with the `url`?

libs/product/driver/shopify/src/shopifyDriverConfig.ts
```diff
@@ -0,0 +1,9 @@
+export function shopifyDriverConfig(uri: string, token: string) {
+  return {
+    uri: uri + '/api/2025-01/graphql.json',
```
griest024: I prefer template strings. Using `+` for string concatenation can result in unexpected behavior.

libs/product/driver/src/createApolloConfig.ts
```diff
@@ -0,0 +1,41 @@
+import { inject } from '@angular/core';
+import { InMemoryCache } from '@apollo/client/cache';
+import { ApolloLink } from '@apollo/client/core';
+import { setContext } from '@apollo/client/link/context';
+import { onError } from '@apollo/client/link/error';
+import { HttpLink } from 'apollo-angular/http';
+
+// Should be a generic factory function whenever Apollo Clients need to be created (i.e. Shopify, Magento, etc.)
+export function createApolloConfig({
```
griest024: This has nothing to do with product, it should be in `@daffodil/driver/graphql`.

griest024 (changes requested):

griest024:
> I don't know how to test the get and getByURL methods as they return observables. I will have to look into this more.

Run the tests in a browser and check your console for errors. I suspect its because the flushed data doesn't conform to your new `ProductNode` interface. The *ideal* solution is to create a testing package (see `@daffodil/product/magento/testing`) that holds factories that will autogenerate testing data with the appropriate structure.

griest024:
@leozhang14 has any code in this PR been generated by AI tools?

libs/driver/in-memory/src/backend/delegate.class.ts
leozhang14: I think I accidentally deleted this line when I was experimenting with some things in an attempt to debug an issue I was having a while ago. Sorry.

leozhang14: To clarify, I didn't mean to make this change and will undo it.

libs/product/driver/shopify/src/product.service.ts
leozhang14: This one is a bit tricky.

Shopify Storefront has 2 ways of querying a single product, by id or by handle. Querying by handle would be a valid solution if the onlineStoreUrl is just the product name appended to the end of the store url. However, I can't seem to get the onlineStoreURL field since it always returns null.

The [documentation](https://shopify.dev/docs/api/storefront/2024-10/queries/product) says this is because the product isn't published to the online store sales channel, but in the daffodil-demo it clearly is active in the online store sales channel. I've seen others encounter this same bug online (seems to be because the store is a "Developer Store"?).

For this reason, I just left the function as a query of all products, then checking each product to see if the onlineStoreUrl lined up.

On a side note, the test suite for the shopify product driver doesn't test getBestSellers and getByURL, just FYI.

leozhang14:
> > I don't know how to test the get and getByURL methods as they return observables. I will have to look into this more.
> 
> Run the tests in a browser and check your console for errors. I suspect its because the flushed data doesn't conform to your new `ProductNode` interface. The _ideal_ solution is to create a testing package (see `@daffodil/product/magento/testing`) that holds factories that will autogenerate testing data with the appropriate structure.

This was the issue, thanks. I'll look into the "testing package" for magento more. Just to confirm, it would be more appropriate to make this change in a separate PR, correct?

libs/driver/in-memory/src/backend/delegate.class.ts
damienwebdev: @griest024 @leozhang14 this change is necessary in order to use the Shopify driver with the in-memory backend enabled.

Without it, the `passThru` feature of the in-memory backend doesn't work as expected.

leozhang14:
> @leozhang14 has any code in this PR been generated by AI tools?

Just these lines below were generated with AI,  to perform the check in getByUrl to see if any product in the list of products had a matching URL.

const matchedProduct = products.find((edge) => edge.node.onlineStoreUrl === url)?.node;
          if (!matchedProduct) {
            throw new Error(`Product with URL ${url} not found.`);
          }
          
Apologies if this was an issue—I wasn’t sure if using AI tools was against guidelines. Could you clarify if this is acceptable? Sorry in advance if I misunderstood.

libs/product/driver/shopify/src/product.service.ts
damienwebdev: @leozhang14 handle appears to be a property of the Shopify product: https://shopify.dev/docs/api/storefront/2025-01/objects/Product#field-handle

For the Daffodil sticker store, what's the value for the sticker product?

griest024:
> > @leozhang14 has any code in this PR been generated by AI tools?
> 
> Just these lines below were generated with AI, to perform the check in getByUrl to see if any product in the list of products had a matching URL.
> 
> const matchedProduct = products.find((edge) => edge.node.onlineStoreUrl === url)?.node; if (!matchedProduct) { throw new Error(`Product with URL ${url} not found.`); }
> 
> Apologies if this was an issue—I wasn’t sure if using AI tools was against guidelines. Could you clarify if this is acceptable? Sorry in advance if I misunderstood.

its allowed but our [contribution rules](https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#-submitting-a-pull-request-pr) state that you must openly disclose it in the description of your PR. It's not a big deal and we really don't make that notice very prominent but its good for reviewers to know so we can pay extra close attention.

Also, in my personal opinion, while I think AI can be a powerful and potentially valuable tool in programming, its important to make sure you always understand the code that you end up using (whether generated by AI or copied from SO, etc.). AI should be a labor-saving tool not a skill crutch. Overall, whatever results in the best final product is fine by me.

griest024:
> > > I don't know how to test the get and getByURL methods as they return observables. I will have to look into this more.
> > 
> > 
> > Run the tests in a browser and check your console for errors. I suspect its because the flushed data doesn't conform to your new `ProductNode` interface. The _ideal_ solution is to create a testing package (see `@daffodil/product/magento/testing`) that holds factories that will autogenerate testing data with the appropriate structure.
> 
> This was the issue, thanks. I'll look into the "testing package" for magento more. Just to confirm, it would be more appropriate to make this change in a separate PR, correct?

Generally, the factories for a model should be introduced in the same PR that the models themselves are. I would be okay with adding them to this PR but if you wanted to extract the model + factory change to its own PR, that's fine too.

libs/product/driver/shopify/src/product.service.ts
griest024: if `onlineStoreUrl` is always `null`, then the `find` would always fail anyway. I think you should implement this as if the backend was behaving correctly. At least in the test you can mock what the response *should* be.

Also, (@damienwebdev feel free to weigh in) I don't think we need `getBestSellers`. It should honestly just be deprecated. I do think you should add a spec for `getByUrl` though.

libs/driver/in-memory/src/backend/delegate.class.ts
griest024: hmm can you repro the use case in an integration test in a separate PR? I don't follow.

libs/product/driver/shopify/src/product.service.ts
damienwebdev: @griest024 we should definitely remove `getBestSellers` from the interface, it has nothing to do with the driver spec and is a remnant of a thought of an idea from 7 years ago.

libs/product/driver/shopify/src/product.service.ts
leozhang14: Done.

libs/product/driver/shopify/src/product.service.ts
leozhang14: Done.

libs/product/driver/shopify/src/shopifyDriverConfig.ts
leozhang14: Done.

libs/product/driver/src/createApolloConfig.ts
leozhang14: There is no existing folder called graphql in @daffodil/driver. Am I correct in assuming you want me to create a graphql folder in libs/driver, then place this function inside?

leozhang14:
> > > > I don't know how to test the get and getByURL methods as they return observables. I will have to look into this more.
> > > 
> > > 
> > > Run the tests in a browser and check your console for errors. I suspect its because the flushed data doesn't conform to your new `ProductNode` interface. The _ideal_ solution is to create a testing package (see `@daffodil/product/magento/testing`) that holds factories that will autogenerate testing data with the appropriate structure.
> > 
> > 
> > This was the issue, thanks. I'll look into the "testing package" for magento more. Just to confirm, it would be more appropriate to make this change in a separate PR, correct?
> 
> Generally, the factories for a model should be introduced in the same PR that the models themselves are. I would be okay with adding them to this PR but if you wanted to extract the model + factory change to its own PR, that's fine too.

I will try to add the model + factory changes to this PR as well. Thanks for the advice.

libs/product/driver/src/createApolloConfig.ts
leozhang14: @griest024 Could I get an update on this? Thanks.

libs/product/driver/shopify/src/models/shopify-money.ts
```diff
@@ -0,0 +1,4 @@
+export interface ShopifyMoneyV2 {
+  amount: number;
+  currencyCode: string;
+}
```
griest024: This model (and the associated factory) should be in `@daffodil/driver/shopify` as it is a generic model and doesn't have anything to do with products specifically. I really like how you are breaking up the types!

libs/product/driver/shopify/src/models/shopify-product.ts
```diff
@@ -0,0 +1,13 @@
+import { ProductImages } from './product-images';
+import { ProductPriceRange } from './product-price-range';
+
+export interface ProductNode {
```
griest024: All these models should be prefixed with `Shopify` since they specifically refer to the shape of data on the shopify backend. So this would be `ShopifyProductNode`.

libs/product/driver/src/createApolloConfig.ts
griest024: hm, I'm torn between `@daffodil/driver/graphql` and `@daffodil/driver/apollo`. @damienwebdev whats your opinion? @leozhang14 you should definitely add a driver subpackage for this though, see other subpackages for examples of file structure. yes, export this function from that subpackage. This should be in a separate PR.

griest024 (changes requested):
I'm liking the changes so far!

damienwebdev:
@griest024 @leozhang14 instead of manually defining the types, as we have, what if instead we use [graphql-codegen](https://www.shopify.com/partners/blog/introducing-codegen-for-hydrogen) to only gen the types our queries use. This does mean that factories may break if we modify the outgoing query, but it will make the maintenance devx a bit nicer.

damienwebdev:
Re: https://github.com/graycoreio/daffodil/pull/3457#discussion_r1927711326 (IDK why Github wont let me comment in line) I'd vote for `@daffodil/driver/graphql` I don't intend to support multiple graphql clients simultaneously.

griest024:
> @griest024 @leozhang14 instead of manually defining the types, as we have, what if instead we use [graphql-codegen](https://www.shopify.com/partners/blog/introducing-codegen-for-hydrogen) to only gen the types our queries use. This does mean that factories may break if we modify the outgoing query, but it will make the maintenance devx a bit nicer. 

I like this idea. Anything to relieve maintenance burden

leozhang14:
> @griest024 @leozhang14 instead of manually defining the types, as we have, what if instead we use [graphql-codegen](https://www.shopify.com/partners/blog/introducing-codegen-for-hydrogen) to only gen the types our queries use. This does mean that factories may break if we modify the outgoing query, but it will make the maintenance devx a bit nicer.

Will we implement this using the Shopify CLI? Also, should this change be in this PR, or another one?

damienwebdev:
Id say do what works, I have no strong opinion

libs/product/driver/shopify/testing/src/factories/product-images.factory.ts
```diff
@@ -0,0 +1,24 @@
...
+import { DaffModelFactory } from '@daffodil/core/testing';
+import { ShopifyProductImages } from '@daffodil/product/driver/shopify';
+
+class MockShopifyProductImages implements ShopifyProductImages {
+  nodes = [
+    {
+      id: `gid://shopify/ProductImage/${faker.datatype.number({ min: 10000000000 })}`,
+      url: faker.image.imageUrl(),
+      altText: faker.random.words(5),
+    },
```
griest024: Same here, use the child factory

libs/product/driver/shopify/testing/src/factories/product-node.factory.ts
```diff
@@ -0,0 +1,41 @@
...
+    },
+  };
+  images = {
+    nodes: [
+      {
+        id: `gid://shopify/ProductImage/${faker.datatype.number({ min: 10000000000 })}`,
+        url: faker.image.imageUrl(),
+        altText: faker.random.words(5),
+      },
+    ],
```
griest024: Use `ShopifyProductImagesFactory.createMany` to generate these. See `DaffCategoryFactory` for an example of how to use child factories.

griest024 (changes requested):
You need to use child factories where appropriate. Factories should only mock out the fields that their corresponding type declares. So `ShopifyProductGraph` only has one field: `nodes`, `MockShopifyProductGraph` should only mock out `nodes` at the top level and rely on the child factory mock to handle its internal fields.

libs/driver/shopify/ng-package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$schema": "../../node_modules/ng-packagr/ng-entrypoint.schema.json",
```
griest024: I think this needs three `../`, make sure this path resolves.

libs/driver/shopify/src/shopify.module.ts
griest024, line 1: Remove this empty file

griest024 (changes requested):
I like this change!

libs/product/driver/shopify/src/config/shopify-driver-config.ts
```diff
@@ -0,0 +1,9 @@
+export function shopifyDriverConfig(uri: string, token: string) {
```
damienwebdev: This function isn't related to `product` it looks like core shopify driver config.

libs/driver/shopify/src/graphql/create-apollo-config.ts
```diff
@@ -0,0 +1,38 @@
+import { inject } from '@angular/core';
+import { InMemoryCache } from '@apollo/client/cache';
+import { ApolloLink } from '@apollo/client/core';
+import { setContext } from '@apollo/client/link/context';
+import { onError } from '@apollo/client/link/error';
+import { HttpLink } from 'apollo-angular/http';
+
+export function createApolloConfig({
```
damienwebdev: This API is unclear, what is `uri`? and what are `headers`? Are users expected to know how to configure this?

The API of `shopifyDriverConfig` is more user-friendly and I think captures the true intent of this function.

apps/demo/src/app/drivers/shopify.module.ts
```diff
@@ -1,58 +1,46 @@
...
     DaffCartInMemoryDriverModule.forRoot(),
     DaffNewsletterInMemoryDriverModule.forRoot(),
+    DaffCategoryInMemoryDriverModule.forRoot(),
+    DaffNavigationInMemoryDriverModule.forRoot(),
+    DaffAuthorizeNetInMemoryDriverModule.forRoot(),
+  ],
+  providers: [
+    provideDaffProductExtraFactoryTypes(DaffDefaultProductFactory),
+    provideApollo(
+      () => createApolloConfig(shopifyDriverConfig(
```
damienwebdev: This provider and subsequent functions are pretty uncomfortable to read, can you adjust this to look like:

```ts
provideShopifyApolloDriver(domain, accessToken)
```

libs/product/driver/shopify/src/product.service.ts
```diff
@@ -110,54 +31,69 @@ export const DaffShopifyProductTransformer = (node: ProductNode): DaffProduct =>
...
+          console.log(nodes);
+          return nodes.map((node) => daffShopifyProductTransformer(node));
+        }),
+      );
   }
 
-  //todo: add actual getBestSellers apollo call. Right now, it just makes the getAll() call
+  // todo: implement getBestSellers
+  // todo: move to separate getBestSellers module
   getBestSellers(): Observable<DaffProduct[]> {
```
damienwebdev, line 58: This should be removed from the interface in v0.81, please remove this method.

damienwebdev (changes requested):

damienwebdev:
I also see two further changes:

- [x] Many of the functions are missing documentation, please add doc blocks and usage docs for each function
- [x] Needs a rebase

libs/driver/shopify/src/graphql/provide-shopify-apollo-driver.ts
leozhang14, line 1: ***In Progress. Function may need to be refactored (split into modules). Also requires additional documentation.

libs/product/driver/shopify/src/product.service.ts
leozhang14: It was deprecated in version v0.81.0, and will be removed (from the interface) in version 0.84.0. I don't believe it has been removed yet.
[Product Service Interface](https://github.com/graycoreio/daffodil/blob/develop/libs/product/driver/src/interfaces/product-service.interface.ts)
[History](https://github.com/graycoreio/daffodil/commits/13f40a416062b8ee8bbbbe44572c1678f384676e/libs/product/driver/src/interfaces/product-service.interface.ts)

griest024: yeah I just deprecated it. I think this is fine the way it is. Don't bother testing it and it will get removed when the method gets removed from the driver interface.

damienwebdev:
Superseded by #3576

PR #3474, opened 2025-02-04
feat(design)!: allow individual card type imports

### LINKED ISSUE #3462: [FEAT] Refactor Card Component
# :bulb: Feature request

## Feature Name
Refactor Card Component

## The Desired Behavior
Every card type should be its own component, rather than it all being in one single component. They still share the same css class of `.daff-card`, with an extension of the variant name, i.e. `.daff-raised-card`

## Your Use Case
As a developer, I would like to reduce the file size of the card component and unnecessary import of all card variants.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

xelaint:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #3462 

```
[x] Yes
[ ] No
```

BREAKING CHANGE: the `DAFF_CARD_COMPONENTS` array now only exports the `DaffCardComponent`. To use `DaffRaisedComponent` or `DaffStrokedComponent`, used the `DAFF_RAISED_CARD_COMPONENTS` and `DAFF_STROKED_CARD_COMPONENTS` arrays. The `DAFF_ALL_CARD_COMPONENTS` array can be used if all card types are needed.

## Other information


damienwebdev:
@xelaint needs a rebase, but I'm also surprised that this change doesn't break demo/daffio, I would have expected it to.

xelaint:
> @xelaint needs a rebase, but I'm also surprised that this change doesn't break demo/daffio, I would have expected it to.

Why would it break demo/daffio? Demo doesn't have any usages of the card component. The daffio cards all use `<daff-card>`, so the original import of `DAFF_CARD_COMPONENTS`/`DaffCardModule` is still correct.

damienwebdev:
@xelaint 

> Why would it break demo/daffio? Demo doesn't have any usages of the card component.

Indeed. I was suprised to see that the [product-card](https://github.com/graycoreio/daffodil/blob/develop/apps/demo/src/app/product/components/product-card/product-card.component.html) doesn't use a daff-card.

>  The daffio cards all use <daff-card>, so the original import of DAFF_CARD_COMPONENTS/DaffCardModule is still correct.

Ah, this is a nice point, I didn't realize it.

libs/design/card/README.md
```diff
@@ -12,7 +12,9 @@ The example below is a default, filled card that includes all of a card's pre-st
 ## Usage
 
 ### Within a standalone component
-To use card in a standalone component, import it directly into your custom component:
+To use card in a standalone component, import it directly into your custom component.
+
+The following array exports can be used, depending on which card is being used: `DAFF_CARD_COMPONENTS`, `DAFF_RAISED_COMPONENTS`, `DAFF_STROKED_CARD_COMPONENTS`, and `DAFF_ALL_CARD_COMPONENTS`.
```
damienwebdev: I don't think we need to clarify that these are arrays, we can simply say:

The following imports are available, depending on which card you are choosing to use: `DAFF_CARD_COMPONENTS`, `DAFF_RAISED_COMPONENTS`, `DAFF_STROKED_CARD_COMPONENTS`. While we don't recommend it, if you need to use all the cards types available in your component, you can use `DAFF_ALL_CARD_COMPONENTS`.

libs/design/card/src/card-base.directive.ts
```diff
@@ -0,0 +1,60 @@
...
+ * This attribute determines what orientation the the card contents are stacked.
+ */
+export type DaffCardOrientation = 'vertical' | 'horizontal';
+export enum DaffCardOrientationEnum {
+  Vertical = 'vertical',
+  Horizontal = 'horizontal',
+}
+
+@Directive({
+  selector: '[daffCardBase]',
```
damienwebdev, line 22: Is it possible to remove this selector? I wouldn't want someone to accidentally use this directive.

damienwebdev (changes requested):

libs/design/card/README.md
xelaint: `While we don't recommend it` leaving this out because if you are using all card types within a single component, `DAFF_ALL_CARD_COMPONENTS` would be the correct import to use.

xelaint: updated

libs/design/card/src/card-base.directive.ts
xelaint: don't need to remove since this directive is not exposed in the public api.

PR #3523, opened 2025-02-24
feat(design)!: update breadcrumb active state implementation

### LINKED ISSUE #3514: [FEAT] Set the last breadcrumb item to be active by default
# :bulb: Feature request

## Feature Name
Set the last breadcrumb item to be active by default

## The Desired Behavior
The last breadcrumb item in a list of breadcrumbs should always be the active item. The `active` input property should no longer be an input property as you can't decide on which breadcrumb should be the active one.

The purpose of the breadcrumb component is to always show the last one as the current page, so the current implementation doesn't fully make sense, allowing a user to decide what is active.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

xelaint:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Enduser is allowed to use the `active` property, but it should just default to the last breadcrumb item.

Fixes: #3514 

## What is the new behavior?
The active state is automatically set to the last breadcrumb item in a list of breadcrumbs.

```
[x] Yes
[ ] No
```

BREAKING CHANGE: T `active` property is no longer available to use. The last breadcrumb item will automatically be set to the active state.

## Other information

### DIFF
--- a/libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.directive.ts
+++ b/libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.directive.ts
@@ -1,8 +1,7 @@
 import {
+  ChangeDetectorRef,
   Directive,
-  ElementRef,
   HostBinding,
-  Renderer2,
 } from '@angular/core';
 
 @Directive({
@@ -14,21 +13,28 @@ export class DaffBreadcrumbItemDirective {
    */
   @HostBinding('class.daff-breadcrumb__item') class = true;
 
+  /**
+   * @docs-private
+   */
+  @HostBinding('class.active') get activeClass() {
+    return this._active;
+  }
+
+  /**
+   * @docs-private
+   */
+  @HostBinding('attr.aria-current') get ariaCurrent() {
+    return this._active ? 'page' : null;
+  }
+
   private _active = false;
 
-  constructor(
-    private elementRef: ElementRef,
-    private renderer: Renderer2,
-  ) {}
+  constructor( private cdRef: ChangeDetectorRef ) {}
 
   /** Called by the DaffBreadcrumbComponent to set the active state */
   setActive(value: boolean) {
     this._active = value;
-    if (value) {
-      this.renderer.addClass(this.elementRef.nativeElement, 'active');
-      this.renderer.setAttribute(this.elementRef.nativeElement, 'aria-current', 'page');
-    } else {
-      this.renderer.removeClass(this.elementRef.nativeElement, 'active');
-    }
+
+    this.cdRef.detectChanges();
   }
 }
--- a/libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.directive.ts
+++ b/libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.directive.ts
@@ -1,21 +1,34 @@
 import {
   Directive,
+  ElementRef,
   HostBinding,
-  Input,
+  Renderer2,
 } from '@angular/core';
 
 @Directive({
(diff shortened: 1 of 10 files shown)


libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -24,9 +29,35 @@ import {
...
+   * @docs-private
+   */
+  @ContentChildren(DaffBreadcrumbItemDirective) breadcrumbItems!: QueryList<DaffBreadcrumbItemDirective>;
+
+  /**
+   * @docs-private
+   */
+  ngAfterContentInit() {
+    this.updateActiveState();
+    this.breadcrumbItems.changes.subscribe(() => this.updateActiveState());
```
damienwebdev: This needs to be takeUntilDestroyed to prevent memory leaks.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -24,9 +29,35 @@ import {
...
+  }
+
+  private updateActiveState() {
+    if (!this.breadcrumbItems.length) {
+      return;
+    }
+
+    // Sets only the last breadcrumb item as active
+    this.breadcrumbItems.forEach((item, index) => {
```
damienwebdev: This loop is unnecessary, you can just do `this.breadcrumbItems.last.setActive(true)`

damienwebdev (changes requested):

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
xelaint: updated

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
xelaint: updated

libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.directive.ts
```diff
@@ -1,21 +1,34 @@
...
+  constructor(
+    private elementRef: ElementRef,
+    private renderer: Renderer2,
+  ) {}
 
+  /** Called by the DaffBreadcrumbComponent to set the active state */
+  setActive(value: boolean) {
+    this._active = value;
+    if (value) {
+      this.renderer.addClass(this.elementRef.nativeElement, 'active');
```
damienwebdev: I think it makes more sense to keep the original implementation and just remove the original `@Input`. I think access to the renderer should be used sparingly.

damienwebdev (changes requested):

libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.directive.ts
xelaint: updated

damienwebdev:
@xelaint can you add an example with an `@for`? I've seen afterContentInit operations like this new code performs have errors when the child content is inside an `@for`.

xelaint:
@damienwebdev Added the example. The implementation in next.daff.io/docs also uses `@for` and it seems to be fine w/ no errors.

damienwebdev (approved):
lgtm

PR #3777, opened 2025-06-12
feat(category,product): resolve entity with `route` query

griest024:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[x] Yes
[ ] No
```

URL truncation configuration has been removed from `@daffodil/product/driver/magento` and `@daffodil/category/driver/magento` and moved to `@daffodil/driver`

## Other information


libs/driver/src/url-truncation/service.ts
```diff
@@ -0,0 +1,25 @@
+import {
```
damienwebdev: Is this service necessary? Can we use use a factory provider to make a function?

damienwebdev (changes requested):

libs/driver/src/url-truncation/service.ts
griest024: a factory on another token? sure, but that could be said of any single method service. Do you prefer that pattern here?

damienwebdev: I feel like this particular service is unlikely to ever have additional methods, which makes me feel that the service is overkill.

griest024: okay changed

libs/product/driver/magento/src/product.service.spec.ts
```diff
@@ -135,7 +135,7 @@ describe('Product | Magento | ProductService', () => {
 
       const op = controller.expectOne(addTypenameToDocument(getProductByUrl()));
 
-      expect(op.operation.variables.url).toEqual('TESTING_URL');
+      expect(op.operation.variables.url).toEqual('path/to/TESTING_URL');
```
griest024: @damienwebdev we shouldn't need to truncate leading path segments for products, right? I can't remember why this behavior was necessary

griest024: bump^

damienwebdev: I believe this does matter since Magento's GraphQl product lookup is by url_key which does not include category paths for example.

Think:

`furniture/chairs/brown-chair.html`

vs.

`brown-chair.html`.

This is how the current driver works. Perhaps we could use the `route` node instead in the product driver to obviate this need?

griest024: So `DaffMagentoProductService.getByUrl` should successfully return for `furniture/chairs/brown-chair.html` even though the platform doesn't recognize that as a product?

damienwebdev: Well, it is the `url` of a specific product, albeit with some leading paths.

So, just for clarity because as its hard to hold this mental model in my head.

Current State (with URL truncation)
==

We need to validate that this is exactly what the driver does currently, but I believe it is.

```ts
//URL Key - works
DaffMagentoProductService.getByUrl("brown-chair.html");

// Path to product in category - works
DaffMagentoProductService.getByUrl("chairs/brown-chair.html");

// Not a real path -- works
DaffMagentoProductService.getByUrl("tacos/brown-chair.html");
```

Proposed State (without URL truncation)
==

```ts
//URL Key - works
DaffMagentoProductService.getByUrl("brown-chair.html");

// Path to product in category - works
DaffMagentoProductService.getByUrl("chairs/brown-chair.html");

// Not a real path -- 404
DaffMagentoProductService.getByUrl("tacos/brown-chair.html");
```

I believe the proposed state is better. I think we should use the `routes` query to perform the product lookup and remove the "remove leading path" behavior.

libs/driver/src/url-truncation/config-default.ts
```diff
@@ -0,0 +1,10 @@
+import { DaffDriverUrlTruncationConfig } from './config.type';
+import { daffDriverTruncateUrl } from './transform';
+
+/**
+ * The default configuration for the {@link DaffDriverUrlTruncationConfig}.
+ */
+export const DAFF_DRIVER_URL_TRUNCATION_CONFIG_DEFAULT: DaffDriverUrlTruncationConfig = {
```
damienwebdev: Do you think this token should be done on a platform by platform basis?

griest024: that's what I'm considering. Based on what the product driver needs, probably yes

libs/driver/src/url-truncation/config-default.ts
```diff
@@ -0,0 +1,10 @@
...
+  truncateUrl: true,
+  uriTruncationStrategy: daffDriverTruncateUrl,
```
damienwebdev: I think the default strategy should be a noop strategy.

libs/driver/src/url-truncation/config-default.ts
griest024: That leads me to believe that we shouldn't even have configurable URL truncation. Just have the driver transform the URL from the form that it should be in daffodil (relative to the domain root with a leading slash, i.e. `/path/to/entity`) to the form that the platform needs it.

Or is the form that the platform needs it a data issue? Will Magento always resolve `path/to/entity` for the `route` query or might it need `path/to/entity.html` or `/path/to/entity.html`, depending on the settings?

libs/product/driver/magento/src/product.service.spec.ts
griest024: changed to use the route query

PR #3781, opened 2025-06-13
feat(core): add validator type and helper

griest024:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/libs/core/src/validator/field-presence.ts
+++ b/libs/core/src/validator/field-presence.ts
@@ -1,2 +1,7 @@
-export const validateFieldPresence = <T = unknown>(model: T, ...fields: Array<keyof T>): boolean =>
-  fields.reduce((acc, field) => acc && !!model[field], true);
+/**
+ * Validates whether all the specified fields are defined in the passed object.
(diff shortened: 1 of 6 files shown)


libs/core/src/validator/field-presence.ts
```diff
@@ -0,0 +1,2 @@
+export const validateFieldPresence = <T = unknown>(model: T, ...fields: Array<keyof T>): boolean =>
```
damienwebdev: I think this fails when the field exists but is false.

damienwebdev (changes requested):

libs/core/src/validator/field-presence.ts
griest024: good catch, fixed

damienwebdev (approved):

PR #3785, opened 2025-06-13
feat(design): create a DaffStickyTracker to track when an element is pinned

### LINKED ISSUE #2825: [FEAT] Directive to track when a sticky element gets pinned
# :bulb: Feature request

## Feature Name
Directive to track when a sticky element gets pinned

## The Desired Behavior
A directive (DaffStickyTracker) to calculate when a sticky elements gets pinned so it can add and remove a `.is-pinned` class to the element. This is so that we can change styles when it's sticky and when it's not.

i.e. when the footer buttons are sticky, there should be a top box shadow on the element. When it's no longer sticky and back as a relative element on the page, the box shadow should be removed.

## Your Use Case
As a designer, I would like to be able to add and remove styles on sticky elements.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

AmmarA06:
## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit
- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Fixes: #2825 

## What is the new behavior?
The new behavior adds a .is-pinned class to a sticky element when it becomes pinned and removes it when it's not, allowing for different styles in each state.

```
[ ] Yes
[X] No
```


xelaint (changes requested):
Can you move the demo to `design-land`?

libs/design/src/core/sticky/daffStickyTracker-test.html
```diff
@@ -0,0 +1,243 @@
+<!DOCTYPE html>
```
xelaint: I don't see the `daffStickyTracker` directive being used in this demo. Are you just duplicating the directive code in the script?

If so, the demo needs to be moved to `design-land` so you can demo the `DaffStickyTrackerDirective` in a custom component.

xelaint (changes requested):

libs/design/src/core/sticky/sticky-tracker.directive.ts
```diff
@@ -0,0 +1,120 @@
...
+
+  private setupStickyElement(): void {
+    const element = this.elementRef.nativeElement;
+    const computedStyle = this.document.defaultView?.getComputedStyle(element);
+
+    if (computedStyle.position !== 'sticky') {
+      this.renderer.setStyle(element, 'position', 'sticky');
+    }
+
+    if (!element.style.top && (!computedStyle.top || computedStyle.top === 'auto')) {
```
xelaint: The directive also needs to have the option to pin an element to the bottom of the screen.

e.g. https://www.lostgolfballs.com/checkout/cart on a mobile view: when I scroll down to the checkout and paypal buttons, the buttons should be pinned to the bottom of the screen as I continue to scroll down.

xelaint (changes requested):

libs/design/src/core/sticky/sticky-tracker.directive.ts
AmmarA06: Should I just add an input to the directive to control whether the element is pinned to the top or bottom?

damienwebdev: I believe this should be decided automatically based upon the position set on the element, i.e. if top: 0, then top, if bottom: 0 then bottom. If they're both set, how does the native browser position sticky behave?

damienwebdev:
@xelaint needs your review.

libs/design/src/core/sticky/sticky-tracker.directive.ts
```diff
@@ -0,0 +1,216 @@
...
+    private readonly renderer: Renderer2,
+    private readonly ngZone: NgZone,
+    @Inject(DOCUMENT) private readonly document: Document,
+  ) {}
+
+  /**
+   * @docs-private
+   */
+  ngAfterViewInit(): void {
+    this.ngZone.runOutsideAngular(() => {
```
damienwebdev: I think, instead of doing this outside the zone, we should do this with a `afterNextRender` callback. Can you do a performance profile of this before/after this change?

damienwebdev (changes requested):

libs/design/src/core/sticky/sticky-tracker.directive.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  private createSentinel(): void {
+    const stickyElement = this.elementRef.nativeElement;
+    const parentElement = stickyElement.parentElement;
+
+    if (!parentElement) {
+      throw new Error('DaffStickyTracker: Sticky element must have a parent element');
+    }
+
+    this.sentinelElement = this.renderer.createElement('div');
```
xelaint, line 82: What is the purpose of adding this empty div?

AmmarA06: It creates a sentinel element that is tracked by the `createSentinelObserver()` method. Based on the sentinel's visibility, the observer adds or removes the `is-pinned` class from the sticky element to indicate its pinned state.

xelaint (approved):

damienwebdev (approved):
lgtm

PR #3853, opened 2025-07-23
fix(dgeni): fix Windows path separators and enable sequential execution for Windows

AmmarA06:
## PR Checklist

- [X] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit

## PR Type

```
[X] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Illegal characters in path/invalid path construction issues occur due to improper string replacement, and multiple dgeni processes running in parallel caused Windows file locking errors.

## What is the new behavior?
Dgeni processes now run in sequence on Windows to prevent file locking errors, and path.relative() properly handles Windows vs Unix path differences and absolute path calculations.

```
[ ] Yes
[X] No
```

### DIFF
--- a/tools/dgeni/build.ts
+++ b/tools/dgeni/build.ts
@@ -16,37 +16,68 @@ import { daffodilRoutesPackage } from './src/transforms/daffodil-routes-package'
 
 rimraf('../../dist/docs/*', { glob: true }).then(() => {
   new Dgeni([apiDocs]).generate().then(() => {
-    // base docs
-    Promise.all([
-      new Dgeni([packageDocsPackage]).generate().catch((err) => {
-        console.log(err);
-        process.exit(1);
-      }),
-      new Dgeni([guideDocsPackage]).generate().catch((err) => {
-        console.log(err);
-        process.exit(1);
-      }),
-      new Dgeni([explanationDocsPackage]).generate().catch((err) => {
-        console.log(err);
-        process.exit(1);
-      }),
-    ]).then(() => {
+    // base docs 
+    const runBaseDocsSequentially = async () => {
+      if (process.platform === 'win32') {
+        // Sequential execution on Windows
+        try {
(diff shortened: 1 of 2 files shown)


damienwebdev:
@AmmarA06 can you explain the illegal path combinations? `runDesignDocsSequentially` sucks to have to do on Windows, but I understand it.

AmmarA06:
> @AmmarA06 can you explain the illegal path combinations? `runDesignDocsSequentially` sucks to have to do on Windows, but I understand it.

The issue was that dgeni was trying to create file paths like:
`C:\Users\USER\Downloads\Graycore\daffodil\dist\docs\search-index\api\C:\Users\USER\Downloads\Graycore\daffodil\libs\auth\driver\src`

The `outputPathsConfigurator` was using `API_SOURCE_PATH` for all document types, but guide documents have different IDs than API docs. i.e API docs would be the full path and guide docs would just be the file name. So, when `path.relative("libs", "introduction")` was called, it returned `"../introduction"`, creating illegal paths.

Now guides use `docs/guides` as the base path, resulting in proper relative paths like `introduction` instead of `../introduction`.

griest024 (changes requested):
This causes issues on linux, the API doc paths are prefixed with `tools/dgeni`. I need more time to dig into this but likely the part of the configurator that is causing issues `replace(API_SOURCE_PATH, '')` will be moved into the API transform code. The configurator really shouldn't be dealing with API (or any kind/role) specific behaviors.

griest024:
If you open a PR for the windows file lock stuff without the path stuff you can get it merged without being blocked by this path bug.

damienwebdev:
Closing until we need this. Unfortunately, I don't build on windows so this is hard to test.

PR #4020, opened 2025-09-26
build: set `module: "bundler"` in angular libs

griest024:
## PR Checklist

## PR Type

- [ ] Bug fix
- [ ] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [x] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #

## Breaking change?

- [ ] Yes
- [ ] No

### DIFF
--- a/libs/core/state/src/operation/entity/selectors.ts
+++ b/libs/core/state/src/operation/entity/selectors.ts
@@ -1,9 +1,9 @@
-import type { EntitySelectors } from '@ngrx/entity/src/models';
 import {
   createSelector,
   defaultMemoize,
   MemoizedSelector,
 } from '@ngrx/store';
+import type { EntitySelectors } from 'node_modules/@ngrx/entity/src/models';
 
 import { DaffIdentifiable } from '@daffodil/core';
 
--- a/libs/core/state/src/operation/entity/selectors.ts
+++ b/libs/core/state/src/operation/entity/selectors.ts
@@ -1,13 +1,17 @@
+import { EntityAdapter } from '@ngrx/entity';
 import {
   createSelector,
   defaultMemoize,
   MemoizedSelector,
 } from '@ngrx/store';
-import type { EntitySelectors } from 'node_modules/@ngrx/entity/src/models';
 
 import { DaffIdentifiable } from '@daffodil/core';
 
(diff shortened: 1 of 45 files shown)


libs/core/state/src/operation/entity/selectors.ts
```diff
@@ -1,9 +1,9 @@
-import type { EntitySelectors } from '@ngrx/entity/src/models';
 import {
   createSelector,
   defaultMemoize,
   MemoizedSelector,
 } from '@ngrx/store';
+import type { EntitySelectors } from 'node_modules/@ngrx/entity/src/models';
```
damienwebdev: This look very wrong, otherwise LGTM

damienwebdev (changes requested):

libs/core/state/src/operation/entity/selectors.ts
griest024: Well really we shouldn't ever be importing anything from `src`. The reason why this change is required is because `bundler` respects the packages `exports` field, which obviously doesn't list `src` as an export. There *is* a fairly complex way of extracting this type that doesn't rely on an internal import, should I do that?

damienwebdev:
Does: https://github.com/graycoreio/daffodil/blob/develop/libs/driver/hubspot/testing/src/factories/hubspot-response.factory.spec.ts#L3 need fixed in this as well?

griest024:
> Does: https://github.com/graycoreio/daffodil/blob/develop/libs/driver/hubspot/testing/src/factories/hubspot-response.factory.spec.ts#L3 need fixed in this as well?

oddly no, I would have expected that to fail here. I'll upstream that change separately since its strictly better.

libs/core/state/src/operation/entity/selectors.ts
damienwebdev: Yes, It's a PITA to read I'd imagine, but it's worth not having to pull a type directly from a node_modules path.

griest024: okay done

damienwebdev (approved):

PR #4142, opened 2025-10-19
fix(design): paginator is hydration incompatible

### LINKED ISSUE #3426: Make design components hydration compatible
We should avoid direct DOM manipulation in design components as this will [break hydration](https://angular.dev/guide/hydration#direct-dom-manipulation). ~~Its also an open issue that [template outlets break hydration and require a workaround](https://github.com/angular/angular/issues/50543)~~ fixed in angular 20. This issue should track the various instances of design components that currently have hydration-incompatible implementations:

- paginator: https://github.com/graycoreio/daffodil/blob/25766422ef010b4c336025e56570025d0ac24c9c/libs/design/paginator/src/paginator/paginator.component.ts#L71-L72

griest024:
## PR Checklist

## PR Type

- [ ] Bug fix
- [ ] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #3426
Part of: # 

## Breaking change?

- [ ] Yes
- [ ] No

### DIFF
--- a/libs/design/paginator/src/paginator/paginator.component.html
+++ b/libs/design/paginator/src/paginator/paginator.component.html
@@ -2,7 +2,7 @@
   <button type="button" class="daff-paginator__previous"
     [disabled]="_disablePrev"
     tabindex="0"
-    attr.aria-label="Go to Previous Page of {{_paginatorId}} Paginator"
+    attr.aria-label="Go to Previous Page of {{ariaLabel}} Paginator"
     (click)="_onNotifyPrevPageChange()">
     <fa-icon [icon]="faChevronLeft" size="sm"></fa-icon> Previous
   </button>
@@ -10,7 +10,7 @@
 @if (linkMode) {
   @if (!_disablePrev) {
     <a class="daff-paginator__previous"
-      attr.aria-label="Go to Previous Page of {{_paginatorId}} Paginator"
+      attr.aria-label="Go to Previous Page of {{ariaLabel}} Paginator"
       [routerLink]="url"
       queryParamsHandling="merge"
       [queryParams]="_buildPageQueryParams(currentPage - 1)">
@@ -19,7 +19,7 @@
   }
   @if (_disablePrev) {
     <span class="daff-paginator__previous disabled"
-      attr.aria-label="Go to Previous Page of {{_paginatorId}} Paginator"
+      attr.aria-label="Go to Previous Page of {{ariaLabel}} Paginator"
       [attr.disabled]="true">
       <fa-icon [icon]="faChevronLeft" size="sm"></fa-icon><span>Previous</span>
     </span>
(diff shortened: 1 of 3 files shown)


libs/design/paginator/src/paginator/paginator.component.ts
```diff
@@ -86,6 +75,11 @@ export class DaffPaginatorComponent implements OnChanges {
    */
   @Input() queryParam = 'page';
 
+  /**
+   * The aria label.
+   */
+  @Input() ariaLabel: string;
```
damienwebdev: Can this be `aria-label` instead? Will it automatically pick up the label using the native attribute?

damienwebdev (changes requested):

libs/design/paginator/src/paginator/paginator.component.ts
griest024: no input names should not collide with DOM attributes. However I had forgotten to add the host binding so I've done that

xelaint: Can you set this the HostBinding on the `host` property instead?

libs/design/paginator/src/paginator/paginator.component.html
```diff
@@ -2,15 +2,15 @@
   <button type="button" class="daff-paginator__previous"
     [disabled]="_disablePrev"
     tabindex="0"
-    attr.aria-label="Go to Previous Page of {{_paginatorId}} Paginator"
+    attr.aria-label="Go to Previous Page of {{ariaLabel}} Paginator"
```
damienwebdev: I was looking at various paginator implementations and I believe that adding the label in here is non-normative.

https://a11ymatters.com/pattern/pagination/
https://design-system.w3.org/components/pagination.html

I believe this should solely read: "Go to Previous Page".

@xelaint

damienwebdev (changes requested):

libs/design/paginator/src/paginator/paginator.component.html
xelaint: Yeah, I don't think the aria-label is necessary in the component. `aria-label` should not be set on `<daff-paginator>` like it currently is. Instead, it needs to be wrapped in a `<nav>` element where the `aria-label` is set on it.

This is more of a documentation issue since:

Current example is:

```
<daff-paginator aria-label="Search results page" [numberOfPages]="numberOfPages" [currentPage]="currentPage" (notifyPageChange)="onPageChange($event)"></daff-paginator>
```

Correct example should be:

```
<nav aria-label="Search Results Pagination">
  <daff-paginator [numberOfPages]="numberOfPages" [currentPage]="currentPage" (notifyPageChange)="onPageChange($event)"></daff-paginator>
</nav>
```

damienwebdev: Ok. To finalize, the appropriate change here would be to make the buttons say:

"Go to page 'x'"
"Go to next page"
"Go to previous page"

And any reference to the aria label would be removed.

griest024: changed

damienwebdev (approved):
Lgtm

PR #4144, opened 2025-10-26
feat(driver): add routable objects system for in-memory drivers

### LINKED ISSUE #4143: Automatically configured in-memory routes
### Feature description

Improving the InMemoryExternal Router Api

## The problem

Currently, the in-memory external router's API is essentially a DIY `DaffExternalRouterDriverInMemoryResolver` that you're forced to configure if you want to resolve routes to entities in your app. 

This configuration is provided via `provideDaffExternalRouterInMemoryDriver`'s config argument as follows (in its simplest form):

```ts
const myResolverFunction: DaffExternalRouterDriverInMemoryResolver = (url: string) => {
    throw new DaffExternalRouterNotFoundError();
}
export const appConfig: ApplicationConfig = {
  providers: [
    ...,
    provideDaffExternalRouterInMemoryDriver(config: { }),
  ]
};
```

It can also be configured like:

```ts
export const EXAMPLE_EXTERNAL_ROUTER_DRIVER_IN_MEMORY_CONFIG
  = new InjectionToken<DaffExternalRouterDriverInMemoryConfig>('EXAMPLE_EXTERNAL_ROUTER_DRIVER_IN_MEMORY_CONFIG', {
    factory: () => {
        return {
            resolver: (url: string) => {
                throw new DaffExternalRouterNotFoundError();
            }
        }
      };
    },
  );
```

None of the other `@daffodil/*/driver/in-memory` packages have any interplay with this resolver function out of the the box, resulting in manual configuration each time a new kind of entity/package is added. This is onerous, verbose, and likely entirely overhead that can be removed from user-facing apps.

## The goal

What I would like to see is that the default configuration provided by `provideDaffExternalRouterInMemoryDriver()` with no arguments utilizes some sort of default `resolver` function (that's also in the public API so that others can customize it further if necessary). In this way, if someone were to do something like:

```ts
export const appConfig: ApplicationConfig = {
  providers: [
    ...,
    provideExternalRouter(),
    provideDaffInMemoryDriver({ apiBase: 'daff-in-memory-web-api', passThroughUnknownRequests: true }),
    provideDaffProductInMemoryDriver(),
    provideDaffNavigationInMemoryDriver(),
    provideDaffExternalRouterInMemoryDriver(),
  ]
};  
```

This configuration would be sufficient to enable routing to the randomly generated entity data provided by the other drivers (presuming that the user's `routes` support that data).

### Use case

Simplified InMemoryDriver setup for `@daffodil/external-router`.

### Prior work

N/A

### Daffodil version

0.90.0

### Additional context

_No response_

damienwebdev:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior
See #4143 -- in short, making the in-memory drivers routable is annoying and tedious.

Fixes: #4143

## New behavior
Introduces a new routing system that allows in-memory drivers to register their entities with URLs, enabling automatic URL resolution by the @daffodil/external-router or another similar package. This provides a centralized way for in-memory drivers to contribute routable objects to the application.

## Breaking change?

- [ ] Yes
- [x] No

### DIFF
--- a/libs/driver/in-memory/src/routing/provider.ts
+++ b/libs/driver/in-memory/src/routing/provider.ts
@@ -12,7 +12,7 @@ import {
 
 /**
  * A function that returns an array of entities with URL properties.
- * Used to provide routable objects to the in-memory routing system.
+ * Used to provide routable in-memory object information to the routing system.
  */
 export type DaffInMemoryRoutableObjectsResolver = () => Array<{ url: string }>;
 
@@ -25,8 +25,11 @@ export type DaffInMemoryRoutableObjectsResolver = () => Array<{ url: string }>;
  *
  * @docs-private
  */
-const DAFF_INMEMORY_ROUTABLE_OBJECTS_RESOLVER = new InjectionToken<{ type: string; resolver: InjectionToken<DaffInMemoryRoutableObjectsResolver> }>(
-  'DAFF_INMEMORY_ROUTABLE_OBJECTS_RESOLVER',
+const DAFF_INMEMORY_ROUTABLE_OBJECTS_RESOLVER = new InjectionToken<{ type: string; resolver: DaffInMemoryRoutableObjectsResolver }[]>(
+  'DAFF_INMEMORY_ROUTABLE_OBJECTS_RESOLVER', {
+    providedIn: 'root',
+    factory: () => [],
+  },
 );
 
 /**
@@ -50,9 +53,8 @@ export const DAFF_INMEMORY_ROUTABLE_OBJECTS = new InjectionToken<DaffInMemoryRou
         throw new Error('DAFF_INMEMORY_ROUTABLE_OBJECTS_RESOLVER must be an array');
       }
 
-      resolvers.forEach((entry: { type: string; resolver: InjectionToken<DaffInMemoryRoutableObjectsResolver> }) => {
-        const resolverFn = inject(entry.resolver);
-        const entities = resolverFn();
+      resolvers.forEach((entry) => {
+        const entities = entry.resolver();
         entities.forEach(entity => {
           if (map.has(entity.url)) {
             const existing = map.get(entity.url);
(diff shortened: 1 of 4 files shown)


libs/driver/in-memory/src/routing/provider.ts
```diff
@@ -0,0 +1,124 @@
...
+ * export const appConfig: ApplicationConfig = {
+ *   providers: [
+ *     provideDaffInMemoryRoutableObjects("PRODUCTS", DAFF_PRODUCT_INMEMORY_ROUTABLE_OBJECTS),
+ *   ]
+ * };
+ * ```
+ */
+export const provideDaffInMemoryRoutableObjects = (
+  type: string,
+  resolver: InjectionToken<DaffInMemoryRoutableObjectsResolver>,
```
griest024: why provide a token of the value instead of just the value?

libs/driver/in-memory/src/routing/routable-object.ts
```diff
@@ -0,0 +1,22 @@
...
+   * The type identifier for this routable object (e.g., "PRODUCTS", "NAVIGATION").
+   */
+  type: string;
+}
+
+/**
+ * A map of URL paths to their corresponding routable objects.
+ * The key is the URL path.
+ */
+export type DaffInMemoryRoutableObjects = Map<string, DaffInMemoryRoutableObject>;
```
griest024: Its probably a slight improvement to type this as a `ReadOnlyMap`.

griest024 (changes requested):

libs/driver/in-memory/src/routing/routable-object.ts
damienwebdev: https://github.com/graycoreio/daffodil/pull/4144/commits/5ca3d4a7ebaca882349509126a4c3a0eb3bce7ed

libs/driver/in-memory/src/routing/provider.ts
damienwebdev: Good catch. I didn't realise that it was running in an injection context when I wrote it and that I could use `inject`. https://github.com/graycoreio/daffodil/pull/4144/commits/1644f2ccb8a0643af53f27bc36097fbab91686c8

libs/driver/in-memory/src/routing/provider.ts
```diff
@@ -0,0 +1,123 @@
...
+    factory: () => {
+      const resolvers = inject(DAFF_INMEMORY_ROUTABLE_OBJECTS_RESOLVER, { optional: true }) || [];
+      const map = new Map<string, DaffInMemoryRoutableObject>();
+
+      if (!Array.isArray(resolvers)) {
+        throw new Error('DAFF_INMEMORY_ROUTABLE_OBJECTS_RESOLVER must be an array');
+      }
+
+      resolvers.forEach((entry) => {
+        const entities = entry.resolver();
```
griest024, line 57: I think you need to invoke it with `runInInjectionContext(entry.resolver)`. AFAIK the injection context does not automatically transfer like this.

griest024 (changes requested):

libs/driver/in-memory/src/routing/provider.ts
damienwebdev: I did try this locally in the demo app and it did appear to `inject` as expected. I was surprised by this as well.

griest024: hmmm did they change this in Angular 20 perhaps? that's a big change (but welcome)

damienwebdev: Perhaps. I believe since it's called within a factory, this may have always been the case, but I'd need to try it out on older versions to be sure.

griest024: huh yeah, testing in a playground indicates that injection context survives in child functions.

griest024 (approved):

PR #4147, opened 2025-10-26
fix(product): getByUrl on DaffInMemoryProductService works with urls

damienwebdev:
## PR Checklist

## PR Type

- [x] Bug fix
- [ ] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior
Currently, `getByUrl` will always fail with legitimate paths since we currently assume that a path is just an `id`. This is a confusing and mistaken assumption. 

Part of: #4143 

## New behavior
I removed the stale `best-sellers` functionality since it's entirely unused in the framework. Any product `id` that contains the string `.html` will be treated as a `url` lookup. Everything else will still continue to be treated as an `id`. Additionally, product urls stored in the in-memory backend will now appropriately be URLs.

## Breaking change?

- [ ] Yes
- [x] No

### DIFF
--- a/libs/product/driver/in-memory/src/backend/product.service.ts
+++ b/libs/product/driver/in-memory/src/backend/product.service.ts
@@ -1,20 +1,21 @@
-import { Injectable } from '@angular/core';
+import {
+  inject,
+  Injectable,
+} from '@angular/core';
 import {
   InMemoryDbService,
   RequestInfoUtilities,
   ParsedRequestUrl,
+  RequestInfo,
   STATUS,
 } from 'angular-in-memory-web-api';
 
+import { daffUriTruncateLeadingSlash } from '@daffodil/core/routing';
 import { DaffInMemorySingleRouteableBackend } from '@daffodil/driver/in-memory';
 import { DaffProduct } from '@daffodil/product';
-import {
-  DaffProductImageFactory,
-  DaffProductExtensionFactory,
-} from '@daffodil/product/testing';
+import { DaffProductExtensionFactory } from '@daffodil/product/testing';
 
 import { DAFF_PRODUCT_IN_MEMORY_COLLECTION_NAME } from '../collection-name.const';
-
 /**
  * An in-memory service that stubs out the backend services for getting products.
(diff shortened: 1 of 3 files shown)


libs/product/driver/in-memory/src/backend/product.service.ts
```diff
@@ -37,46 +38,10 @@ export class DaffInMemoryBackendProductService implements InMemoryDbService, Daf
...
-      '1031',
-      '1032',
-      '1033',
-      '1034',
-      '1035',
-    ].map(id => this.productFactory.create({ id, url: `/${id}` }));
+  productFactory = inject(DaffProductExtensionFactory);
+
+  constructor(){
+    this._products = this.productFactory.createMany(35);
```
griest024, line 44: It seems that this will conflict with the transfer state PR

damienwebdev: It will. I'll adjust depending on which is merged first.

griest024 (approved):

libs/product/driver/in-memory/src/backend/product.service.ts
```diff
@@ -109,14 +74,19 @@ export class DaffInMemoryBackendProductService implements InMemoryDbService, Daf
...
    * @returns An http response object
    */
-  get(reqInfo: any) {
-    if (reqInfo.id === 'best-sellers') {
+  get(reqInfo: RequestInfo) {
+    if(reqInfo.id?.includes('.html')) {
+      const product =  this._products.find((p) => daffUriTruncateLeadingSlash(p.url)  === reqInfo.id);
+
+      if(!product) {
+        return undefined;
```
griest024, line 82:
```suggestion
        return reqInfo.utils.createResponse$(() => ({
          status: STATUS.NOT_FOUND,
        }))
```

griest024 (changes requested):

PR #4148, opened 2025-10-26
feat(product): transfer in-memory state when hydrating backend 

damienwebdev:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Part of: #4143 

## New behavior
Product backend data is transferred when hydrating backend from server to client.

## Breaking change?

- [ ] Yes
- [x] No

## Additional context

Depends on #4147

### DIFF
--- a/libs/product/driver/in-memory/src/backend/product.service.ts
+++ b/libs/product/driver/in-memory/src/backend/product.service.ts
@@ -55,6 +55,7 @@ export class DaffInMemoryBackendProductService implements InMemoryDbService, Daf
         ).data,
       );
     } else {
+      this._products = productFactory.createMany(35);
       transferState.set(TRANSFER_STATE_KEY, { data: JSON.stringify(this._products) });
     }
   }
--- a/libs/product/driver/in-memory/src/backend/product.service.ts
+++ b/libs/product/driver/in-memory/src/backend/product.service.ts
@@ -1,6 +1,9 @@
+import { isPlatformBrowser } from '@angular/common';
 import {
   inject,
   Injectable,
+  PLATFORM_ID,
+  TransferState,
 } from '@angular/core';
 import {
   InMemoryDbService,
@@ -16,6 +19,7 @@ import { DaffProduct } from '@daffodil/product';
 import { DaffProductExtensionFactory } from '@daffodil/product/testing';
 
 import { DAFF_PRODUCT_IN_MEMORY_COLLECTION_NAME } from '../collection-name.const';
+import { TRANSFER_STATE_KEY } from './transfer-state-key';
 /**
  * An in-memory service that stubs out the backend services for getting products.
  *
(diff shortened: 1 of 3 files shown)


libs/product/driver/in-memory/src/backend/product.service.ts
```diff
@@ -37,46 +42,23 @@ export class DaffInMemoryBackendProductService implements InMemoryDbService, Daf
...
-      '1032',
-      '1033',
-      '1034',
-      '1035',
-    ].map(id => this.productFactory.create({ id, url: `/${id}` }));
+  transferState = inject(TransferState);
+
+  platform = inject(PLATFORM_ID);
+
+  productFactory = inject(DaffProductExtensionFactory);
```
griest024: It doesn't seem like these need to be fields of this class, just inject in constructor and use locally.

libs/product/driver/in-memory/src/backend/product.service.ts
```diff
@@ -37,46 +42,23 @@ export class DaffInMemoryBackendProductService implements InMemoryDbService, Daf
...
+  constructor(){
+    if(isPlatformBrowser(this.platform)) {
+      this._products = JSON.parse(
+        this.transferState.get(
+          TRANSFER_STATE_KEY,
+          { data: JSON.stringify(this.productFactory.createMany(35)) },
+        ).data,
+      );
+    } else {
+      this.transferState.set(TRANSFER_STATE_KEY, { data: JSON.stringify(this._products) });
```
griest024: Where are the products actually being created?

libs/product/driver/in-memory/src/backend/product.service.ts
```diff
@@ -37,46 +42,23 @@ export class DaffInMemoryBackendProductService implements InMemoryDbService, Daf
...
+  platform = inject(PLATFORM_ID);
+
+  productFactory = inject(DaffProductExtensionFactory);
+
+  constructor(){
+    if(isPlatformBrowser(this.platform)) {
+      this._products = JSON.parse(
+        this.transferState.get(
+          TRANSFER_STATE_KEY,
+          { data: JSON.stringify(this.productFactory.createMany(35)) },
```
griest024: Do we need to do serialization ourselves? `TransferState` should handle that:
```
The values in the store are serialized/deserialized using JSON.stringify/JSON.parse. So only boolean, number, string, null and non-class objects will be serialized and deserialized in a non-lossy manner.
```

griest024 (changes requested):

libs/product/driver/in-memory/src/backend/product.service.ts
damienwebdev: fixed 1deb33bfa23494d0d1196db09bf1265f3c84b1f9

libs/product/driver/in-memory/src/backend/product.service.ts
damienwebdev: 5e425da7d354b19f601beaca8ce4c6414e5f306d - nice catch.

libs/product/driver/in-memory/src/backend/product.service.ts
damienwebdev: I didn't notice this and mistakenly typed `makeStateKey` for `TRANSFER_STATE_KEY` as string.  c022e7cd7f277142238dfd1fb71367e8fc689744

PR #4171, opened 2025-11-05
fix(design): correct `DaffStatusableDirective` default status logic

### LINKED ISSUE #4170: User provided value to `status` is not working in `DaffStatusableDirective`
### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

User provided value to `status` is not working in `DaffStatusableDirective`

### Environment

- Daffodil version: 0.90.0
- Angular version:
- Browser(s):
- Operating system:

### Steps to reproduce

Steps to reproduce:
1. Serve `design-land`
2. Set a `defaultStatus` on the button component
3. See in the statusable-button example that examples with a different status from the `defaultStatus` is not respected.

### Expected behavior

User provided status should be respected.

### Actual behavior

2. The logic in `ngOnInit` is wrong, overriding user inputs instead of applying defaults only when needed

### Additional context

_No response_

xelaint:
## PR Checklist

## PR Type

- [x] Bug fix
- [ ] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #4170

## Breaking change?

- [ ] Yes
- [x] No

### DIFF
--- a/libs/design/src/core/statusable/statusable.directive.ts
+++ b/libs/design/src/core/statusable/statusable.directive.ts
@@ -122,7 +122,7 @@ export class DaffStatusableDirective implements DaffStatusable, OnChanges, OnIni
   ngOnInit() {
     validateStatus(this.status);
 
-    if (this.status !== this.defaultStatus && this.defaultStatus) {
+    if (this.defaultStatus && !this.status) {
       this.status = this.defaultStatus;
     }
   }
--- a/libs/design/src/core/statusable/specs/default.directive.spec.ts
+++ b/libs/design/src/core/statusable/specs/default.directive.spec.ts
@@ -1,17 +1,18 @@
-import { Component } from '@angular/core';
+import {
+  Component,
+  DebugElement,
+} from '@angular/core';
 import {
   waitForAsync,
   ComponentFixture,
   TestBed,
 } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
 
-import {
-  DaffStatus,
-  DaffStatusableDirective,
-} from '@daffodil/design';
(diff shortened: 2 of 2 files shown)


libs/design/src/core/statusable/statusable.directive.ts
```diff
@@ -122,7 +122,7 @@ export class DaffStatusableDirective implements DaffStatusable, OnChanges, OnIni
   ngOnInit() {
     validateStatus(this.status);
 
-    if (this.status !== this.defaultStatus && this.defaultStatus) {
+    if (this.defaultStatus && !this.status) {
```
damienwebdev, line 125: Can you add a test checking the status?

damienwebdev (changes requested):

libs/design/src/core/statusable/statusable.directive.ts
xelaint: Does the `default.directive.spec.ts` test not cover this?

damienwebdev: I would have expected it to, but clearly it must not have otherwise this PR wouldn't exist.

damienwebdev: I think you should write a test replicating the issue you saw BEFORE this change, and then prove that this change remedies that issue.

xelaint: Added

damienwebdev (approved):
lgtm

PR #4174, opened 2025-11-05
feat(design): add roving tab index

### LINKED ISSUE #1955: [FEAT] Roving tabIndex
# :bulb: Feature request

## Feature Name
Roving tabindex

## The Desired Behavior
Setting the tabindex of the focused element to "0" ensures that if the user tabs away from the widget and then returns, the selected item within the group retains focus. Note that updating the tabindex to "0" requires also updating the previously selected item to tabindex="-1". This technique involves programmatically moving focus in response to key events and updating the tabindex to reflect the currently focused item. To do this:

Bind a key down handler to each element in the group, and when an arrow key is used to move to another element:

1. programmatically apply focus to the new element,
2. update the tabindex of the focused element to "0", and
3. update the tabindex of the previously focused element to "-1".

Example: https://files.paciellogroup.com/training/WWW2012/samples/Samples/aria/tree/index.html

## Your Use Case
As a keyboard user, I would like to be able to tab out of a group of items without having to move focus through the entire group.

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

griest024:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #1955 
Part of: # 

## Breaking change?

- [ ] Yes
- [x] No


griest024:
@xelaint I need you to test out the behavior and see if you need any changes. For example, arrow keys on the outermost elements doesn't move focus, only inside a group, seems like we might always want arrow keys to move focus around.

xelaint:
> @xelaint I need you to test out the behavior and see if you need any changes. For example, arrow keys on the outermost elements doesn't move focus, only inside a group, seems like we might always want arrow keys to move focus around.

@griest024 Are you saying the arrow key should be able to move focus from `I get tabbed to first` to `I get tabbed to second`

libs/design/roving-tab-index/src/roving-tab-index.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+  Directive,
+  input,
+  signal,
+} from '@angular/core';
+
+import { DaffRovingTabIndexService } from './roving-tab-index-group.service';
+
+@Directive({
+  // eslint-disable-next-line @angular-eslint/directive-selector
+  selector: '[rti]',
```
xelaint: should probably be `[daffRTI]`?

griest024:
> > @xelaint I need you to test out the behavior and see if you need any changes. For example, arrow keys on the outermost elements doesn't move focus, only inside a group, seems like we might always want arrow keys to move focus around.
> 
> @griest024 Are you saying the arrow key should be able to move focus from `I get tabbed to first` to `I get tabbed to second`

yes

libs/design/roving-tab-index/src/roving-tab-index.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+    this.groupService.group() === this.rti()
+      ? 0
+      : -1,
+  );
+
+  constructor(
+    private groupService: DaffRovingTabIndexService,
+  ) {}
+
+  enterGroup(evt: Event) {
```
xelaint: `event` instead of `evt`? to be consistent with the codebase

xelaint (changes requested):

xelaint:
> > > @xelaint I need you to test out the behavior and see if you need any changes. For example, arrow keys on the outermost elements doesn't move focus, only inside a group, seems like we might always want arrow keys to move focus around.
> 
> > 
> 
> > @griest024 Are you saying the arrow key should be able to move focus from `I get tabbed to first` to `I get tabbed to second`
> 
> 
> 
> yes

I don't think so. Native behavior for moving around main interactive elements on a page without extra features is typically with tabs. e.g. moving from button to button

libs/design/roving-tab-index/src/roving-tab-index.directive.ts
griest024: no, `event` is an (albeit deprecated) global. It should be avoided as a var name. Its not that big a deal but `evt` is preferred

libs/design/roving-tab-index/src/roving-tab-index.directive.ts
griest024: possibly? I'm not sure core util things need namespacing but I'll defer to your preference here

xelaint: core util are prefixed. should move the directive to under `src/core` then.

griest024: I had it in a subpackage so that builds wouldn't take forever while prototyping. Do you think it should be in core then? I'll prefix it.

xelaint: Yeah, you can move to core. I think behavior is good. @damienwebdev Can you also review?

xelaint (changes requested):
Tabbing from this group `<div rti rtiBoundary="group">` should move focus to the next focusable element on the page, in this case it's the HTML tab, but tab cycles focus back to `<div rti>I get tabbed to first</div>`

libs/design/roving-tab-index/src/roving-tab-index-group.service.ts
```diff
@@ -0,0 +1,85 @@
...
+    if ('key' in evt) {
+      switch ((<KeyboardEvent>evt).key) {
+        case 'Tab':
+          evt.preventDefault();
+          if (this._group()) {
+            this.leave();
+          }
+          // eslint-disable-next-line no-constant-condition
+          if (true) {
+            const ary = Array.from(this.document.querySelectorAll<HTMLElement>(this._group() ? '[data-rti-boundary]' : '[data-rti=""]'));
```
damienwebdev: This looks rather non-performant as we'd be querying on every keystroke.

damienwebdev (changes requested):

damienwebdev:
So, this appears to have a few issues, once tabbed into a area with an RTI field, you're no longer able to tab to anything else labelled with RTI.

Additionally, if there are not RTI labelled elements (think `button`, `a` etc), inside the RTI, they aren't tabbable when I think they should be.

libs/design/roving-tab-index/src/roving-tab-index-group.service.ts
griest024: The alternative is that we rely on the browser's native handling of tabbing and instead modify the `tabindex` attribute in a way that replicates the behavior that we want, do you think we should do that?

griest024:
> So, this appears to have a few issues, once tabbed into a area with an RTI field, you're no longer able to tab to anything else labelled with RTI.

I thought this is what we wanted, tab moves focus outside of the group, arrow keys moves focus inside the group

> Additionally, if there are not RTI labelled elements (think `button`, `a` etc), inside the RTI, they aren't tabbable when I think they should be.

So we should retain browser native behavior of tabbing then?

damienwebdev:
image

griest024:
- tabbing should not escape group
- `a`s and `button`s need to be implicitly included in tabbing order
- do not intercept keyboard events, modify `tabindex` to control tabbing

griest024:
@damienwebdev I've made the requested changes, the implementation is fairly clean as well!

griest024:
focus trap inside rti boundary

griest024:
> focus trap inside rti boundary

implemented

xelaint:
lgtm! can be moved to core

libs/design/tsconfig.json
```diff
@@ -91,6 +91,9 @@
       "@daffodil/design/radio": [
         "libs/design/radio/src"
       ],
+      "@daffodil/design/roving-tab-index": [
```
xelaint: I don't think roving tab index should be its own subpackage? It's part of the core helper directives within `@daffodil/design`

apps/design-land/src/app/roving-tab-index/roving-tab-index.component.html
```diff
@@ -0,0 +1,3 @@
+<h1>Roving Tab Index</h1>
```
xelaint: remove since design-land no longer exists.

xelaint (changes requested):

libs/design/tsconfig.json
griest024: yeah its not, I just forgot to remove this part. good catch

libs/design-examples/roving-tab-index/src/examples.ts
```diff
@@ -0,0 +1,5 @@
+import { NestedGroupsRtiExampleComponent } from './rti-nested-groups/nested-groups.component';
```
xelaint: Update this to the new provider function implementation

xelaint (changes requested):

xelaint (commented):
should `/docs/design/components/core/roving-tab-index` be a valid URL in this PR?

griest024:
> should `/docs/design/components/core/roving-tab-index` be a valid URL in this PR?

no, that's blocked by #4330

libs/design/src/core/roving-tab-index/README.md
```diff
@@ -0,0 +1,19 @@
+# Roving Tab Index
+
+`@daffodil/design` provides an implementation of roving tab index that aims to address every common use case.
```
xelaint:
```suggestion
Roving tab index (RTI) manages keyboard focus across groups of interactive elements.
```

libs/design/src/core/roving-tab-index/README.md
```diff
@@ -0,0 +1,19 @@
...
+
```
xelaint, line 4: Add an `overview` section:

```
## Overview
It automatically handles `<a>` and `<button>` elements, supports arbitrarily nested groups with boundary-based scoping, and provides full keyboard navigation support.
```

libs/design/src/core/roving-tab-index/README.md
```diff
@@ -0,0 +1,19 @@
...
+## RTI Targets
+
+All `<a>` and `<button>` tags are automatically treated as RTI targets. Other elements can be an RTI target by using the `rti` directive. The targets can be navigated between using tab and shift+tab, as usual.
```
xelaint:
```suggestion
All `<a>` and `<button>` elements are automatically treated as RTI targets. Other elements can become targets by using the `rti` directive.
```

libs/design/src/core/roving-tab-index/README.md
```diff
@@ -0,0 +1,19 @@
+# Roving Tab Index
+
+`@daffodil/design` provides an implementation of roving tab index that aims to address every common use case.
+
+## RTI Targets
```
xelaint:
```suggestion
## Targets
```

libs/design/src/core/roving-tab-index/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+All `<a>` and `<button>` tags are automatically treated as RTI targets. Other elements can be an RTI target by using the `rti` directive. The targets can be navigated between using tab and shift+tab, as usual.
+
+## Groups
+
+RTI has the concept of *groups* delineated by boundaries. There is a default "root" group under which all RTI targets and other groups reside. This group is automatically managed and does not require a declared boundary. Specific groups can be created using the `rtiBoundary`. These groups can be arbitrarily deeply nested. A group boundary is automatically treated as an RTI target.
+
+Boundaries are given unique anonymous IDs unless a value is passed to `rtiBoundary`. Targets are considered in the group of the nearest ancestor that is an `rtiBoundary`. This can be overridden by passing a group ID to `rti` but this is only valid if the boundary has also been given a name.
+
+## Behavior
```
xelaint: replace w/

```
## Keyboard navigation

| Key | Action |
|---|---|
| <kbd>Tab</kbd> / <kbd>Shift+Tab</kbd> | Moves focus between RTI targets |
| <kbd>&uarr;</kbd> / <kbd>&darr;</kbd> | Moves focus between targets within a group |
| <kbd>Space</kbd> | Enters the focused group |
| <kbd>ESC</kbd> | Leaves the current group |

> Note: the parent group isn't necessarily the containing group in the DOM. Overriding boundary and target values can produce an RTI hierarchy that differs from the DOM tree.
```

libs/design/src/core/roving-tab-index/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Tab and shift+tab moves focus around between RTI targets in the current group. While having focus on a group, space enters that group. While inside a group, arrow down and up navigate the same as tab and shift+tab, respectively. Escape leaves the current group and enters the previous group. Note that the previous group isn't necessarily the same as the containing group in the DOM as overriding boundary and target values can generate an RTI group hierarchy that is different from the DOM.
+
+<daffio-example-viewer example="rti-nested-groups"></daffio-example-viewer>
```
xelaint: move up to end of `## Groups`

libs/design/src/core/roving-tab-index/roving-tab-index-boundary.directive.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+  signal,
+  WritableSignal,
+} from '@angular/core';
+import {
+  ComponentFixture,
+  TestBed,
+} from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+
+import { DaffRovingTabIndexBoundaryDirective } from './roving-tab-index-boundary.directive';
```
xelaint, line 13: should be imported from `@daffodil/design`

libs/design/src/core/roving-tab-index/roving-tab-index-boundary.directive.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+import { DaffRovingTabIndexService } from './roving-tab-index-group.service';
```
xelaint: should be imported from `@daffodil/design`

libs/design/src/core/roving-tab-index/roving-tab-index-boundary.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+   */
+  private static _uniqueIdCounter = 0;
+
+  /**
+   * Don't touch this directly. Use `_uniqueId`.
+   */
+  private _cachedUniqueId: string | undefined;
+  private get _uniqueId(): string {
+    if (!this._cachedUniqueId) {
+      this._cachedUniqueId = `ε-rtiBoundary-${DaffRovingTabIndexBoundaryDirective._uniqueIdCounter++}`;
```
xelaint, line 40: Why use `ε-`?

libs/design/src/core/roving-tab-index/roving-tab-index.directive.spec.ts
```diff
@@ -0,0 +1,272 @@
...
+} from '@angular/core';
+import {
+  ComponentFixture,
+  TestBed,
+} from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+
+import { DaffRovingTabIndexBoundaryDirective } from './roving-tab-index-boundary.directive';
+import { DaffRovingTabIndexService } from './roving-tab-index-group.service';
+import { DaffRovingTabIndexDirective } from './roving-tab-index.directive';
```
xelaint: import from `@daffodil/design`

libs/design/src/core/roving-tab-index/roving-tab-index.directive.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  next(evt: Event) {
+    evt.stopPropagation();
+    this.service.next();
+  }
+
+  previous(evt: Event) {
+    evt.stopPropagation();
+    this.service.previous();
+  }
```
xelaint, line 75: do these need to be marked as `@docs-private`?

libs/design/src/core/roving-tab-index/roving-tab-index-boundary.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+  Directive,
+  effect,
+  forwardRef,
+  input,
+} from '@angular/core';
+
+import { DaffRovingTabIndexService } from './roving-tab-index-group.service';
+import { DaffRovingTabIndexDirective } from './roving-tab-index.directive';
+
+@Directive({
```
xelaint, line 16: Docs needed here?

libs/design/src/core/roving-tab-index/roving-tab-index.directive.ts
```diff
@@ -0,0 +1,56 @@
...
+  Directive,
+  input,
+  Optional,
+  SkipSelf,
+} from '@angular/core';
+
+import { DaffRovingTabIndexBoundaryDirective } from './roving-tab-index-boundary.directive';
+import { DaffRovingTabIndexService } from './roving-tab-index-group.service';
+
+@Directive({
```
xelaint, line 16: Docs needed here?

libs/design/src/core/roving-tab-index/roving-tab-index-boundary.directive.ts
```diff
@@ -0,0 +1,59 @@
...
+  constructor(
+    private groupService: DaffRovingTabIndexService,
+    private focusTrap: CdkTrapFocus,
+  ) {
+    effect(() => {
+      this.focusTrap.enabled = this.effectiveBoundary() === this.groupService.group();
+    });
+  }
+
+  enterGroup(evt: Event) {
```
xelaint, line 67: does this need to be marked as `@docs-private?`?

xelaint (changes requested):

griest024:
@xelaint I will improve the docs. In the meantime, you can use https://github.com/graycoreio/daffodil/pull/4387 to preview the docs in daffio and give feedback on the docsgen for core design stuff

libs/design/src/core/roving-tab-index/roving-tab-index-boundary.directive.ts
griest024: hm probably so

libs/design/src/core/roving-tab-index/roving-tab-index.directive.spec.ts
griest024: no the SUT should be imported relative, the others should be though, good catch

libs/design/src/core/roving-tab-index/roving-tab-index.directive.ts
griest024: probably so

libs/design/src/core/roving-tab-index/public_api.ts
```diff
@@ -0,0 +1,3 @@
+export { DaffRovingTabIndexService } from './roving-tab-index-group.service';
```
xelaint: Does the service need to be part of the public api?

libs/design/src/core/roving-tab-index/roving-tab-index-boundary.directive.ts
griest024: to ensure that we don't collide with any IDs that users happen to give their custom groups. Perhaps overkill but better safe than sorry

libs/design/src/core/roving-tab-index/public_api.ts
griest024: I guess it doesn't *need* to be but it could be useful. I'll remove it if you prefer.

xelaint: Can you add an example of when the service would be used? And everything in service will need to be documented or marked as `@docs-private` if it's going to be part of the public api.

griest024: I just removed it

xelaint (approved):

PR #4176, opened 2025-11-07
feat(docs): add copy code snippet button to code blocks

### LINKED ISSUE #3949: Add copy code snippet button to code blocks
### Feature description

Add copy code snippet button to code blocks.

- [ ] Make DaffDocsCopyButtonComponent
- [ ] Modify the effect in the [guides viewer](https://github.com/graycoreio/daffodil/blob/develop/apps/daffio/src/app/docs/guides/components/guides-content/guides-content.component.ts#L35) to append this button (and attach any associated event handlers).

### Use case

As a developer, I would like to be able to easily copy a code block section.

### Prior work

_No response_

### Daffodil version

0.88.2

### Additional context

_No response_

joannalauu:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #3949 

## New behavior
Adds a copy button to the code blocks in the daffio guides. Code block is copied to clipboard on click.

## Breaking change?

- [ ] Yes
- [x] No

## Additional context

https://github.com/user-attachments/assets/efdf593d-4783-4008-855b-7c19c1f3353e


joannalauu:
Please let me know if I should change the code structure in the docs package – I wasn't sure how to organize the files because there wasn't an existing structure to base off of. Also let me know if I should move the addCopyButtonsToCodeBlocks and cleanupButtons functions outside of guides-content.component.ts and then just import them in.

griest024 (dismissed):
So ideally we want to add the HTML content in dgeni as I described here: https://github.com/graycoreio/daffodil/issues/3949#issuecomment-3308764140. The way you've done it in daffio here is not [hydration safe](https://angular.dev/guide/hydration#direct-dom-manipulation)

The tricky part of that is that you won't be able to use Angular components. So unless we want to use custom elements at this time (@damienwebdev we don't, right?) you'll need to convert `DaffDocsCopyButtonComponent` to pure HTML/JS/CSS and add it to the content in dgeni.

EDIT: @sunray4 actually standby, it might be hydration safe because of `afterRenderEffect`

griest024:
@sunray4 thanks for your contribution! I hope your exams went well!

apps/daffio/src/app/docs/guides/components/guides-content/guides-content.component.ts
```diff
@@ -29,12 +36,74 @@ export class DaffioDocsGuidesContentComponent implements DaffioDocsDynamicConten
...
-      onCleanup(() => this.tocRegistry.set([]));
+      onCleanup(() => {
+        this.tocRegistry.set([]);
+      });
+    });
+
+    afterRenderEffect(() => {
+      this.doc();
+      this.addCopyButtonsToCodeBlocks();
+    });
```
griest024: Prefer explicitly specifying the phase here. Can you get it working without using `mixedReadWrite`?

> You should prefer using the read and write phases over the `earlyRead` and `mixedReadWrite` phases when possible, to avoid performance degradation.

https://angular.dev/api/core/afterRenderEffect

apps/daffio/src/app/docs/guides/components/guides-content/guides-content.component.ts
```diff
@@ -29,12 +36,74 @@ export class DaffioDocsGuidesContentComponent implements DaffioDocsDynamicConten
...
+
+      const innerButton = buttonElement.querySelector('button');
+      if (innerButton) {
+        innerButton.classList.add('daff-copy-code-block');
+      }
+
+      // Add copy button to the code block
+      pre.appendChild(buttonElement);
+
+      this.buttonRefs.push(buttonRef);
```
griest024: We'll need to call this in every doc kind component so this should be extracted out so that it is reusable.

griest024 (changes requested):

apps/daffio/src/app/docs/guides/components/guides-content/guides-content.component.ts
joannalauu: Sounds good, in this case I'll just put the addCopyButtonsToCodeBlocks and cleanup functionality into a service inside the doc package – may I confirm if this works?

griest024: yes but be careful about the injection hierarchy of the service. It must inherit the injector of the component to get the correct `elementRef`, etc.

apps/daffio/src/app/docs/guides/components/guides-content/guides-content.component.ts
```diff
@@ -29,12 +36,30 @@ export class DaffioDocsGuidesContentComponent implements DaffioDocsDynamicConten
...
     effect((onCleanup) => {
       this.tocRegistry.set(this.doc().tableOfContents);
-      onCleanup(() => this.tocRegistry.set([]));
+      onCleanup(() => {
+        this.tocRegistry.set([]);
+      });
+    });
+
+    afterRenderEffect({
+      write: () => {
```
griest024: This callback should accept a cleanup arg. I forget the exact API but you shouldn't need to touch `DestroyRef` directly

griest024 (changes requested):

apps/daffio/src/app/docs/guides/components/guides-content/guides-content.component.ts
```diff
@@ -29,12 +35,29 @@ export class DaffioDocsGuidesContentComponent implements DaffioDocsDynamicConten
...
-      onCleanup(() => this.tocRegistry.set([]));
+      onCleanup(() => {
+        this.tocRegistry.set([]);
+      });
+    });
+
+    afterRenderEffect({
+      write: (onCleanup) => {
+        this.doc();
+        this.copyButtonService.addCopyButtonsToCodeBlocks(this.elementRef.nativeElement, this.viewContainerRef);
```
griest024: You can directly inject the args here, the constructor is a valid injection context.

griest024 (dismissed):
Nice code! :clap: you need to add this to all the other doc kinds now: API, design, packages, etc. Take care that the copy button is not added to the source code block (which all the API pages have)

libs/docs/src/copy-button/service/copy-code-block.service.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    codeBlocks.forEach((pre: HTMLPreElement) => {
+      const code = pre.querySelector('code');
+      if (!code) {
+        return;
+      }
+
+      const textContent = code.textContent || '';
+
+      pre.style.position = 'relative';
```
xelaint: We should just update the `pre` style in `DaffArticleComponent` to include this style.

libs/docs/src/copy-button/service/copy-code-block.service.ts
```diff
@@ -0,0 +1,61 @@
...
+      // Create the copy button component
+      const buttonRef = viewContainerRef.createComponent(DaffDocsCopyButtonComponent);
+
+      // Pass the code snippet to the button
+      buttonRef.setInput('content', textContent);
+
+      const buttonElement = buttonRef.location.nativeElement;
+
+      const innerButton = buttonElement.querySelector('button');
+      if (innerButton) {
```
xelaint: I don't think adding a class to innerButton on copy is necessary. The positioning can just be applied to the host component.

libs/docs/src/copy-button/copy-button/copy-button.component.scss
```diff
@@ -0,0 +1,20 @@
+.daff-docs-copy-button {
```
xelaint: This style file should just be:

```
:host {
	display: block;
	position: absolute;
	top: 0.5rem;
	right: 0.5rem;
	z-index: 1;
}
```

state styles (hover and focus) should be handled by the button component.

libs/docs/src/copy-button/copy-button/copy-button.component.html
```diff
@@ -0,0 +1,13 @@
+<button
+  daff-icon-button
+  size="sm"
+  type="button"
+  class="daff-docs-copy-button"
```
xelaint: class can be removed

libs/docs/src/copy-button/copy-button/copy-button.component.html
```diff
@@ -0,0 +1,13 @@
...
+  [attr.aria-label]="copied() ? 'Copied to clipboard' : 'Copy to clipboard'"
```
xelaint: move logic to component

libs/docs/src/copy-button/service/copy-code-block.service.ts
```diff
@@ -0,0 +1,61 @@
+import {
+  Injectable,
+  ComponentRef,
+  ViewContainerRef,
+} from '@angular/core';
+
+import { DaffDocsCopyButtonComponent } from '../copy-button/copy-button.component';
+
+@Injectable()
+export class CodeBlockCopyButtonService {
```
xelaint: Prefix this service with `DaffDocs`

xelaint (changes requested):

libs/docs/src/copy-button/copy-button/copy-button.component.ts
```diff
@@ -0,0 +1,66 @@
...
+ * ```html
+ * <daff-docs-copy-button [content]="codeSnippet" />
+ * ```
+ */
+@Component({
+  selector: 'daff-docs-copy-button',
+  templateUrl: './copy-button.component.html',
+  styleUrls: ['./copy-button.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  standalone: true,
```
xelaint: can remove `standalone: true;` since it's the default

xelaint (changes requested):

libs/docs/src/copy-button/copy-button/copy-button.component.ts
```diff
@@ -0,0 +1,68 @@
...
+ * <daff-docs-copy-button [content]="codeSnippet" />
+ * ```
+ */
+@Component({
+  selector: 'daff-docs-copy-button',
+  templateUrl: './copy-button.component.html',
+  styleUrls: ['./copy-button.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  host: {
+    '[attr.aria-label]': 'copied() ? "Copied to clipboard" : "Copy to clipboard"',
```
xelaint: My requested change was to move the `copied() ? "Copied to clipboard" : "Copy to clipboard"` logic into the component as a function (e.g. DesignLandSidebarViewportComponent), but keep the `aria-label` attribute on the button itself

xelaint (changes requested):
1. A change you made moved the `<daff-docs-copy-button>` from within `<pre>` to outside of `<pre>`. The button should remain inside of `<pre>`, so you don't need to create an extra div to set a position relative style on it.
2. Design API tabs don't have the copy button

joannalauu:
> 1. A change you made moved the `<daff-docs-copy-button>` from within `<pre>` to outside of `<pre>`. The button should remain inside of `<pre>`, so you don't need to create an extra div to set a position relative style on it.

While testing the button, I realized that when the pre element has horizontal scroll, the button moves with scroll. My workaround was to wrap the pre and then put the button in the wrapper – please let me know if there's a better way to do this!!

joannalauu:
Tests in daffio package component-content.component.spec.ts are failing because of a missing property that is required in DaffPackageGuideDoc. When I checked out to the most recent commit in the upstream develop branch, the test still fails. May I ask how do I resolve this?

xelaint:
> > 1. A change you made moved the `<daff-docs-copy-button>` from within `<pre>` to outside of `<pre>`. The button should remain inside of `<pre>`, so you don't need to create an extra div to set a position relative style on it.
> 
> While testing the button, I realized that when the pre element has horizontal scroll, the button moves with scroll. My workaround was to wrap the pre and then put the button in the wrapper – please let me know if there's a better way to do this!!

Ah, good catch. What you did with the div is fine then, no further changes needed.

griest024:
> Tests in daffio package component-content.component.spec.ts are failing because of a missing property that is required in DaffPackageGuideDoc. When I checked out to the most recent commit in the upstream develop branch, the test still fails. May I ask how do I resolve this? 

It's my fault, rebase on develop and you should be good.

libs/docs/package.json
```diff
@@ -29,10 +29,14 @@
     "@angular/common": "0.0.0-PLACEHOLDER",
     "@angular/core": "0.0.0-PLACEHOLDER",
     "@daffodil/core": "0.0.0-PLACEHOLDER",
-    "@daffodil/docs-utils": "0.0.0-PLACEHOLDER"
+    "@daffodil/design": "0.0.0-PLACEHOLDER",
```
damienwebdev, line 32: I was just looking through this PR as I had a moment, I'm just wondering if we couldn't move the `copy` functionality fully into `DaffArticleComponent`. @griest024 @xelaint @sunray4

joannalauu: @damienwebdev Since all code blocks in the daffio should have a copy button, it would make sense for the `copy` functionality to be part of DaffArticleComponent. If there might be use cases of the pre element in DaffArticleComponent where we don't want the element to have the copy button, I can implement a check for a `nocopy` attribute in the pre element which would prevent the copy button from being added. Let me know if that works!

xelaint: I think the option to have a no copy button is good.

joannalauu: Sounds good, I'll start working on implementing this!!

joannalauu:
Hey, I've just finished moving the `copy` functionality into the article component. However, there was one error I couldn't resolve: Once I moved the addCopyButton service into `DaffArticleComponent`, the button page on daffio returns 404. I have attached a clip showing the error. This error didn't exist in my previous implementation, where the addCopyButton service was run in each section of daffio respectively. Please let me know how I can resolve this!

https://github.com/user-attachments/assets/8753243c-87d4-4b55-b1a8-cd053ee258fe

joannalauu:
> Hey, I've just finished moving the `copy` functionality into the article component. However, there was one error I couldn't resolve: Once I moved the addCopyButton service into `DaffArticleComponent`, the button page on daffio returns 404. I have attached a clip showing the error. This error didn't exist in my previous implementation, where the addCopyButton service was run in each section of daffio respectively. Please let me know how I can resolve this!

UPDATE: I fixed the error by implementing the button styles from scratch instead of using DaffIconButtonComponent

joannalauu:
@damienwebdev I've moved the `copy` functionality into `DaffArticleComponent`. May I ask if this implementation is what you were looking for?

Some key changes:
- All code blocks in `DaffArticleComponent` now has a copy button by default. This copy button can be removed by adding a `nocopy` attribute into the `pre` element.
- `DaffDocsCopyButtonComponent` is moved into the `libs/design/article/src` folder and renamed to `DaffArticleCopyButtonComponent`
- `MutationObserver` is used in `DaffArticleComponent` to check for new `pre` elements that may be created within an existing `daff-article`. This may cause unnecessary overhead if a user implements `daff-article` but doesn't include code blocks
- `DaffArticleCopyButtonComponent` is exported as part of `DAFF_ARTICLE_COMPONENTS` from `@daffodil/design/article`
- `DaffArticleCopyButtonComponent` theming is implemented from scratch because importing `DaffIconComponent` into `DaffArticleCopyButtonComponent` breaks www.daff.io/docs/design/components/button


https://github.com/user-attachments/assets/97dce9f6-3f70-49c5-9bbc-416d897ca196

joannalauu:
@griest024 Hey I've made all requested changes, may I ask if there is anything else you'd like me to fix/change?

griest024:
> @griest024 Hey I've made all requested changes, may I ask if there is anything else you'd like me to fix/change?

sorry my review is taking so long. Nag me if I haven't reviewed this by day after tomorrow haha

libs/docs/src/remove-me.spec.ts
```diff
@@ -1,3 +0,0 @@
-describe('@daffodil/docs | remove me', () => {
-  it('should be removed', () => {});
```
damienwebdev, line 2: Remove this diff from this PR.

libs/design/article/src/public_api.ts
```diff
@@ -2,4 +2,5 @@ export { DaffArticleModule } from './article.module';
 export { DAFF_ARTICLE_COMPONENTS } from './article';
 
 export { DaffArticleComponent } from './article/article.component';
+export { DaffArticleCopyButtonComponent } from './article-copy-button/copy-button/copy-button.component';
```
damienwebdev: This should not be in the public api.

libs/design/article/src/article.ts
```diff
@@ -1,4 +1,5 @@
 import { DaffArticleComponent } from './article/article.component';
+import { DaffArticleCopyButtonComponent } from './article-copy-button/copy-button/copy-button.component';
```
damienwebdev: This should not be in the public api.

libs/design/article/src/article.module.ts
```diff
@@ -11,10 +12,12 @@ import { DaffArticleMetaDirective } from './article-meta/article-meta.directive'
   imports: [
     CommonModule,
     DaffArticleComponent,
+    DaffArticleCopyButtonComponent,
```
damienwebdev: This module is deprecated and doesn't need this change.

libs/design/article/src/article/article.component.ts
```diff
@@ -18,5 +25,56 @@ import {
...
+    const viewContainerRef = inject(ViewContainerRef);
+    const document = inject(DOCUMENT);
+
+    let observer: MutationObserver | null = null;
+
+    if (
+      document?.defaultView &&
+      typeof document.defaultView.MutationObserver !== 'undefined'
+    ) {
+      observer = new document.defaultView.MutationObserver(mutations => {
```
damienwebdev: This is very confusing. I don't think MutationObserver is the right mechanism for this. I would prefer to use contentChildren signals.

libs/design/article/src/article-copy-button/copy-button/copy-button.component.ts
```diff
@@ -0,0 +1,66 @@
...
+    }
+  }
+
+  async copyToClipboard(): Promise<void> {
+    try {
+      // Write content to clipboard
+      await navigator.clipboard.writeText(this.content());
+      this.copied.set(true);
+
+      this.timeoutId = setTimeout(() => {
```
damienwebdev: This timeout should be run outside the ngZone.

apps/daffio/src/app/docs/guides/components/guides-content/guides-content.component.ts
```diff
@@ -34,7 +34,9 @@ export class DaffioDocsGuidesContentComponent implements DaffioDocsDynamicConten
   ) {
     effect((onCleanup) => {
       this.tocRegistry.set(this.doc().tableOfContents);
-      onCleanup(() => this.tocRegistry.set([]));
```
damienwebdev, line 37: Remove this diff.

apps/daffio/src/app/docs/design/components/component-content/component-content.component.ts
```diff
@@ -36,6 +36,7 @@ import { DaffioDocsTocHeaderDirective } from '../../../toc/header.directive';
 import { DaffioDocsTocService } from '../../../toc/toc.service';
 import { DaffioDocsDesignApiSortSectionLabels } from '../../pipes/sort-api-section-labels.pipe';
 
+
```
damienwebdev: Remove this diff.

damienwebdev (changes requested):
I'm going to take a stab at this by basing off your branch. There are aspects of this implementation that I find cumbersome and hard to understand.

libs/design/article/src/article/article.component.ts
joannalauu: Sounds good. I initially used MutationObserver instead of contentChildren because it is hard to track elements injected via `[innerHTML]`, but I'll work on fixing this!!

joannalauu: Hey, I've looked into using contentChildren signals to track content inside `daff-article`, and I don't think this is possible because contentChildren can only track components within the Angular component tree, but the content inside `daff-article` is not part of the Angular component tree since it is rendered through `[innerHTML]`. However, please let me know if I am thinking about this issue incorrectly!!

If the use of MutationObserver isn't ideal, we could revert back to commit [`0ae8bb0`](https://github.com/graycoreio/daffodil/pull/4176/files/0ae8bb0e8330a9909e527ddc752802a9e780a63f), where the copy button is implemented in `docs` package and then imported into daffio.

damienwebdev: That's a very good point that I hadn't considered. I haven't had a chance to dig further into this yet, but that's a very good catch.

damienwebdev:
@sunray4 I've added a diff to your branch, please take a look:

In short:

- adjust copy-button to use daff-icon-button
- add specs for daffio scenario with innerhtml
- rework article to use afterNextRender over MutationObserver
- Adjust article scss to use position:relative on pre and overflow scroll on code
- add some specs for good measure
- added a timeout clear so that clicking "again" before the timeout doesn't cause strange UI behavior

Notably, I was able to remove the mutation observer using `afterNextRender` in the `write` phase (along with a small check to see if a `pre` already has an associated button`).

I disliked the `wrapper` component as its unnecessary DOM manipulation that can be avoided.

The specs I added specifically cover the scenario in daff.io, so those should eliminate any confusion about the specific scenario.

Some small comments for @xelaint 

 - This PR adjusts the styles from the `pre` element to the `code` element, I'd like your thoughts on this. 
 - I changed the `daff-icon-button` color to `primary`. This probably isn't ideal, however, on small screens the icon button has no background, so its blends into any text underneath it when the codeblock is scrolled.

As per an earlier comment @sunray4 :

> DaffArticleCopyButtonComponent theming is implemented from scratch because importing DaffIconComponent into DaffArticleCopyButtonComponent breaks www.daff.io/docs/design/components/button

I didn't see this anymore. Do you still?

joannalauu:
@damienwebdev This looks great, thank you for helping with this PR!!

> > DaffArticleCopyButtonComponent theming is implemented from scratch because importing DaffIconComponent into DaffArticleCopyButtonComponent breaks [www.daff.io/docs/design/components/button](http://www.daff.io/docs/design/components/button)
> 
> I didn't see this anymore. Do you still?

When I serve daffio on my device, [www.daff.io/docs/design/components/button](http://www.daff.io/docs/design/components/button) currently returns 404. I've ran `npm ci` and rebuilt all packages before serving. Running the website in incognito also doesn't work. If the page works on your device, then it's likely a problem on my end, but perhaps double check by running on a third device just to be sure?

libs/design/article/src/article/article.component.scss
```diff
@@ -100,18 +100,23 @@
...
+		position: relative;
 
 		&:last-child {
 			margin-bottom: 0;
 		}
 
 		code {
-			padding: 0;
+			white-space: pre;
+			-webkit-overflow-scrolling: touch;
```
xelaint: Remove. It's an invalid property and no longer needed as of safari 13 (https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes)

libs/design/article/src/article/article.component.scss
```diff
@@ -100,18 +100,23 @@
...
+			overflow-x: auto;
+			display: block;
+			max-width: 100%;
+			min-width: 100px;
```
xelaint: Don't think these styles are needed.

libs/design/article/src/article/article.component.scss
```diff
@@ -100,18 +100,23 @@
 
 	pre {
 		border-radius: 0.25rem;
-		font-size: t.$font-size-sm;
-		line-height: 1.5rem;
-		margin: 1.5rem 0;
-		padding: 1.5rem;
-		overflow-x: auto;
+		overflow: hidden;
```
xelaint: `overflow: hidden;` is not needed

libs/design/article/src/article-theme.scss
```diff
@@ -36,6 +36,14 @@
 				}
 			}
 
+			&__copy-button {
```
xelaint, line 39: for light:
```
background: daff-color($neutral, 10);
border: 1px solid daff-color($neutral, 20);
```

for dark:
```
background: daff-color($neutral, 90);
border: 1px solid daff-color($neutral, 80);
```

libs/design/article/src/article-copy-button/copy-button/copy-button.component.html
```diff
@@ -0,0 +1,13 @@
+<button
+  daff-icon-button
+  class="daff-article__copy-button"
+  type="button"
+  color="primary"
```
xelaint: Remove `color="primary"`

libs/design/article/src/article-copy-button/copy-button/copy-button.component.scss
```diff
@@ -0,0 +1,9 @@
+@use '../../../../scss/interactions';
```
xelaint: Remove. This is not used.

libs/design/article/src/article-copy-button/copy-button/copy-button.component.scss
```diff
@@ -0,0 +1,9 @@
...
+
+:host {
+	display: block;
+	position: absolute;
+	top: 0.2rem;
+	right: 0.2rem;
```
xelaint:
```
top: 0.25rem;
right : 0.25rem;
```

libs/design/article/src/article/article.component.scss
```diff
@@ -100,18 +100,23 @@
...
+		position: relative;
 
 		&:last-child {
 			margin-bottom: 0;
 		}
 
 		code {
-			padding: 0;
+			white-space: pre;
```
xelaint: Can you reorganize this code with display styles first, with the rest listed alphabetically?

xelaint (changes requested):

xelaint (dismissed):

damienwebdev (dismissed):

damienwebdev (approved):
lgtm

PR #4209, opened 2025-11-24
feat(content): introduce schema types

damienwebdev:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior
Currently, we have no way to render content from ecommerce to Daffodil storefronts beyond simple HTML.

## New behavior
This defines an AST so that we can bring content from these platforms in a structured format without requiring custom elements.

## Breaking change?

- [ ] Yes
- [x] No

### DIFF
--- a/libs/content/guides/schema.md
+++ b/libs/content/guides/schema.md
@@ -30,7 +30,7 @@ interface DaffTextSchema {
 ```typescript
 const textSchema: DaffTextSchema = {
   type: 'textSchema',
-  text: 'This is a paragraph of text.'
+  text: 'This is some text.'
 };
 ```
 
--- a/libs/content/guides/schema.md
+++ b/libs/content/guides/schema.md
@@ -1,0 +1,259 @@
+# Content Schema
+
+The schema system provides a flexible, type-safe way to define content structure in `@daffodil/content`. Schemas are hierarchical definitions that can represent elements, components, and text, allowing you to build complex page layouts programmatically. 
+
+Simply stated, this schema corresponds to a JSON-structure representing HTML.
+
+## Overview
+
+A content schema is a tree structure defined by the type `DaffContentSchema` where each node can be one of three types:
+
+- [**Text Schema**](#text-schema): Represents text nodes
+- [**Element Schema**](#element-schema): Represents HTML elements with attributes and styles
+- [**Component Schema**](#component-schema): Represents Angular components with inputs
+
+## Text Schema
+
+Text schemas represent simple text nodes.
+
+### Interface
+
+```typescript
+interface DaffTextSchema {
+  type: 'textSchema';
+  text: string;
+}
+```
+
+### Example
+
+```typescript
+const textSchema: DaffTextSchema = {
+  type: 'textSchema',
+  text: 'This is a paragraph of text.'
+};
+```
+
+## Element Schema
+
+`DaffContentElementSchema` define HTML elements with optional attributes, styles, and children.
+
+### Example: Basic Element
+
+```typescript
+const divSchema: DaffContentElementSchema = {
+  type: 'elementSchema',
+  element: 'div',
+  attributes: {
+    class: 'container',
+    id: 'main-content'
+  },
+  children: [
+    {
+      type: 'textSchema',
+      text: 'Hello World'
+    }
+  ]
+};
+```
+
+### Example: Responsive Styles
+
+Element schemas support responsive styling through breakpoint definitions:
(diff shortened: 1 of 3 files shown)


libs/content/guides/schema.md
```diff
@@ -0,0 +1,227 @@
...
+  text: string;
+}
+```
+
+### Example
+
+```typescript
+const textSchema: DaffTextSchema = {
+  type: 'textSchema',
+  text: 'This is a paragraph of text.'
```
griest024: How is it actually rendered in dom? As a p?

libs/content/guides/schema.md
```diff
@@ -0,0 +1,227 @@
...
+};
+```
+
+## Element Schema
+
+`DaffContentElementSchema` define HTML elements with optional attributes, styles, and children.
+
+### Example: Basic Element
```
griest024, line 32: I think all of these sections should show the interface, if any of them do

libs/content/guides/schema.md
```diff
@@ -0,0 +1,227 @@
...
+
+`DaffContentComponentSchema` allow you to render Angular components dynamically with specified inputs.
+
+### Example: Component with Inputs
+
+```typescript
+const productCardSchema: DaffContentComponentSchema = {
+  type: 'componentSchema',
+  name: 'ProductCard',
+  inputs: {
```
griest024, line 102: Can I bind dynamic values?

libs/content/guides/schema.md
```diff
@@ -0,0 +1,227 @@
...
+
+```typescript
+const cardSchema: DaffContentComponentSchema = {
+  type: 'componentSchema',
+  name: 'Card',
+  inputs: {
+    title: 'Featured Product',
+    variant: 'elevated'
+  },
+  children: [
```
griest024, line 120: Does this support all the content projection features available when authoring traditional angular templates?

griest024 (changes requested):

libs/content/guides/schema.md
damienwebdev: There's a small discrepancy on this point between the "editable-renderer" (which admins use to modify content) and the "renderer".

- `editable-renderer` renders `<span contenteditable>This is a paragraph of text.</span>`
- `schema-renderer` renders `This is a paragraph of text.` inside the renderer root

libs/content/guides/schema.md
damienwebdev: What do you mean? I want to leave that to the API docs so that I don't have to maintain this content.

libs/content/guides/schema.md
damienwebdev: No. Schema is purely static. You could re-compute schema dynamically if you'd like.

libs/content/guides/schema.md
damienwebdev: It doesn't currently. That's a feature target that I want to work towards. Currently everything is transcluded into the main `ng-content` of the `componentSchema`.

That said, I don't think the schema would be the one defining which slot your element is targeting. Depending on the attributes (just like with regular Angular templates) certain slots would be targeted.

libs/content/guides/schema.md
griest024: I think its misleading to word it like that then, reading this I would assume that it gets rendered as `<p>`.  Just change the content of the text to not reference a paragraph

libs/content/guides/schema.md
griest024: Seems good, what is your motivation for showing the text schema interface?

libs/content/guides/schema.md
griest024: Its probably worth adding a note here to that effect.

libs/content/guides/schema.md
griest024: I was mainly thinking of `exportAs`, but yeah, we'll want this full feature support eventually

libs/content/guides/schema.md
damienwebdev: I removed it.

libs/content/guides/schema.md
damienwebdev: Added a note.

griest024 (approved):

PR #4224, opened 2025-11-25
feat(driver): add magento driver features

griest024:
## PR Checklist

## PR Type

- [ ] Bug fix
- [ ] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: # 
Part of: # 

## Breaking change?

- [ ] Yes
- [ ] No

### DIFF
--- a/libs/driver/magento/src/features/cacheable-operations.ts
+++ b/libs/driver/magento/src/features/cacheable-operations.ts
@@ -1,3 +1,5 @@
+import { makeEnvironmentProviders } from '@angular/core';
+
 import { MagentoDriverFeatureKind } from './kind.enum';
 import { makeMagentoDriverFeature } from './make-feature';
 import { MagentoDriverFeature } from './type';
@@ -11,5 +13,5 @@ import {
  */
 export const withDaffMagentoCacheableOperations = (...extraOperations: Array<string>): MagentoDriverFeature => makeMagentoDriverFeature(MagentoDriverFeatureKind.CacheableOperations, [
   provideDaffMagentoApolloCacheableOperations(),
-  ...provideManyDaffMagentoCacheableOperations(...extraOperations),
+  makeEnvironmentProviders(provideManyDaffMagentoCacheableOperations(...extraOperations)),
 ]);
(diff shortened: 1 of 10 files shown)


libs/driver/magento/src/features/public_api.ts
```diff
@@ -0,0 +1,22 @@
...
+export { withDaffMagentoCacheHeader } from './cache-header';
+export { withDaffMagentoCacheableOperations } from './cacheable-operations';
+export {
+  DAFF_DRIVER_MAGENTO_ERROR_HANDLER,
+  withDaffDriverMagentoErrorHandler,
+  provideDaffDriverMagentoErrorHandler,
+} from './error-handler';
+export { MagentoDriverFeatureKind } from './kind.enum';
+export { makeMagentoDriverFeature } from './make-feature';
+export {
```
damienwebdev, line 16: If we're going to be moving forward with this `withFeature` model, do we need to export the providers into the public api as well?

libs/driver/magento/src/features/cacheable-operations.ts
```diff
@@ -0,0 +1,15 @@
...
+  provideDaffMagentoApolloCacheableOperations,
+  provideManyDaffMagentoCacheableOperations,
+} from '../graphql/public_api';
+
+/**
+ * A {@link provideMagentoDriver} feature that sets up {@link provideDaffMagentoApolloCacheableOperations}.
+ */
+export const withDaffMagentoCacheableOperations = (...extraOperations: Array<string>): MagentoDriverFeature => makeMagentoDriverFeature(MagentoDriverFeatureKind.CacheableOperations, [
+  provideDaffMagentoApolloCacheableOperations(),
+  ...provideManyDaffMagentoCacheableOperations(...extraOperations),
```
damienwebdev: I don't think we need this spread.

damienwebdev (changes requested):

libs/driver/magento/src/features/public_api.ts
griest024: they are useful for app devs that aren't using `provideMagentoDriver`. If we absolutely won't support that scenario, then no, but I prefer to support any desired level of customization.

libs/driver/magento/src/features/cacheable-operations.ts
griest024: The only part of the type signature that supports the unspread provider is `any[]` which I don't want to rely on. It very well might work in runtime but that feels brittle.

damienwebdev (approved):
lgtm

PR #3569, opened 2025-03-19
feat(design)!: update breadcrumb component to support truncation

xelaint:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Other... Please describe:
```

## What is the current behavior?

Fixes: #3513 

## Does this PR introduce a breaking change?
```
[ ] Yes
[ ] No
```

## Other information

PR #3998, opened 2025-09-22
feat(design): add keyboard support and basic aria roles for menu

xelaint:
## PR Checklist

- [x] Commit message follows our [contributing guidelines](https://github.com/graycoreio/daffodil/blob/develop/CONTRIBUTING.md#commit)
- [x] Tests added/updated (for bug fixes/features)
- [x] Documentation added/updated (for bug fixes/features)

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Part of: #2841 

## New behavior
- `aria-haspopup` and `aria-expanded` roles added.
- keyboard support via `FocusKeyManager`

## Breaking change?

- [ ] Yes
- [x] No

PR #4278, opened 2026-01-07
feat(design): add truncated design to `DaffBreadcrumbComponent`

### LINKED ISSUE #3513: Add truncated design and support for injecting the `DaffMenuModule` to `DaffBreadcrumbComponent`
# :bulb: Feature request

## Feature Name
Add truncated design and support for injecting the `DaffMenuModule` to `DaffBreadcrumbComponent`

## The Desired Behavior
- Provide the UI for full breadcrumb list
- The UI can show a truncated design (i.e. dots) if it's too long 
- On desktop/mobile, truncated design opens a little menu (using the `DaffMenuModule)`
- Breadcrumb accepts a "char count" for total breadcrumb length

## Environment

<pre><code>
Daffodil version: X.Y.Z
Angular version: X.Y.Z 

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

joannalauu:
This also deprecates the DaffMenuModule.

## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #3513

Developed on top of #3569

## New behavior

- Truncates breadcrumbs into an overflow menu on mobile and when more than five items are present on desktop
- Deprecates DaffMenuModule

## Breaking change?

- [ ] Yes
- [x] No


xelaint:
Blocked by #3998 (needs to be merged before this)

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.html
```diff
@@ -1 +1,34 @@
...
+			[daffMenuActivator]="breadcrumbItem.target === 'full' ? fullMenu : partialMenu">
+				<button class="daff-breadcrumb__menu-activator">...</button>
+		</li>
+	}
+}
+
+<ng-template #fullMenu>
+	<daff-menu>
+		@for (breadcrumbItem of _fullMenuItems(); track breadcrumbItem; let index = $index) {
+			<button daff-menu-item>
```
xelaint: We need to ability to choose between a `<button daff-menu-item>` or `<a daff-menu-item>` and pass in the link to the `daff-menu-item`.

This structure is bad:
Image

This should be:

```
<a daff-menu-item href="/docs">
  <div class="daff-menu-item__content">
    Docs
  </div>
</a>
```

xelaint (changes requested):
I don't think it's fully working. This is what I see when I open menu in mobile, close it, then go back to desktop view.

Image

To repro:
1. Update viewport to mobile
2. Click on `...` to open menu
3. Close menu
4. Go back to desktop viewport

xelaint (changes requested):
There's a few things breaking:

1. `active` is no longer set on the active breadcrumb item
2. breadcrumb items are not being rendered correctly on page change

https://github.com/user-attachments/assets/033123ec-f2d4-4e04-b202-55d8f43b7448

joannalauu:
@xelaint I really sorry about not thoroughly review my code before requesting a review. I have fixed the breadcrumb behaviour to work as intended, aside from the daff-menu structure which I have elaborated more in the discussion thread.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.html
joannalauu: I have not been able to find a good solution to achieve this without breaking other behaviour. 

Right now, the breadcrumb is passed into the menu through 
`<ng-container [ngTemplateOutlet]="breadcrumbItem.itemRef"></ng-container>` 
as `<a href="/docs">Docs</a>`, and there isn't really a way to extract the href and innerHTML while keeping the menu content up-to-date, because the menu is usually not re-rendered when switched to another page. 

Perhaps it might be better to just keep the structure as it is, even through it isn't ideal? Please let me know if you have any ideas on an effective solution to this.

damienwebdev:
We need to share the interface between MenuItem and BreadcrumbItem so that the actual HTML element is an a or button in both circumstances. We don't want buttons inside buttons or a in a, or new a where it could have been the original a element from the breadcrumb.

xelaint (changes requested):
`active` is still not set on the active breadcrumb item

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.scss
```diff
@@ -32,6 +34,45 @@
...
+
+		&.on-desktop { /* stylelint-disable-line selector-class-pattern */
+			display: none;
+
+			@include layout.breakpoint(mobile) {
+				display: inline-block;
+			}
+		}
+
+		&:nth-child(2),
```
xelaint, line 54: Can the logic for calculating what is shown be set by `toRenderType` instead of having to do it via css?

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -62,4 +122,111 @@ describe('@daffodil/design/breadcrumb | DaffBreadcrumbComponent', () => {
...
+      breadcrumbComponent = de4.componentInstance;
+      fixture4.detectChanges();
+    });
+
+    it('should render 4 breadcrumb items', () => {
+      const breadcrumbItems = breadcrumbComponent._breadcrumbItems();
+      expect(breadcrumbItems.length).toBe(4);
+    });
+
+    it('should compute 5 breadcrumb items', () => {
```
xelaint: Why would 5 breadcrumb items be computed if there are less than 5 breadcrumbs? Shouldn't the max be 4?

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -22,6 +23,58 @@ class WrapperComponent {
...
+      <li daffBreadcrumbItem><a href="/cat/sub">Subcategory</a></li>
+      <li daffBreadcrumbItem><a href="/cat/sub/curr">Current</a></li>
+    </ol>
+  `,
+  imports: [
+    DaffBreadcrumbComponent,
+    DaffBreadcrumbItemComponent,
+  ],
+})
+class WrapperComponentWith4Items {}
```
xelaint: Please avoid using digits in class names

libs/design/breadcrumb/README.md
```diff
@@ -85,6 +85,7 @@ Breadcrumb follows the [Breadcrumb WAI-ARIA design pattern](https://www.w3.org/W
 ### Daffodil provides
 - `aria-current="page"` automatically applied to the last breadcrumb item
 - Enforces semantic HTML structure (requires `<ol>` and `<li>` elements)
+- Truncates breadcrumbs into an overflow menu on mobile and when more than five items are present on desktop
```
xelaint: Doesn't belong in accessibility. Should be moved into a `## Features` section after `## Anatomy`:

```
## Features

### Truncation
Breadcrumbs are automatically truncated into an overflow menu on mobile viewports. On desktop, truncation occurs when more than five items are present.
```

xelaint (changes requested):

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -62,4 +122,111 @@ describe('@daffodil/design/breadcrumb | DaffBreadcrumbComponent', () => {
...
+    let de7: DebugElement;
+
+    beforeEach(() => {
+      fixture7 = TestBed.createComponent(WrapperComponentWithSevenItems);
+      de7 = fixture7.debugElement.query(By.css('ol[daff-breadcrumb]'));
+      breadcrumbComponent = de7.componentInstance;
+      fixture7.detectChanges();
+    });
+
+    it('should compute 5 breadcrumb items', () => {
```
xelaint: this test is confusing. if there are more than 5 breadcrumb items, it should compute more than just 5.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.html
```diff
@@ -1 +1,29 @@
...
+		</li>
+	} @else {
+		<li class="daff-breadcrumb__item"
+			[daffMenuActivator]="breadcrumbItem.target === 'full' ? fullMenu : partialMenu">
+			<button class="daff-breadcrumb__menu-activator">...</button>
+		</li>
+	}
+}
+
+<ng-template #fullMenu>
```
xelaint: @damienwebdev thoughts on updating this template name to `#mobileMenu` and the `#partialMenu` as `#desktopMenu`? I think this is more semantically correct since these menus are rendered depending on what screen size the display is.

xelaint (changes requested):

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.html
damienwebdev: Yes, this makes sense.

libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.component.spec.ts
```diff
@@ -9,21 +9,21 @@ import {
 } from '@angular/core/testing';
 import { By } from '@angular/platform-browser';
 
-import { DaffBreadcrumbItemDirective } from './breadcrumb-item.directive';
+import { DaffBreadcrumbItemComponent } from './breadcrumb-item.component';
```
xelaint: import from `@daffodil/design/breadcrumb`

libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.component.spec.ts
```diff
@@ -9,21 +9,21 @@ import {
...
 })
 class WrapperComponent {}
 
-describe('@daffodil/design/breadcrumb | DaffBreadcrumbItemDirective', () => {
+describe('@daffodil/design/breadcrumb | DaffBreadcrumbItemComponent', () => {
   let wrapper: WrapperComponent;
   let de: DebugElement;
   let fixture: ComponentFixture<WrapperComponent>;
-  let directive: DaffBreadcrumbItemDirective;
+  let directive: DaffBreadcrumbItemComponent;
```
xelaint: update to `component` since it was converted from directive to component

libs/design/menu/src/menu/menu.component.ts
```diff
@@ -41,7 +44,10 @@ import { DaffMenuService } from '../services/menu.service';
 @Component({
   selector: 'daff-menu',
   templateUrl: './menu.component.html',
-  styleUrl: './menu.component.scss',
```
xelaint, line 44: this change is unnecessary.

libs/design/breadcrumb/src/breadcrumb-menu-item/breadcrumb-menu-item.directive.ts
```diff
@@ -0,0 +1,65 @@
...
+ * Applied to an `ng-container` that projects breadcrumb item templates into `daff-menu`,
+ * allowing breadcrumb items to take on `daff-menu-item` styling and behavior.
+ *
+ * Provides {@link DAFF_MENU_ITEM_TOKEN} so that `DaffMenuComponent` can discover
+ * the component as a menu item. Also adds the `daff-menu-item` class for styling and
+ * implements `focus()` for keyboard navigation.
+ */
+@Directive({
+  selector: '[daffBreadcrumbMenuItem]',
+  host: {
```
xelaint: this class can be removed since it's not used for anything.

libs/design/breadcrumb/src/breadcrumb-menu-item/breadcrumb-menu-item.directive.ts
```diff
@@ -0,0 +1,65 @@
...
+  DaffMenuService,
+} from '@daffodil/design/menu';
+
+/**
+ * @docs-private
+ *
+ * Applied to an `ng-container` that projects breadcrumb item templates into `daff-menu`,
+ * allowing breadcrumb items to take on `daff-menu-item` styling and behavior.
+ *
+ * Provides {@link DAFF_MENU_ITEM_TOKEN} so that `DaffMenuComponent` can discover
```
xelaint: don't need to include this section.

libs/design/breadcrumb/src/breadcrumb-render/breadcrumb-render.type.ts
```diff
@@ -0,0 +1,13 @@
+import { DaffBreadcrumbItemComponent } from '../public_api';
+
+export interface DaffBreadcrumbRenderCrumb {
```
xelaint: `DaffBreadcrumbRenderItem`

libs/design/breadcrumb/src/breadcrumb-render/breadcrumb-render.type.ts
```diff
@@ -0,0 +1,13 @@
...
+  item: DaffBreadcrumbItemComponent;
+  type: 'breadcrumb';
+}
+
+export interface DaffBreadcrumbRenderMenu {
+  type: 'menu';
+  target: 'full' | 'partial';
```
xelaint: `target: 'mobileMenu' | 'desktopMenu'`

libs/design/breadcrumb/src/breadcrumb-render/breadcrumb-render.type.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export type DaffBreadcrumbRender = DaffBreadcrumbRenderCrumb | DaffBreadcrumbRenderMenu;
```
xelaint: `DaffBreadcrumbRenderType`

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -50,16 +64,78 @@ import { DaffBreadcrumbItemDirective } from '../breadcrumb-item/breadcrumb-item.
...
   /**
    * @docs-private
    */
-  @ContentChildren(DaffBreadcrumbItemDirective) breadcrumbItems!: QueryList<DaffBreadcrumbItemDirective>;
+  @ContentChildren(DaffBreadcrumbItemComponent) breadcrumbItems!: QueryList<DaffBreadcrumbItemComponent>;
+
+  /**
+   * @docs-private
+   */
+  @ViewChild('fullMenu', { static: true }) fullMenu: TemplateRef<unknown>;
```
xelaint: `mobileMenu`

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -50,16 +64,78 @@ import { DaffBreadcrumbItemDirective } from '../breadcrumb-item/breadcrumb-item.
...
+
+  /**
+   * @docs-private
+   */
+  @ViewChild('partialMenu', { static: true }) partialMenu: TemplateRef<unknown>;
```
xelaint: `desktopMenu`

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -50,16 +64,78 @@ import { DaffBreadcrumbItemDirective } from '../breadcrumb-item/breadcrumb-item.
...
+      const res = toRenderType(item, items.length, index, this._isMobile());
+      if(Array.isArray(res)) {
+        return [...acc, ...res];
+      } else {
+        return [...acc,  res];
+      }
+    }, []);
+  });
+
+  _partialMenuItems = computed(() => {
```
xelaint: `_desktopMenuItems`

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -50,16 +64,78 @@ import { DaffBreadcrumbItemDirective } from '../breadcrumb-item/breadcrumb-item.
...
+      ) {
+        return [...acc, item];
+      } else {
+        return acc;
+      }
+    }, []);
+    return res;
+  });
+
+  _fullMenuItems = computed(() => {
```
xelaint: `_mobileMenuItems`

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -50,16 +64,78 @@ import { DaffBreadcrumbItemDirective } from '../breadcrumb-item/breadcrumb-item.
...
+   * @docs-private
+   */
+  @ViewChild('partialMenu', { static: true }) partialMenu: TemplateRef<unknown>;
+
+  /**
+   * @docs-private
+   */
+  _breadcrumbItems = contentChildren(DaffBreadcrumbItemComponent);
+
+  _computedBreadcrumbItems = computed(() => {
```
xelaint: All of these private functions should be marked as `@docs-private`

xelaint (changes requested):
This PR has broken the `/docs/design/components/menu` page. It directs me to a 404 error.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -62,4 +122,111 @@ describe('@daffodil/design/breadcrumb | DaffBreadcrumbComponent', () => {
...
+    let de7: DebugElement;
+
+    beforeEach(() => {
+      fixture7 = TestBed.createComponent(WrapperComponentWithSevenItems);
+      de7 = fixture7.debugElement.query(By.css('ol[daff-breadcrumb]'));
+      breadcrumbComponent = de7.componentInstance;
+      fixture7.detectChanges();
+    });
+
+    it('should compute the first 2 and last 2 breadcrumb items and an ellipsis', () => {
```
xelaint: The tests should be broken up into sections about desktop and mobile, because this is test param is only relevant on desktop, not because there are more than 5 breadcrumb items.

I would create separate spec files, e.g.:

`desktop.spec.ts`
`mobile.spec.ts`

with test cases for each less than or more than five items in desktop specifically. it's irrelevant for mobile since any amount of breadcrumbs over one gets truncated into the menu.

xelaint (changes requested):

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.desktop.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    TestBed.configureTestingModule({
+      imports: [
+        WrapperComponentWithFourItems,
+        WrapperComponentWithSevenItems,
+      ],
+    })
+      .compileComponents();
+  }));
+
+  beforeEach(() => {
```
xelaint: was this generated by LLM? it's not used at all in this test.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.desktop.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    spyOn(DaffBreadcrumbItemComponent.prototype, 'setActive').and.callThrough();
+  });
+
+  describe('when given 4 breadcrumb items', () => {
```
xelaint: this should also test that activator is not rendered.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.mobile.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(WrapperComponentWithFourItems);
+    de = fixture.debugElement.query(By.css('ol[daff-breadcrumb]'));
+    breadcrumbComponent = de.componentInstance;
+
+    fixture.detectChanges();
+  });
+
+  beforeEach(() => {
```
xelaint: was this generated by LLM? it's not used at all in this test.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.mobile.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    </ol>
+  `,
+  imports: [
+    DaffBreadcrumbComponent,
+    DaffBreadcrumbItemComponent,
+  ],
+})
+class WrapperComponentWithFourItems {}
+
+describe('@daffodil/design/breadcrumb | DaffBreadcrumbComponent mobile', () => {
```
xelaint:
```suggestion
describe('@daffodil/design/breadcrumb | DaffBreadcrumbComponent - Mobile Viewport', () => {
```

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.desktop.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+		</ol>
+	`,
+  imports: [
+    DaffBreadcrumbComponent,
+    DaffBreadcrumbItemComponent,
+  ],
+})
+class WrapperComponentWithSevenItems {}
+
+describe('@daffodil/design/breadcrumb | DaffBreadcrumbComponent desktop', () => {
```
xelaint:
```suggestion
describe('@daffodil/design/breadcrumb | DaffBreadcrumbComponent - Desktop Viewport', () => {
```

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.desktop.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    it('should compute 4 breadcrumb items', () => {
+      expect(breadcrumbComponent._computedBreadcrumbItems().length).toBe(4);
+    });
+
+    it('desktop menu should be empty', () => {
+      expect(breadcrumbComponent._desktopMenuItems().length).toBe(0);
+    });
+
+    it('should have the active class only on the last breadcrumb item', () => {
```
xelaint: This test can be moved to the defaults test file (`breadcrumb.component.spec.ts`) since this is true regardless of desktop or mobile view.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.mobile.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  it('mobile menu should contain all items except the last one', () => {
+    const mobileMenuItems = breadcrumbComponent._mobileMenuItems();
+    const allBreadcrumbItems = breadcrumbComponent._breadcrumbItems();
+    expect(mobileMenuItems.length).toBe(3);
+    expect(mobileMenuItems).toEqual(allBreadcrumbItems.slice(0, allBreadcrumbItems.length - 1));
+  });
+
+  it('should have the active class only on the last breadcrumb item', () => {
```
xelaint: This test can be moved to the defaults test file (`breadcrumb.component.spec.ts`) since this is true regardless of desktop or mobile view.

xelaint (changes requested):
1. This should be reverted [ac97b0b](https://github.com/graycoreio/daffodil/commit/ac97b0b7c7adf21b5a9f8b3b75d66ab7e3847a26) since it causes a flash on render. I didn't realize SSR doesn't know what the screen size is. sorry about this!

2. move the desktop and mobile specs into a spec folder under `src`. the name of the spec doesn't need to include `breadcrumb.component`. this is our general formatting guideline for specs in design. (e.g. input)

joannalauu:
> 1. This should be reverted [ac97b0b](https://github.com/graycoreio/daffodil/commit/ac97b0b7c7adf21b5a9f8b3b75d66ab7e3847a26) since it causes a flash on render. I didn't realize SSR doesn't know what the screen size is. sorry about this!

Since we're dropping commit [ac97b0b](https://github.com/graycoreio/daffodil/commit/ac97b0b7c7adf21b5a9f8b3b75d66ab7e3847a26), I don't think there is any value in creating separate spec files for desktop and mobile, because the breadcrumb rendered is the exact same in both cases - the scss controls which items are viewable depending on viewport size .

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.spec.ts
```diff
@@ -9,19 +9,48 @@ import {
...
-import { DaffBreadcrumbComponent } from './breadcrumb.component';
+import {
+  DaffBreadcrumbComponent,
+  DaffBreadcrumbItemComponent,
+} from '@daffodil/design/breadcrumb';
 
 @Component({
-  template: `<ol daff-breadcrumb [skeleton]="skeleton"></ol>`,
+  template: `
+    <ol daff-breadcrumb [skeleton]="skeleton">
```
xelaint, line 19: @damienwebdev can you provide more information on how this should be updated to an iterated list and the test cases we discussed?

libs/design/breadcrumb/src/breadcrumb-item/breadcrumb-item.component.ts
```diff
@@ -14,16 +19,32 @@ import {
...
   selector: 'li[daffBreadcrumbItem]',
+  template: `
+    <ng-template #item>
+      <ng-content></ng-content>
+    </ng-template>
+  `,
+  changeDetection: ChangeDetectionStrategy.OnPush,
   host: {
     'class': 'daff-breadcrumb__item',
     '[class.active]': '_active',
```
xelaint: `'[class.active]': '_active',` is no longer needed since the active class is added in the template.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.html
```diff
@@ -1 +1,31 @@
...
+	@if (breadcrumbItem.type === 'breadcrumb') {
+		<li class="daff-breadcrumb__item"
+			[class.active]="breadcrumbItem.item._active">
+			<ng-container *ngTemplateOutlet="breadcrumbItem.item.itemRef"></ng-container>
+		</li>
+	} @else {
+		<li class="daff-breadcrumb__item"
+			[class.on-mobile]="breadcrumbItem.target === 'mobileMenu'"
+			[class.on-desktop]="breadcrumbItem.target === 'desktopMenu'"
+			[daffMenuActivator]="breadcrumbItem.target === 'mobileMenu' ? mobileMenu : desktopMenu">
```
xelaint: the `[daffMenuActivator]` should be moved down into the button.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.html
```diff
@@ -1 +1,31 @@
...
+			<button class="daff-breadcrumb__menu-activator">...</button>
```
xelaint: add `aria-label="Show more breadcrumb items"`

xelaint (changes requested):

xelaint (dismissed):

xelaint:
@damienwebdev needs a final review from you.

libs/design/menu/src/menu-item/menu-item.token.ts
```diff
@@ -0,0 +1,11 @@
...
+import { InjectionToken } from '@angular/core';
+
+/**
+ * Injection token used to identify menu items.
+ *
+ * `DaffMenuComponent` uses `@ContentChildren(DAFF_MENU_ITEM_TOKEN)` to discover
+ * its menu items. Providing this token allows custom directives and components
+ * to be treated as menu items alongside native `daff-menu-item` elements.
+ */
+export const DAFF_MENU_ITEM_TOKEN = new InjectionToken<FocusableOption>('DAFF_MENU_ITEM_TOKEN');
```
griest024: This should be created with the service token factory so you can have a provider too

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -50,35 +67,111 @@ import { DaffBreadcrumbItemDirective } from '../breadcrumb-item/breadcrumb-item.
...
-    this.breadcrumbItems.last.setActive(true);
+  /**
+   * @docs-private
+   */
+  _computedBreadcrumbItems = computed(() => {
+    this._isMobile(); // signal rerenders breadcrumb on viewport change
+
+    const items = this._breadcrumbItems();
+
+    return items.reduce<DaffBreadcrumbRender[]>((acc, item, index) => {
```
griest024: Seems like this is equivalent to 
```ts
items.flatMap((item, index) => toRenderType(item, items.length, index))
```
do you agree?

libs/design/breadcrumb/src/breadcrumb-render/to-render-type.ts
```diff
@@ -0,0 +1,25 @@
+import { DaffBreadcrumbItemComponent } from '../public_api';
+import { DaffBreadcrumbRender } from './breadcrumb-render.type';
+
+export const toRenderType = (item: DaffBreadcrumbItemComponent, length: number, index: number): DaffBreadcrumbRender | DaffBreadcrumbRender[] => {
```
griest024: this needs docs. The return type is confusing. What does it mean to return an array vs just an object?

Edit: looking at this function more makes me think the return should just be `DaffBreadcrumbRender | null`. The calling of it in the component also becomes more clear. The default case should just be to return `null`, which indicates that it should be in the menu.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.ts
```diff
@@ -50,35 +67,111 @@ import { DaffBreadcrumbItemDirective } from '../breadcrumb-item/breadcrumb-item.
...
+  _desktopMenuItems = computed(() => {
+    const items = this._breadcrumbItems();
+    const res = items.reduce<DaffBreadcrumbItemComponent[]>((acc, item, index) => {
+      if(items.length >= 5
+          && (index !== 0 && index !== 1)
+          && (index !== items.length - 1 && index !== items.length - 2)
+      ) {
+        return [...acc, item];
+      } else {
+        return acc;
```
griest024: this is just duplicating business logic that is already in `toRenderType`, not to mention that these conditionals are really hard to read. What I think is the better approach is to use `_computedBreadcrumbItems` to partition the items into visible and menu lists. Then `_desktopMenuItems` simply becomes selecting the `menu` key from your partition.

libs/design/breadcrumb/src/breadcrumb/breadcrumb.component.html
```diff
@@ -1 +1,32 @@
-<ng-content select="[daffBreadcrumbItem]"></ng-content>
\ No newline at end of file
+@for (breadcrumbItem of _computedBreadcrumbItems(); track breadcrumbItem; let index = $index) {
```
griest024: where is `index` used, why is it bound into the scope? Seems like it can just be removed

libs/design/breadcrumb/src/breadcrumb-menu-item/breadcrumb-menu-item.directive.ts
```diff
@@ -0,0 +1,58 @@
...
+  focus() {
+    this._focusableElement?.focus(); // Allows `FocusKeyManager` to focus on the element
+  }
+
+  private _findFocusableElement(): HTMLElement | null {
+    let node = this._elementRef.nativeElement.previousElementSibling;
+    while (node && node.nodeType !== 1) {
+      node = node.nextSibling;
+    }
+    return node;
```
griest024: This relies on angular's render ordering in the DOM which is an implementation detail subject to change without notice. I think a much less brittle implementation that relies on `ViewContainerRef` is:
```ts
const view = <EmbeddedViewRef<unknown>>this._viewContainerRef.get(0);
return view?.rootNodes.find((n): n is HTMLElement => n.nodeType === Node.ELEMENT_NODE) ?? null;
```

griest024 (dismissed):

xelaint (approved):

damienwebdev (approved):
lgtm

PR #4321, opened 2026-02-06
feat(design): `DaffModalService.open` returns reference instead of modal instance

### LINKED ISSUE #4270: `DaffModalService.open` should return a handle or ref, not the modal component
### Feature description

`DaffModalService.open` should not return `DaffModalComponent`. It should return some kind of handle that contains all the required info without exposing the modal component itself.

### Use case

As an app dev, I don't want to be enabled to use antipatterns by touching `DaffModalComponent`.

### Prior work

#4045

### Daffodil version

0.90

### Additional context

_No response_

joannalauu:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #4270 
Part of: # 

## New behavior

`DaffModalService.open` returns reference instead of modal instance. The reference provides access to `close()` and the `afterClosed` observable.

## Breaking change?

- [ ] Yes
- [x] No


libs/design/modal/src/modal/interfaces/modal.ts
```diff
@@ -0,0 +1,32 @@
+import { OverlayRef } from '@angular/cdk/overlay';
+import { ComponentRef } from '@angular/core';
+import { Observable } from 'rxjs';
+
+import { DaffModalComponent } from '../modal.component';
+
+export interface DaffModal {
```
damienwebdev, line 6: split this into two files I think to avoid accidental public_api exposure.

damienwebdev (changes requested):
I think this is a good starting point.

damienwebdev (dismissed):
lgtm

xelaint (dismissed):

libs/design/modal/src/modal/interfaces/modal-ref.ts
```diff
@@ -0,0 +1,16 @@
...
+  /**
+   * Closes the modal.
+   */
+  close(): void;
+
+  /**
+   * Emits when the modal close animation completes.
+   */
+  afterClosed: Observable<boolean>;
+}
```
griest024: While I think there is an argument for YAGNI here, I think it makes sense to expose the component instance (NOT the modal instance) on this object. The fact that [material exposes this](https://material.angular.dev/components/dialog/api#MatDialogRef) convinces me that its a common enough use case.

So you should add a field to this interface: `instance: T` and add the type param to the interface.

libs/design/modal/src/service/modal.service.ts
```diff
@@ -99,7 +100,7 @@ export class DaffModalService {
   open(
     component: Type<any>,
     configuration?: Partial<DaffModalConfiguration>,
-  ): DaffModalComponent {
+  ): DaffModalRef {
```
griest024: This function will need a type param as well, which will change the type of `component` to `Type<T>` and the return to `DaffModalRef<T>`

griest024 (changes requested):

griest024 (approved):

PR #4342, opened 2026-02-23
ci(commerce): Build `@daffodil/commerce` schematics

### LINKED ISSUE #4253: ci: test cases for `@daffodil/commerce` that validate generated app
### Feature description

We need e2e test cases that validate that the app generated by `@daffodil/commerce` runs and behaves as expected.

### Use case

As a daffodil maintainer, I would like to know if my change will break the commerce schematics.

### Prior work

_No response_

### Daffodil version

0.90

### Additional context

Essentially there needs to be an automated script that can run in CI that:
1. scaffolds out an angular app (`ng new`)
2. run the ecommerce schematic (`ng add @daffodil/ecommerce`)
3. builds the resulting app
4. determines whether or not that build was successful (do we try to serve the app and run some e2e tests?)

joannalauu:
## PR Checklist

## PR Type

- [ ] Bug fix
- [ ] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [x] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #4253

## New behavior

Builds the following commerce schematic configurations:
- driver=demo, standalone, app.config.ts
- driver=in-memory, standalone, app.config.ts
- driver=magento, standalone, app.config.ts
- driver=shopify, standalone, app.config.ts
- driver=demo, skip-package-json, standalone, app.config.ts
- driver=demo, module-based, app.config.ts
- driver=demo, standalone, main.ts fallback

## Breaking change?

- [ ] Yes
- [x] No


tools/schematics/e2e/test-commerce-schematics.sh
```diff
@@ -0,0 +1,275 @@
...
+    1)
+      # Driver: demo, standalone, app.config.ts
+      desc="driver=demo (standalone, app.config.ts)"
+      copy_base_app "$STANDALONE_APP_DIR" "$WORK_DIR/case-1"
+      raise_budget
+      if run_ng_add "demo" && npx ng build 2>&1; then
+        report_result "$case_num" "$desc" "pass"
+      else
+        report_result "$case_num" "$desc" "fail"
+      fi
```
griest024: Each case data should be in some kind of serializable format, e.g. JSON, so that cases can easily be added and changed without modifying this file.

This will require some abstraction around what you're doing in case 7. Maybe the JSON can reference some override files that this script will know how to use. Or we have a few hardcoded starting points that the script can use.

Hopefully that makes sense. The goal is that this just becomes a for loop and the switch case is removed.

griest024 (dismissed):

tools/schematics/e2e/test-commerce-schematics.sh
```diff
@@ -0,0 +1,275 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# ============================================================
```
damienwebdev: From a discussion with @joannalauu -- I think it would be very nice to move a lot of this into the github actions workflow to both parallelize the tests as well as make this a bit easier to read.

damienwebdev (changes requested):

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - uses: graycoreio/github-actions/setup-node@main
+        with:
+          node-version: 22.21.x
+          use-stamp-cache: true
+
+      - name: Start Verdaccio
```
damienwebdev: I think this can be replaced with a container running in the pipeline: https://www.verdaccio.org/docs/docker/

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+          config: ${{ env.VERDACCIO_CONFIG }}
+          storage: ${{ env.VERDACCIO_STORAGE }}
+
+      - name: Set test version
+        run: |
+          npm version "${{ env.TEST_VERSION }}" --no-git-tag-version
+          sed -i "s/version = '.*'/version = '${{ env.TEST_VERSION }}'/" \
+            tools/schematics/ng-add/generators/version.ts
+
+      - name: Build @daffodil/* packages
```
damienwebdev: We need some way to describe these as dependencies of `@daffodil/commerce` (perhaps devDependencies) so that we don't name them explicitly here.

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+        run: |
+          npx nx run-many -t build \
+            --projects=@daffodil/core,@daffodil/driver,@daffodil/product,@daffodil/dev-tools,@daffodil/navigation,@daffodil/external-router
+          cd tools/schematics && npm run build
+
+      - uses: graycoreio/github-actions/set-versions-from-root@main
+
+      - name: Publish @daffodil/* packages to Verdaccio
+        run: |
+          for pkg in ${{ env.DAFFODIL_PACKAGES }}; do
```
damienwebdev: similarly to the build task, I think this should just be a `npx nx run @daffodil/commerce:publish`

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+            build: true
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 22.21.x
```
damienwebdev: Move to matrix.

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+
+      - name: Restore base apps and Verdaccio storage
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            ${{ env.WORK_DIR }}
+            ${{ env.VERDACCIO_STORAGE }}
```
damienwebdev: I don't think we should cache verdaccio. build / publish in the job should be fast.

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+          key: commerce-test-${{ github.run_id }}
+
+      - name: Start Verdaccio
```
damienwebdev: I think this can be replaced with a container running in the pipeline: https://www.verdaccio.org/docs/docker/

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+    dev-tools
+    navigation
+    external-router
+
+jobs:
+  setup:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
damienwebdev: use latest action version.

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+            skip_package_json: false
+            succeed: true
+            build: true
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+
+      - uses: actions/setup-node@v4
```
damienwebdev: Use the graycore action.

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+        run: |
+          BASE_APP="${{ env.WORK_DIR }}/test-daff-app"
+          if [ "${{ matrix.base }}" = "css" ]; then
+            BASE_APP="${{ env.WORK_DIR }}/test-daff-app-css"
+          fi
+          TEST_DIR="${{ env.WORK_DIR }}/test-${{ matrix.name }}"
+          rm -rf "$TEST_DIR"
+          cp -a "$BASE_APP" "$TEST_DIR"
+
+      - name: Apply app modifications
```
damienwebdev: This should be done with `--standalone=false` during generation.

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+          });
+          MAIN
+          rm -f "src/app/app.config.ts"
+          fi
+
+          if [ "${{ matrix.routing }}" = "false" ]; then
+          rm -f "src/app/app.routes.ts"
+          fi
+
+      - name: Generate schematic
```
damienwebdev: To guarantee that users can do the exact command as we do in CI can you just do `npx ng add @daffodil/commerce@$TEST_VERSION`?

.github/workflows/build-commerce-schematic.yml
```diff
@@ -0,0 +1,275 @@
...
+        working-directory: ${{ env.WORK_DIR }}/test-${{ matrix.name }}
+        run: |
+          EXTRA_FLAGS=""
+          if [ "${{ matrix.skip_package_json }}" = "true" ]; then
+            EXTRA_FLAGS="--skip-package-json"
+          fi
+
+          if [ "${{ matrix.succeed }}" = "true" ] || [ "${{ matrix.build }}" = "true" ]; then
```
damienwebdev: Instead of complex bash conditionals, I would just do `continue-on-error: ${{ matrix.succeed }}` for the tests we know are currently problematic.

damienwebdev (changes requested):

.github/workflows/build-commerce-schematic.yml
joannalauu: I looked into this approach, and I believe that if we replace the existing composite action with service containers, we'll need to restart the container within `setup` job and `test` job. This is because the service starts before workflow steps, but the `VERDACCIO_CONFIG` repository file and the Verdaccio cache is only available during the workflow steps. These restarts seem wasteful - do you think it is still worth it to go forward with this change, or is there an existing approach that can prevent these restarts while using service containers?

damienwebdev: I don't think the Verdaccio cache is particularly valuable, this is essentially the same as our build cache which should already be available to this job.

.github/workflows/build-commerce-schematic.yml
joannalauu: build/publish takes approximately a minute total, so I currently continue caching verdaccio, but let me know if you still want me to move this to the `test` job

.github/workflows/build-commerce-schematic.yml
joannalauu: I'm not sure if moving node-version to matrix is worth it, because we will have to generate additional base apps with the different node-versions, but base apps are generated in the `setup` step, before the test matrix is defined. Testing node version support may add more complexity than its worth, but let me know if you'd still like node version to be tested, and I'll think of a structure to support this

.github/workflows/build-commerce-schematic.yml
joannalauu: I am not sure if this is the best approach, because `continue-on-error` will allow the test to pass when we expect the step to fail but the step passes - this would prevent us from catching unexpected behavior and regressions. I implemented a fix that removes duplicate code - let me know if that would be sufficient

tools/schematics/e2e/test-commerce-schematics.sh
damienwebdev: This matrix was added.

.github/workflows/build-commerce-schematic.yml
```diff
@@ -90,6 +90,7 @@ jobs:
         with:
           cache-mode: read-only
           read-only-azure-connection-string: ${{ secrets.AZURE_NX_CACHE_READONLY_CONNECTION_STRING }}
+          read-write-azure-connection-string: ${{ secrets.AZURE_NX_CACHE_READWRITE_CONNECTION_STRING }}
```
damienwebdev: This shouldn't be required.

damienwebdev (changes requested):

.github/workflows/build-commerce-schematic.yml
```diff
@@ -66,15 +66,6 @@ jobs:
             ng-add-succeed: true
             build-succeed: false
 
-          - name: no-app-routes
```
damienwebdev: This test was good, why did you remove it?

joannalauu: To test for `no-app-routes`, we need to modify the base app in the workflow by removing `app.routes.ts`. This modification step was removed in your diff, so this test was not actually testing `no-app-routes` behaviour anymore. I personally thought that it wasn't worth it to add back the modification step, because there isn't really a user situation where `app.routes.ts` would be missing from the codebase. However, if you think this is still a worthwhile condition to test, I can add back the test and the modification step.

damienwebdev: http://angular.dev/cli/new 

`ng new` has a `--routing` flag that can be set to `false` to disable routing.

joannalauu: I've added this change, let me know if there is anything else that I need to fix.

damienwebdev (approved):
lgtm

PR #4372, opened 2026-03-06
fix(design): set disabled attribute from ngControl in form field controls

### LINKED ISSUE #4371: Native select, native input, and native textarea's disabled ngControl does not work
### Is this a regression?

- [x] Yes, this behavior used to work in the previous version

### Description

Native select, native input, and native textarea's disabled ngControl does not work.

This might have been caused by some update in Angular recently.

### Environment

- Daffodil version: 0.91.0
- Angular version:
- Browser(s):
- Operating system:

### Steps to reproduce

Steps to reproduce:
1. 
2.
3.

### Expected behavior

Disabling via `ngControl` should set disable on <select>, <input>, or <textarea>.

### Actual behavior

Form controls are not disabled.

### Additional context

_No response_

xelaint:
## PR Checklist

## PR Type

- [x] Bug fix
- [ ] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #4371

## Breaking change?

- [ ] Yes
- [x] No

### DIFF
--- a/libs/design/textarea/src/specs/disabled.spec.ts
+++ b/libs/design/textarea/src/specs/disabled.spec.ts
@@ -140,8 +140,8 @@ describe('@daffodil/design/textarea | DaffTextareaComponent | Reactive Forms Dis
       expect(component.disabled).toEqual(true);
     });
 
-    it('should set disabledAttribute to true', () => {
-      expect(component.disabledAttribute).toEqual(true);
+    it('should set the disabled attribute native textarea element', () => {
+      expect(componentDE.nativeElement.getAttribute('disabled')).not.toBeNull();
     });
   });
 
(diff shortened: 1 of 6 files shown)


damienwebdev:
@xelaint please add test coverage for each component to prove that we have the `disabled` html prop on each element when using a `formControl`

xelaint:
@damienwebdev done

libs/design/textarea/src/specs/disabled.spec.ts
```diff
@@ -148,5 +152,12 @@ describe('@daffodil/design/textarea | DaffTextareaComponent | Reactive Forms Dis
 
       expect(component.disabled).toEqual(false);
     });
+
+    it('should remove the disabled attribute from the host element', () => {
+      wrapper.message.enable();
+      fixture.detectChanges();
+
+      expect(component.disabledAttribute).toBeFalsy();
```
damienwebdev: These tests are checking the `component.disabledAttribute` property. I think (given that this has seemingly broken unexpectedly a few times) we should explicitly test that the hostElement itself has a `disabled` property defined.

damienwebdev: I think something like `getAttribute('disabled')` being `""` or `"true"` would be sufficient.

xelaint: i used `not.toBeNull();` since disabled is set like this `<textarea disabled></textarea>`

damienwebdev (approved):
lgtm

Issue #4327, opened 2026-02-10
Migrate `@daffodil/design` to use signal inputs


xelaint:
### Summary

https://angular.dev/reference/migrations/signal-inputs

### Objectives

Migrate all usage of `@Input()` to `input()`

- [x] @daffodil/design/accordion (#4557)
- [x] @daffodil/design/article
- [x] @daffodil/design/breadcrumb
- [ ] ~~@daffodil/design/callout~~
- [x] @daffodil/design/card
- [ ] ~~@daffodil/design/checkbox~~ in the process of being refactored
- [ ] ~~@daffodil/design/container~~
- [x] @daffodil/design/form
- [ ] @daffodil/design/form-field
- [ ] ~~@daffodil/design/hero~~
- [x] @daffodil/design/image
- [ ] @daffodil/design/input
- [x] ~~@daffodil/design/link-set~~ this component is deprecated
- [x] @daffodil/design/list
- [x] ~~@daffodil/design/loading-icon~~ this component is deprecated
- [x] @daffodil/design/media-gallery
- [x] @daffodil/design/menu
- [ ] @daffodil/design/modal
- [x] @daffodil/design/navbar
- [x] @daffodil/design/notification
- [x] @daffodil/design/paginator
- [x] @daffodil/design/progress-bar
- [ ] @daffodil/design/quantity-field
- [ ] @daffodil/design/radio
- [x] @daffodil/design/spinner
- [ ] @daffodil/design/select
- [ ] @daffodil/design/sidebar
- [x] @daffodil/design/switch
- [x] @daffodil/design/tabs
- [x] @daffodil/design/tag
- [x] @daffodil/design/text-snippet
- [ ] @daffodil/design/textarea
- [x] @daffodil/design/toast
- [x] @daffodil/design/tree

damienwebdev:
We will need to block the first merge on any of these for us to guarantee that component docs document inputs / outputs correctly in daff.io. i.e. https://next.daff.io/docs/design/components/tree?tab=api-tab

Blocked by https://github.com/graycoreio/daffodil/issues/4366

griest024:
there is an automated migration to do this as well: https://angular.dev/reference/migrations/signal-inputs

PR #4408, opened 2026-04-03
feat(design, daffio): convert tree to signals and then make daffio render the tree sanely

### LINKED ISSUE #2931: [FEAT] A parent tree item should be expanded if there is an active/selected tree item within it

damienwebdev:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

### Design

  - The sidebar navigation tree does not auto-expand to reveal the currently active page. On initial render and on every page change, all parent nodes are collapsed, hiding the active item.                                          
  - The tree component and tree item directive use legacy @Input()/@ContentChild() decorators and OnChanges/OnInit lifecycle hooks with manual change detection. 

### Daffio

  - The sidebar body component is destroyed and recreated on every route navigation, even when the sidebar registration hasn't changed. This causes unnecessary re-initialization and loss of tree expansion state.
  - The storefront and design docs routes share a factory function (daffioDocsDesignRoutesFactory) that takes a section string and extra routes, coupling them unnecessarily. 
 
Fixes: #2931 
Part of: #4327 

### Design

  - Tree auto-expands ancestors of selected nodes. When a tree item's selected input becomes true, all ancestor nodes are automatically opened via the new `daffTreeOpenAncestors` utility, making the active page visible in the sidebar.
  - Tree component and directive use signals. @Input() replaced with input()/input.required(), @ContentChild with contentChild(), derived state with computed(), and side-effects with effect(). The tree component uses a _revision signal to trigger re-flattening instead of manual markForCheck(). 
  
 ### Daffio

  - Sidebar body component persists across navigations. activeRegistration$ now uses distinctUntilChanged on the registration id, preventing unnecessary component destruction. The nav list composable (useDaffioNavList) uses DaffRouterDataService instead of ActivatedRoute so it receives updated route data even when the component isn't recreated.
  - Storefront routes are independent. The route factory is replaced with two plain route arrays: daffioDocsDesignRoutes and daffioDocsStorefrontRoutes. Storefront routes no longer include the behaviors section and the design examples are no longer shared on both areas.  

## Breaking change?

- [x] Yes
- [ ] No

### DIFF
--- a/apps/daffio/src/app/docs/storefront/storefront.routes.ts
+++ b/apps/daffio/src/app/docs/storefront/storefront.routes.ts
@@ -3,32 +3,32 @@ import { Routes } from '@angular/router';
 import { DaffSidebarModeEnum } from '@daffodil/design/sidebar';
 import {
   DAFF_DOC_KIND_PATH_SEGMENT_MAP,
-  DAFF_DOCS_DESIGN_PATH,
+  DAFF_DOCS_STOREFRONT_PATH,
   DaffDocKind,
 } from '@daffodil/docs-utils';
 
-import { provideDaffioDocsDesignComponentContentComponent } from './components/component-content/component-content.provider';
-import { DAFFIO_DOCS_DESIGN_LIST_SIDEBAR_REGISTRATION } from './containers/docs-list/sidebar.provider';
-import { provideDaffioDesignExamplesContent } from './examples/content.provider';
-import { DaffioDocsDesignComponentOverviewPageComponent } from './pages/components-overview/component-overview.component';
-import { DaffioDocsDesignOverviewPageComponent } from './pages/overview/overview.component';
-import { daffioDocsDesignComponentDocResolver } from './resolvers/component-doc.resolver';
-import { daffioDocsDesignComponentListResolverFactory } from './resolvers/component-list.resolver';
-import { daffioDocsDesignIndexResolver } from './resolvers/index.resolver';
-import { provideDaffioDocsDesignIndexService } from './services/index.service';
+import { provideDaffioStorefrontExamplesContent } from './examples/content.provider';
+import { DaffioDocsStorefrontOverviewPageComponent } from './pages/overview/overview.component';
 import { DaffioRoute } from '../../core/router/route.type';
 import { daffioDocsApiRolesProvider } from '../api/roles/api-roles.provider';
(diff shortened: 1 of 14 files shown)


damienwebdev:
@xelaint there's some consideration here for "scrolling" the tree's container to the selected element when the page renders, but I think this needs to be "opt-in" rather than always. Additionally, accomplishing this is a little hacky as I probably wouldn't want it to trigger after the page render.

apps/daffio/src/app/docs/storefront/storefront.routes.ts
```diff
@@ -0,0 +1,87 @@
...
+import { DAFFIO_DOCS_DESIGN_LIST_SIDEBAR_REGISTRATION } from '../design/containers/docs-list/sidebar.provider';
+import { DaffioDocsDesignComponentOverviewPageComponent } from '../design/pages/components-overview/component-overview.component';
+import { daffioDocsDesignComponentDocResolver } from '../design/resolvers/component-doc.resolver';
+import { daffioDocsDesignComponentListResolverFactory } from '../design/resolvers/component-list.resolver';
+import { daffioDocsDesignIndexResolver } from '../design/resolvers/index.resolver';
+import { provideDaffioDocsDesignIndexService } from '../design/services/index.service';
+import { DaffioDocsPageComponent } from '../pages/docs-page/docs-page.component';
+import { DocsResolver } from '../resolvers/docs-resolver.service';
+
+export const daffioDocsStorefrontRoutes = <Routes>[
```
griest024, line 24: these are still very similar to the design routes. You prefer having them separate vs improving the route factory?

damienwebdev: Yes. I believe that similarity is only temporary, it also had the likely unexpected side-effect of causing no diff between the sidebar registration between the two sections.

griest024 (approved):

xelaint (approved):

damienwebdev:
@griest024 @xelaint i will submit PRs breaking this down shortly.

PR #4424, opened 2026-04-10
feat(external-router): Support different Magento driver versions

### LINKED ISSUE #3137: [FEAT] Create strategy for supporting different platform versions
# :bulb: Feature request

## Feature Name

Create strategy for supporting different platform versions

## The Desired Behavior

The following questions need answers:
- When a new version of a platform is released and we update the driver, what happens to the previous version?
- What should the migration path be for apps that upgrade to the new platform version?
- What should the migration path be for apps that *don't* upgrade to the new platform version but upgrade to a new daffodil version?

joannalauu:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #3137

## New behavior

- Added providers for Magento driver v2.4.1 and v2.4.2
- Users can import Magento driver from `@daffodil/external-router/driver/magento/` instead of `@daffodil/external-router/driver/magento/2.4.3`
  - This is done through patching the `./driver/magento` export condition to resolve to the specific driver version in the dist package.json
- Maintainer sets default Magento driver version in `libs/external-router/scripts/patch-magento-exports.mjs`
  - Currently set to v2.4.3
- Users can customize Magento driver version through modifying `angular.json` and `tsconfig.json` 
  - Details in `guides/drivers/magento.md`
- Commerce test matrix now includes tests for all Magento versions
  - Tests verify that only configured Magento driver version is in build bundle
  - Versions are retrieved through compute-matrix script, so the matrix doesn't need to be manually updated when new versions are added
  - When default driver version is updated, `env.DEFAULT_MAGENTO_VERSION` will need to be updated 
- Demo app will still need to import from `@daffodil/external-router/driver/magento/2.4.3` because patch only applies to imports from `node_modules`, not `dist/`

## Breaking change?

- [ ] Yes
- [x] No


joannalauu:
Tests in commit 47d7b9aee0a9f7acdc5cc5d755aaf5c3d8d73bcd uses the updated commerce test workflow, validating expected behavior

joannalauu:
Test failures in commit e1588cc3ab74c931ecd236351b81e0b4dae5ce59 verifies that when using a base provider to control driver version, tree shaking doesn't work and unused driver versions are included in the bundle

damienwebdev:
After a lengthy discussion with @joannalauu we came to a new idea. Essentially there will be a new `@daffodil/*/driver/magento/auto` entrypoint in all Magento supported packages that leverages a new file called `daff.json` which contains version definitions for various platforms / drivers - potentially something like:

```json
{
  "projects": {
    "provider-test": {
      "drivers": {
        "magento": "1.9.4"
      }
    }
  }
}
```

We also will have tests guaranteeing that our bundles sizes remain small.

Below follows the list of notes that we generated during our discussion.

=========

1. What's the definitions for `daff.json`?
   1. Docs
2. What's the command interface / command package / per-package command for converting `daff.json` to angular.json / tsconfig.json?
   1. How does this command handle multiple projects within a repo like `@angular/cli` multiple projects.
      1. `@angular/cli` - `daff.json` + `angular.json`  root with some way to break projects apart
      2. Determine which kind of project dynamically.
   2. Docs
3. What does the `conditions` / `custom-conditions` spec look like? For different platforms? For different versions, etc. What's the minimum possible configurations?
   1. Internal docs
   2. What does it look across packages?
   3. How do we guarantee that packages don't drift in their exports? Linting rule? 
      1. Do we need to guarantee completeness of the `exports.driver/magento` across all Magento versions for all Daffodil packages?
4. We need to guarantee that `provideDaffMagentoXDriver`  is the same. We need an interface guarantee from the root `@daffodil/*/driver/magento` package.
   1. This causes a per-package diff cascade
   2. We'd also have to update every package to provider functions over modules.
5. Extract bundle size tests to a separate workflow than test-commerce-schematic. What does this actually test?
   1. What do we want to verify? 
      1. For each most-recent platform version against the `@daffodil/commerce` schematic:
      	1. That the `command` to generate the original `daff.json` works.
      	2. The the `command` to update `angular.json` from `daff.json` works
      	3. That the `command` to update `angular.json` from a revised `daff.json` works.
      	4. That the bundle only includes the code expected from that version for ONE function from each of the `@daffodil/*` packages installed.
6. Consider the entrypoint `@daffodil/*/driver/magento/auto` as a way to get rid of `@daffodil/*/driver/magento/base`
   1. Users can opt-in to the `auto` import if they would like, allowing this whole setup to be non-breaking.

daff.json - `angular.json` multiple projects
```json
{
  "projects": {
    "provider-test": {
      "drivers": {
        "magento": "1.9.4"
      }
    }
  }
}

`daff.json` for nx projects w/ project.json
```json
{
  "drivers": {
    "magento": "2.4.7",
		"shopify": "2026-01",
		"hubspot": "kjnasdkjhasbdkjahsbdlkjhasbd"
  }
}
```

griest024 (commented):
I don't like this solution. You've confirmed that the unused driver versions don't get tree shaken out?

damienwebdev:
@griest024 have you reviewed my comment above?

joannalauu:
> I don't like this solution. You've confirmed that the unused driver versions don't get tree shaken out?

Yes, test failures in commit https://github.com/graycoreio/daffodil/commit/e1588cc3ab74c931ecd236351b81e0b4dae5ce59 verifies that when using a base provider to control driver version, tree shaking doesn't work and unused driver versions are included in the bundle

griest024:
> @griest024 have you reviewed my comment above?

yes, see the discord thread. I think this requires further discussion

griest024:
superseded by #4629

PR #4621, opened 2026-07-30
fix(navigation): ensure fragment fields on leaf nodes in magento

damienwebdev:
## PR Checklist

## PR Type

- [x] Bug fix

## Current behavior

`getCategoryNodeFragment` builds the recursive `CategoryTree` selection with a reduce, spreading extra fragments (registered via `DAFF_NAVIGATION_MAGENTO_DRIVER_EXTRA_FRAGMENTS`) into each wrapper iteration but not into the reduce seed. The seed is the deepest selection set, so a depth-N query ends up with N+1 selection sets and only N extra fragment spreads. Fields contributed by extra fragments are never requested at the deepest tier of the tree, and any navigation tree transforms reading those fields get `undefined` for the bottom-level nodes.

This regressed in #2889, which inlined the node fields to work around magento/magento2#31086. Before that change the spread lived inside the named `categoryNode` fragment, which the seed referenced, so every level got it.

Fixes: #

## New behavior

The extra fragment spread is computed once and included in both the reduce seed and each wrapper iteration, so every generated `CategoryTree` selection set requests the extra fragment fields, including the deepest one. `daffBuildFragmentNameSpread` returns an empty string when no extra fragments are registered, so the default case produces the same query as before.

I added two regression tests. One walks the generated fragment AST and asserts the spread is present at every nesting level (depth + 1 levels), and one runs a query through the Apollo testing controller and asserts a field contributed by an extra fragment survives on the deepest child node.

## Breaking change?

- [ ] Yes
- [x] No

### DIFF
--- a/libs/navigation/driver/magento/src/queries/fragments/category-node/category-node.spec.ts
+++ b/libs/navigation/driver/magento/src/queries/fragments/category-node/category-node.spec.ts
@@ -17,6 +17,13 @@ import { CategoryNode } from '@daffodil/navigation/driver/magento';
 
 import { getCategoryNodeFragment } from './category-node';
 
+type CategoryNodeWithExtraField = CategoryNode & {
+  extra_field: string;
+  children?: Array<CategoryNode & {
+    extra_field: string;
+  }>;
+};
+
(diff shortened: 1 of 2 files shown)


griest024 (approved):
ooof nasty regression, good find!

PR #4641, opened 2026-08-07
feat(design)!: remove default theme variables from scss 'theme' export

### LINKED ISSUE #4640: Daffodil's default theme should not be forwarded as part of `@daffodil/design/scss/theme`
### Is this a regression?

- [ ] Yes, this behavior used to work in the previous version

### Description

`@daffodil/design/scss/theme` is documented as a utility entrypoint — a collection of Sass variables, mixins, and functions used to build a theme. But it also forwards Daffodil's default theme, so every app that uses the entrypoint gets Daffodil's brand theme whether or not it wants it.

The chain is:

- `libs/design/scss/theme.scss` forwards `theming`
- `libs/design/scss/theming/_index.scss` forwards `daff-theme`
- `libs/design/scss/theming/_daff-theme.scss` builds `$theme` and `$theme-dark` from the `$daff-blue`, `$daff-purple`, and `$daff-turquoise` palettes that are exclusive to Daffodil

The result is that `$theme` and `$theme-dark` are part of the public surface of `@daffodil/design/scss/theme`. An app that only wants `daff-create-theme`, `daff-configure-palette`, or `daff-component-themes` in order to build its own theme still builds Daffodil's default one.

This is a problem for a few reasons:

1. **The default theme is opinionated branding, not a utility.** Daffodil's blue/purple/turquoise palette is a product decision for Daffodil's own branding. It shouldn't be handed to every app by default.
2. **There's no way to avoid it.** Because the default theme is forwarded from the same entrypoint as the theming utilities, an app can't take the utilities without also taking the default theme.
3. **It blurs the line between "the tools to build a theme" and "a theme."** `daff-theme.$theme` reads like the app's theme, when it is really Daffodil's theme. That makes it easy to accidentally ship Daffodil branding.

Using Daffodil's default theme should be an explicit opt-in. Daffodil's default theme configurations should be forwarded in a separate entrypoint: `@daffodil/design/scss/theme/default`

### Environment

- Daffodil version: 0.93.1
- Angular version:
- Browser(s):
- Operating system:

### Steps to reproduce

Steps to reproduce:
1. In an app's `styles.scss`, use the theme entrypoint only to build a custom theme:

   ```scss
   @use '@daffodil/design/scss/theme' as daff-theme;

   $primary: daff-theme.daff-configure-palette(daff-theme.$daff-green, 60);
   $secondary: daff-theme.daff-configure-palette(daff-theme.$daff-orange, 60);
   $tertiary: daff-theme.daff-configure-palette(daff-theme.$daff-red, 60);

   $theme: daff-theme.daff-create-theme((
     'primary': $primary,
     'secondary': $secondary,
     'tertiary': $tertiary,
     'mode': 'light',
   ));

   html {
     @include daff-theme.daff-component-themes($theme);
   }
   ```

2. Note that `daff-theme.$theme` and `daff-theme.$theme-dark` — Daffodil's default theme, built from Daffodil's brand palettes — are still available and still evaluated, even though the app never asked for them.

### Expected behavior

`@daffodil/design/scss/theme` should only expose the tools needed to build and apply a theme: palettes, `daff-configure-palette`, `daff-create-theme`, `daff-component-themes`, and the theme accessor functions.

Daffodil's default theme should live behind its own opt-in entrypoint, so that using it is a deliberate choice:

```scss
@use '@daffodil/design/scss/theme' as daff-theme;
@use '@daffodil/design/scss/theme/default' as daff-default-theme;

html {
  @include daff-theme.daff-component-themes(daff-default-theme.$theme);
}
```

### Actual behavior

`@daffodil/design/scss/theme` forwards `daff-theme`, so `$theme` and `$theme-dark` come along with the theming utilities and there is no way to take one without the other.

### Additional context

This is a breaking change.

damienwebdev:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior
Currently, we expose many theme variables unexpectedly (like `$tertiary`).

Fixes: #4640
Part of: # 

## New behavior
We no longer expose these variables by default.

## Breaking change?

- [x] Yes
- [ ] No

### DIFF
--- a/apps/daffio/src/scss/styles.scss
+++ b/apps/daffio/src/scss/styles.scss
@@ -1,9 +1,10 @@
 // DO NOT IMPORT THIS FILE ANYWHERE ELSE
 
 @use 'sass:meta';
-@use 'global';
-@use 'theme' as daff-theme;
+@use '@daffodil/design/scss/global';
+@use '@daffodil/design/scss/theme' as daff-theme;
 @use './component-themes' as daffio-components;
+@use '@daffodil/design/scss/theming/default' as daff-default-theme;
 
 // @daffodil/branding
 @use '@daffodil/branding/scss/component-themes' as branding;
@@ -17,10 +18,10 @@
 
 	@include meta.load-css('highlight.js/scss/a11y-light');
 
-	@include branding.daff-branding-theme(daff-theme.$theme);
-	@include daff-theme.daff-component-themes(daff-theme.$theme);
-	@include daffio-components.component-themes(daff-theme.$theme);
-	@include docs.daff-docs-component-themes(daff-theme.$theme);
+	@include branding.daff-branding-theme(daff-default-theme.$theme);
+	@include daff-theme.daff-component-themes(daff-default-theme.$theme);
+	@include daffio-components.component-themes(daff-default-theme.$theme);
+	@include docs.daff-docs-component-themes(daff-default-theme.$theme);
 }
 
 .daff-theme-dark {
@@ -29,10 +30,10 @@
 
 	@include meta.load-css('highlight.js/scss/a11y-dark');
 
-	@include branding.daff-branding-theme(daff-theme.$theme-dark);
-	@include daff-theme.daff-component-themes(daff-theme.$theme-dark);
-	@include daffio-components.component-themes(daff-theme.$theme-dark);
-	@include docs.daff-docs-component-themes(daff-theme.$theme-dark);
+	@include branding.daff-branding-theme(daff-default-theme.$theme-dark);
+	@include daff-theme.daff-component-themes(daff-default-theme.$theme-dark);
+	@include daffio-components.component-themes(daff-default-theme.$theme-dark);
+	@include docs.daff-docs-component-themes(daff-default-theme.$theme-dark);
 }
 
 body {
--- a/apps/daffio/src/scss/styles.scss
+++ b/apps/daffio/src/scss/styles.scss
@@ -4,7 +4,7 @@
 @use '@daffodil/design/scss/global';
 @use '@daffodil/design/scss/theme' as daff-theme;
 @use './component-themes' as daffio-components;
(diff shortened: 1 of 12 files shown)


libs/design/scss/theme/default.scss
```diff
@@ -0,0 +1,15 @@
+// This file only exists for builds inside the Daffodil monorepo. Consumers of
```
xelaint, line 1: Don't think this file is used or needed.

xelaint (changes requested):

apps/daffio/src/scss/styles.scss
```diff
@@ -1,9 +1,10 @@
 // DO NOT IMPORT THIS FILE ANYWHERE ELSE
 
 @use 'sass:meta';
-@use 'global';
-@use 'theme' as daff-theme;
+@use '@daffodil/design/scss/global';
+@use '@daffodil/design/scss/theme' as daff-theme;
 @use './component-themes' as daffio-components;
+@use '@daffodil/design/scss/theming/default' as daff-default-theme;
```
xelaint: I would change this to `@use '@daffodil/design/scss/theme/default' as daff-default-theme;` to be consistent with external usage

xelaint (changes requested):

xelaint (changes requested):
`libs/design/guides/getting-started.md#add-a-theme` needs to be updated

apps/daffio/src/scss/styles.scss
damienwebdev: 3f7b2257552f1d1abdf0971f271b4e2600bbd9f0

xelaint (approved):

PR #4661, opened 2026-08-14
feat(design): create `daff-color-shift` function

### LINKED ISSUE #4660: Add a `daff-color-shift` function to select colors relative to a palette's default hue
### Feature description

Component themes currently hardcode the hues used for hover and active states. In `libs/design/button/src/button/basic/button-theme.scss`, the light-mode secondary button is written as:

```scss
&.daff-secondary {
	@include daff-button-theme-variant(
		daff-color($secondary),
		daff-color($secondary, 70),
		daff-color($secondary, 80)
	);
}
```

The base color follows whatever hue the palette was configured with, but `70` and `80` do not. They only line up because `daff-configure-palette` defaults to hue `60`. A theme configured with `daff-configure-palette($daff-purple, 30)` gets a light base with hover and active states that are much darker, instead of one step darker each.

Add a `daff-color-shift($palette, $steps)` function that finds the palette's default hue and returns the color that many steps away from it:

```scss
&.daff-secondary {
	@include daff-button-theme-variant(
		daff-color($secondary),
		daff-color-shift($secondary, 1),
		daff-color-shift($secondary, 2)
	);
}
```

- [x] Update `daff-configure-palette` to record the `$hue` as `default-hue`
- [x] `daff-color-shift` function to shift the default hue `x` number of steps up or down. An error should be thrown if there are no more steps/colors the palette can go.

### Use case

As someone configuring a theme, I'd like hover and active states to follow the hue I chose for a palette, so that a palette configured at a hue other than the standard default "60" will still have proper hover and active state styling.

### Prior work

`daff-illuminate($color, $palette, $nth)` had a similar concept. It was deprecated in 0.87.0 by #3815 and has since been removed. It measured steps as luminance distance from a reference color and chose the direction on its own, returning a darker color if the reference was bright and a lighter one if it was dark.

Two things separate this function from `daff-illuminate`:
- **The direction stays explicit.** `daff-illuminate` inferred it from luminance, which made the result hard to predict. `$steps` is easy to understand, and the `light` and `dark` mixins can be used to set different colors.
- **The result stays inside the palette's hue keys.** No luminance sorting, no reference color, no reordering of the palette. `daff-color-shift($primary, 1)` resolves to a hue in the palette and nothing else.

### Daffodil version

0.93.1

### Additional context

_No response_

xelaint:
## PR Checklist

## PR Type

- [ ] Bug fix
- [x] Feature
- [ ] Style update
- [ ] Refactor
- [ ] Test
- [ ] Build
- [ ] CI
- [ ] Docs
- [ ] Performance
- [ ] Other (please describe)

## Current behavior

Fixes: #4660

## Breaking change?

- [ ] Yes
- [x] No

### DIFF
--- a/libs/design/scss/theming/_configure-palette.scss
+++ b/libs/design/scss/theming/_configure-palette.scss
@@ -7,6 +7,9 @@
 /// as the palette’s default color. The palette can then be used
 /// to configure themes by setting them as primary, secondary, or tertiary colors.
 ///
+/// The hue is recorded alongside the color as `default-hue`, so that colors can
+/// be selected relative to the default with `daff-color-shift`.
+///
 /// @group Theming
 ///
 /// @param {Map} $color-palette — A color palette map (e.g., `$daff-blue`, `$daff-purple`).
@@ -21,7 +24,8 @@
 	@return map.merge(
 		$color-palette,
 		(
-			'default': get-color.daff-color($color-palette, $hue)
+			'default': get-color.daff-color($color-palette, $hue),
+			'default-hue': $hue
 		)
 	);
 }
--- a/libs/design/scss/theming/_index.scss
+++ b/libs/design/scss/theming/_index.scss
@@ -1,8 +1,9 @@
 @forward 'color-palettes';
 @forward 'get-color';
+@forward 'color-shift';
 @forward 'contrast';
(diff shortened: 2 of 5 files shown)


damienwebdev (approved):
lgtm

